PR #1, opened 2018-06-26
Add ropsten to list of networks

pcardune:
This just uses infura to connect to ropsten so you don't have to run a local node.

### DIFF
--- a/truffle.js
+++ b/truffle.js
@@ -2,6 +2,7 @@ require('babel-register')({
   ignore: /node_modules\/(?!zeppelin-solidity)/,
 });
 require('babel-polyfill');
+const HDWalletProvider = require('truffle-hdwallet-provider');
 
 module.exports = {
   // See <http://truffleframework.com/docs/advanced/configuration>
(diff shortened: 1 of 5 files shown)


truffle.js
```diff
@@ -26,6 +27,23 @@ module.exports = {
...
+    ropsten: {
+      network_id: 3,
+      gas: 4000000,
+      provider: () => {
+        if (!process.env.MNEMONIC || !process.env.INFURA_KEY) {
+          throw new Error(
+            'You must set both the MNEMONIC and INFURA_KEY environment variables to use the ropsten network'
+          );
+        }
+        return new HDWalletProvider(
```
jaycenhorton: I cant tell what this is accomplishing. Wouldnt this create an HD wallet and then assign that HD wallet to accounts[0-9] in the deploy/migrations file? If so, this is definitely NOT what we want. We should instead have a static accounts file that uses our ledgers/dev wallet (as I had started to configure in `/test/helpers/accounts/`) thoughts?

jaycenhorton: Not sure what youre using locally, but I was using aledger HW wallet for everything

pcardune: This is just to make it easier to connect to ropsten. It's not necessarily for deploying contracts unless you happen to use something with a mnemonic, in which case I don't see the downside assuming we write the deploy scripts properly. In any case, I was just using this to introspect what you already deployed to ropsten.

PR #3, opened 2018-06-26
Rename artifacts helper to Artifacts

pcardune:
We should strive to make imports match the name of the file being imported. So
instead of

```js
const Artifacts = require('./test/helpers/artifacts');
```

we have

```js
const Artifacts = require('./test/helpers/Artifacts');
```

Separate, I'm not sure I understand the point of this file. What's wrong with
just using `artifacts.require()`?


jaycenhorton:
There's sometimes an issue when the global artifacts isn't passed into the context of a helper file. I think this can be solved by not using arrow functions in certain areas to pass the context properly.

pcardune:
The solution is to not assume globals. I find the fact that truffle migration scripts use those globals to be very unfortunate.

PR #6, opened 2018-06-26
Deploy script that works on both ropsten and locally

pcardune:


migrations/1_initial_migration.js
```diff
@@ -1,7 +1,5 @@
-/* globals artifacts */
-
-const Migrations = artifacts.require('./Migrations.sol');
-
 module.exports = function deploy(deployer) {
-  deployer.deploy(Migrations);
+  deployer.then(async () => {
+    // yeah we won't be using this...
```
jaycenhorton: Why not?
Fyi you can use `truffle migrate -f{the step number to start from}` to run all migrations starting from a given step

migrations/2_deploy_registry.js
```diff
@@ -0,0 +1,52 @@
...
+const { deployUpgradeableContract } = require('../test/helpers/contracts');
+const getNamedAccounts = require('../test/helpers/getNamedAccounts');
+
+const ContractRegistryV0_1_0 = artifacts.require('ContractRegistryV0_1_0');
+
+module.exports = (deployer, network, accounts) => {
+  deployer.then(async () => {
+    let registry, adminAccountAddress;
+    if (network === 'ropsten') {
+      adminAccountAddress = accounts[0];
```
jaycenhorton: We should force a sanity check here so that we can't deploy from a truffle account

test/helpers/contracts.js
```diff
@@ -7,6 +7,11 @@ function getLogs(Event, filter, additionalFilters) {
   return promisify(query.get.bind(query))();
 }
 
+const parseContractName = contractName => {
```
jaycenhorton: What's the benefit of this vs split(v) ? I'm no good with regex, so does this expression satisfy if we have say `contractV10_12_999-beta`?

migrations/1_initial_migration.js
pcardune: because it's completely useless unless you store the Migrations artifact so it can look up the address of the migration contract on the network. With our versioned registry, we've replaced Migrations with something that is much more powerful and which makes the Migrations contract obsolete.

This script basically makes it so that running `migrate` *always* runs from the beginning and puts the registry and all the contracts into the correct state.

migrations/2_deploy_registry.js
pcardune: I'm trying to think of how to do that. It seems like the truffle team periodically changes the default mnemonic that gets used. I will think on this some and try to come up with something good.

jaycenhorton: I'm sure the seed exists in one of the truffle repos. Maybe query it for changes

test/helpers/contracts.js
jaycenhorton: Good catch.

migrations/2_deploy_registry.js
jaycenhorton: You might also check here for some example? Although I think our upgradeable implementation is superior to any I've seen now 😏 https://github.com/JoinColony/colonyNetwork/tree/develop/helpers

pcardune: The open zeppelin example you linked to does not use a shared mnemonic. Looks like it works exactly the same as this with a mnemonic specified via an environment variable.

pcardune: I get the idea. I just don't see a documented way of specifying a mnemonic for truffle to use short of instantiating your own provider, and even when you do that, `truffle develop` ignores the provider you specified.

In any case, I'm not that concerned about someone accidentally deploying to main net using a public seed phrase so long as the ENS entry always points to the right place. And once the registry is properly deployed to main net with the multisig wallet as the owner, it's not like you'll be able to change anything using truffle keys anyway.

jaycenhorton: I am pretty worried actually, but not for the reason described here. Mainly for one in which is using another edcase that addes account[0] to multisig, transfers ownership or sets owner to. 

BUT, there's actually a pretty straightforward solution I think we both overlooked. *Whitelist our two developer addresses. *

pcardune:
Ok, this should be ready to go now.

migrations/2_deploy_registry.js
```diff
@@ -10,6 +10,18 @@ module.exports = (deployer, network, accounts) => {
     let registry, adminAccountAddress;
     if (network === 'ropsten') {
       adminAccountAddress = accounts[0];
+      if (
+        ![
+          '0xf1bcd758cb3d46d15afe4faef942adad36380148',
+          '0x2e4d8353d81b7e903c9e031dab3e9749e8ab69bc',
+        ].includes(adminAccountAddress.toLowerCase())
```
jaycenhorton: why are we `.toLowerCase`ing this? why not use the web3 checksum api?

PR #7, opened 2018-06-28
Ditch non-upgradeable FifoCrcMarket and SelectableCrcMarket

pcardune:


pcardune:
Why does it need to be kept in this repo?

pcardune:
This one should also be ready to go. I just went ahead and created SelectableCrcMarketV0, which uses the non-deprecated base class.

PR #8, opened 2018-06-28
Remove old TonToken contract

pcardune:


package.json
```diff
@@ -32,7 +32,7 @@
     "testnet:compile": "yarn run lint:sol && truffle compile",
     "testnet:deploy": "yarn run lint:sol && truffle deploy --network noritestnet",
     "testnet:migrate": "yarn run lint:sol && truffle migrate --network noritestnet",
-    "test": "yarn run lint:sol && `yarn bin`/truffle test --reset --all",
+    "test": "yarn run lint:sol && truffle test --reset --all",
     "test:gasreport": "cd gas-reporter && truffle test --reset",
```
jaycenhorton: This will prefer the global truffle version if you have one. This can be a tricky thing. I actually need a global truffle version for vscodes Solidity plugin, but don't want to run the global version for command such as this. Thoughts?

test/behaviors/UnstructuredUpgradeableToken.js
```diff
@@ -86,20 +84,6 @@ const shouldBehaveLikeUnstructuredUpgradeableTokenV0 = (
     );
   });
 
-  context(
```
jaycenhorton: So one issue with this is that unstructuredtontokentedts actually was testing almost all of the 777 functionalities before. Removing this drastically reduces code coverage. If we want to remove this we should rewrite the test cases existing inside of here.

PR #13, opened 2018-07-02
add a regitry to keep track of the registry's versions

jaycenhorton:
This was already working before, but this makes it more explicit. It creates a seprate Registry contract which can be versioned and keeps track only of the registries registry


jaycenhorton:
agreed, but dont like that either

pcardune:
Or RegistryToRuleThemAll? (only half joking)

jaycenhorton:
i like that, I was going to do `RegistryOfRegistries` so Im ok with that

pcardune:
Or RootRegistry? It's the root because it doesn't appear in any registry

jaycenhorton:
changed

PR #15, opened 2018-07-02
make it so tests only run via 0_setup_tests

jaycenhorton:
there arent really any changes here other than exporting tests so theyre not auto run via truffle test without allowing them in 0-setup-tests

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -16,6 +16,8 @@ module.exports = {
   globals: {
     contract: false,
     assert: false,
+    web3: true,
+    artifacts: true,
   },
   rules: {
     camelcase: 'off',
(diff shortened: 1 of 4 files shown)


.eslintrc.js
```diff
@@ -16,6 +16,8 @@ module.exports = {
   globals: {
     contract: false,
     assert: false,
+    web3: true,
+    artifacts: true,
```
pcardune: These should be scoped to the files they are actually globals in. which as far as I know is just deploy scripts and 0_setup_tests.js

jaycenhorton: also Artifacts.js and a couple others I believe. I notice I am adding it often though, and not sure why it should be scoped locally

PR #17, opened 2018-07-03
adds all the registry events

jaycenhorton:
## UnstructuredOwnedUpgradeabilityProxy
```solidity
event UpgradeabilityOwnerSet(address upgradeabilityOwner); // called during construction, sets who can upgrade the registry
event RegistryAddrSet(address registryAddress); // called when the registry implementation is set <- which is used to look up either the root registry or contract registry for storing versions/names/etc
```
## VersionRegistry
```solidity
event VersionSet(string contractName, address proxyAddress, string versionName, address newImplementation); // called when upgradeTo or upgradeToAndCall is invoked on a proxy. It calls the contract registry or root registry, sets the latest version, and then tells the world the c name, proxy address, v name, and where the proxy points to
```
## UnstructuredUpgradeabilityProxy
```solidity
event ImplementationPositionSet(address indexed impPosition); // tells where the proxy points to. Called during upgrades.
```
UnstructuredOwnable
```solidity
event OwnerSet(address newOwner); // called when an initial owner, or new owner is set. Typically only useful for construction

```
## ContractRegistryBase
```solidity
event Initialized(address owner); // called to tell who owns the contract registry or root registry. invoked after init
```


jaycenhorton:
[Log events for all contract registry state changes](https://trello.com/c/Gt70jlOq/332-log-events-for-all-contract-registry-state-changes)

contracts/registry/VersionRegistry.sol
```diff
@@ -5,6 +5,8 @@ pragma solidity ^0.4.24;
 /// @dev keeps track of the latest version of all contracts
 contract VersionRegistry {
 
+  event VersionSet(string contractName, address proxyAddress, string versionName, address newImplementation);
```
pcardune: Can strings be indexed? If so, I think it would be useful to index on `contractName`. It would also be useful to index on `proxyAddress`

PR #18, opened 2018-07-03
remove deprecated crc/commodity contracts and fix selectable market tests

jaycenhorton:
tests are fixed in #19


jaycenhorton:

PR #19, opened 2018-07-03
make selectable market upgradeable, remove non upgradeable

jaycenhorton:
This also fixes test issues in #18


jaycenhorton:

PR #20, opened 2018-07-03
deprecate nonupgradeable supplier files

jaycenhorton:


jaycenhorton:

PR #22, opened 2018-07-03
deprecate old 820imp

jaycenhorton:
deprecates use of the old eip820implementer contract


jaycenhorton:

PR #23, opened 2018-07-03
deprecate tontoken finally

jaycenhorton:
removes tontoken contracts, tests, etc. You should be using NoriVN if you're not already


jaycenhorton:

PR #34, opened 2018-07-03
add new balanceOf upgrade

jaycenhorton:
This can be a perfect exercise for us to see how we might handle an upgrade on ropsten, want to spend some time on this together today?

This `balanceOf` effectively deprecates the original version, and so it would be cool if we can work through a live upgrade exercise to prove we can:
- sign independently to the multisig
- upgrade the CRC
- preserve the CRCs previous balances

### DIFF
--- a/contracts/crc/CRCV0_2_0.sol
+++ b/contracts/crc/CRCV0_2_0.sol
@@ -1,0 +1,24 @@
+pragma solidity ^0.4.24;
+import "./CRCV0.sol";
+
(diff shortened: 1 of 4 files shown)


jaycenhorton:
[CRC.balanceOf should be sum of values of owned CRCs](https://trello.com/c/XvbJOGfq/296-crcbalanceof-should-be-sum-of-values-of-owned-crcs)

PR #42, opened 2018-07-05
sync with latest

jaycenhorton:
replaces #34 to make sure its in sync. This still needs to be deployed to ropsten. See #34 for comments on that


migrations/3_deploy_nori.js
```diff
@@ -7,7 +7,7 @@ const ContractRegistryV0_1_0 = artifacts.require('ContractRegistryV0_1_0');
 const RootRegistryV0_1_0 = artifacts.require('RootRegistryV0_1_0');
 const NoriV0_1_0 = artifacts.require('NoriV0_1_0');
 const ParticipantRegistryV0 = artifacts.require('ParticipantRegistryV0');
-const CRCV0_1_0 = artifacts.require('CRCV0_1_0');
+const CRCV0_2_0 = artifacts.require('CRCV0_2_0');
```
pcardune: this is where it would be nice to have a wrapper like:
```
const CRC = artifacts.getLatest('CRC');
```
which would just give you the latest version.

contracts/crc/CRCV0_2_0.sol
```diff
@@ -0,0 +1,24 @@
...
+    uint totalCrcValue = 0;
+    
+    for (uint i = 0; i < supply || i < bundleBalance; i++) {
+      address ownerToCheck = BasicCommodity.ownerOf(i);
+      if (ownerToCheck == _owner) {
+        totalCrcValue += BasicCommodity.getCommodityValueByIndex(i);
+      }
+    }
+    
+    return totalCrcValue;
```
pcardune: I was imagining we would denormalize the data and store a separate balance int somewhere. I guess the relevant questions to ask are:

- will we ever call balanceOf from inside a contract, in which case this function could end up costing a lot of gas
- how many bundles do we expect people to have?
- how error prone is it to keep a denormalized sum field in sync?

What are your thoughts here?

test/behaviors/FifoCrcMarket.js
```diff
@@ -84,10 +84,11 @@ const shouldBehaveLikeFifoCrcMarketV0 = admin => {
           for (let i = 0; i < crcToMint; i++) {
             crc.mint(supplier, '', token(crcValue), '');
           }
+
           suppliersCrcBal += crcToMint;
           assert.equal(
             await crc.balanceOf(supplier),
-            suppliersCrcBal,
+            token(suppliersCrcBal),
```
pcardune: where is `token()` defined?

pcardune:
[CRC.balanceOf should be sum of values of owned CRCs](https://trello.com/c/XvbJOGfq/296-crcbalanceof-should-be-sum-of-values-of-owned-crcs)

contracts/crc/CRCV0_2_0.sol
jaycenhorton: maybe using `mBalnces`? (currently unused). I dont imagine we ever use this inside a contract. Previously we had a JS function that did this same thing external of the blockchain, but that was removed when we deprecated the old `contracts.js` file

test/behaviors/FifoCrcMarket.js
jaycenhorton: one of the helper files. It basically just is a shortcut for web3.toWei

contracts/crc/CRCV0_2_0.sol
pcardune: I guess I would vote for using something like mBalances.

jaycenhorton: Ok. trying to find a delicate balance between "getting features done" and "implementing features right using upgrades". Thoughts on handling this by updating the base files vs handling it via a new version?

jaycenhorton: it requires changes to `mint` , and if we use the upgrade-only journey, we are going to end up with some pretty unreadable code. I think we should reserve upgrades for live changes, and should probably keep to just modifying base files as much as possible until we get the closest thing to a working base first. Thoughts?

jaycenhorton: basically-  its entirely possible to preserve state and do this in an upgraded version, but im not sure theres any state worth preserving just yet. The tradeoff is readability

jaycenhorton:
:+1: on it

PR #43, opened 2018-07-05
Use latest participant registry version semantic

jaycenhorton:
I think this is evreything.


pcardune:
what exactly am i looking at here?

jaycenhorton:
just resolves all the issues that come up when running `yarn lint:sol` basically stops using deprecated solidty types, spacing issues, consistent versions, etc

jaycenhorton:
it was opened into the wrong branch. This should be easier to review

PR #47, opened 2018-07-07
add bundle balance and value balance functions

jaycenhorton:


test/helpers/contracts.js
```diff
@@ -73,6 +86,67 @@ const deployUpgradeableContract = async (
   ];
 };
 
+const getLatestVersionFromFs = async contractName => {
+  const dir = path.join(__dirname, '../../contracts/');
+  let latestVersion = '0_0_0';
+  return new Promise(res => {
```
pcardune: can you just call this `resolve` instead of `res`?

test/helpers/contracts.js
```diff
@@ -73,6 +86,67 @@ const deployUpgradeableContract = async (
...
+        const [latestMajor, latestMinor, latestPatch] = latestVersion.split(
+          '_'
+        );
+        let [, versionToCompare] = parseContractName(file);
+        versionToCompare = versionToCompare.split('.')[0];
+        const [major, minor, patch] = versionToCompare.split('_');
+        if (parseInt(major, 10) > parseInt(latestMajor, 10)) {
+          latestVersion = versionToCompare;
+        } else if (
+          parseInt(major, 10) === parseInt(latestMajor, 10) &&
```
pcardune: this could just be `major === latestMajor`. Unless you wanted `V1_0_0` to match `V01_0_0`.

test/helpers/contracts.js
```diff
@@ -73,6 +86,67 @@ const deployUpgradeableContract = async (
...
+          parseInt(patch, 10) > parseInt(latestPatch, 10)
+        ) {
+          latestVersion = versionToCompare;
+        }
+      });
+      res(latestVersion);
+    });
+  });
+};
+
```
pcardune: There's also https://www.npmjs.com/package/compare-semver

test/behaviors/Crc.js
```diff
@@ -199,15 +252,15 @@ const deployUpgradeableCrc = async (CrcContract, admin, contractRegistry) => {
...
     artifacts,
     null,
-    CrcContract,
+    'CRC',
     contractRegistry,
     initParams
   );
 
-  return [participantRegistry, proxy, crc];
+  return [participantRegistry, proxy, crc, versionName];
```
pcardune: I just noticed that proxy and crc are switched from `deployLatestUpgradeableContract`. Is that intentional? For the future, I think you should absolutely avoid returning arrays and instead return objects with keys so that ordering mistakes are less likely to happen.

jaycenhorton: I agree, I started the object switch with the newer functions and these should also be switched immediately, but need to fix associated tests. Yes I believe it was definitely intentional. Idr why

PR #48, opened 2018-07-10
add ens stuffz

jaycenhorton:
These are the prerequisites for consuming ENS inside of truffle develop/test. It deploys an ENS registry, creates a registrar, and assigns `nori.eth` so it can be consumed in migrations. This pr doesn't really consume it, and subsequent prs begin that journey, but probably arent perfect either

see: https://twitter.com/jaycenhorton/status/1015790432926568448


jaycenhorton:
[Add ENS to truffle](https://trello.com/c/Nv8PK0Mk/379-add-ens-to-truffle)

contracts/ens/FIPSRegistrar.sol
```diff
@@ -0,0 +1,36 @@
+pragma solidity ^0.4.24;
+
+import "./ENS.sol";
+
+/**
+ * @title FIPSRegistrar is a registrar that allocates subdomains to the first person to claim them.
+ */
+contract FIFSRegistrar {
```
pcardune: FIPS or FIFS?

migrations/1_configure_ens.js
```diff
@@ -0,0 +1,51 @@
+const namehash = require('eth-ens-namehash');
+
+const ENS = artifacts.require('./ENSRegistry.sol');
+const FIFSRegistrar = artifacts.require('./FIFSRegistrar.sol');
```
pcardune: don't use file paths. just use class names.

test/helpers/ens.js
```diff
@@ -0,0 +1,22 @@
...
+
+const getENSDetails = async (network, artifacts, web3) => {
+  let ens, resolver;
+  console.log(`Looking up existing registry at nori.test ENS on ${network}`);
+  if (network === 'ropstenGeth' || network === 'ropsten') {
+    ens = new ENS(web3.currentProvider);
+    resolver = ens.resolver('nori.test').addr();
+  } else {
+    try {
+      ens = await artifacts.require('./ENSRegistry.sol').deployed();
```
pcardune: ditto

test/helpers/ens.js
```diff
@@ -0,0 +1,22 @@
...
+      resolver = ens.resolver(namehash.hash('nori.eth'));
+    } catch (e) {
+      console.log('make sure ENS is deployed first');
+    }
+  }
+  return artifacts.require('./RootRegistryV0_1_0.sol').at(await resolver);
```
pcardune: ditto

contracts/ens/ENSRegistry.sol
```diff
@@ -0,0 +1,99 @@
+pragma solidity ^0.4.24;
+
+import "./ENS.sol";
+
+/**
+  * @title The ENS registry contract
+*/
+contract ENSRegistry is ENS {
```
pcardune: Are the ENS contracts not available in an existing repository somewhere? Did you hand code this or copy it from somewhere? Seems weird that we would have to have our own implementation of ENS in here.

contracts/ens/FIPSRegistrar.sol
jaycenhorton: Fifs

jaycenhorton: will fix title note

contracts/ens/ENSRegistry.sol
jaycenhorton: No hand code, stripped bare minimum of what I needed from https://github.com/ensdomains/ens/tree/master/contracts

jaycenhorton:
The reason for this is because ALL existing ens implementations for local development are simply broken for truffle. They try spinning up multiple instances of testrpcs and propogate stack underflows inside truffles console. Ive notified all appropriate node module maintainers. For now this is the easiest solution.

Asking for you to trust me on that bit, as I spent a few hours of research on twittering with maintainers to verify
 https://twitter.com/jaycenhorton/status/1015790432926568448

pcardune:
ok, that makes sense. Assuming those node modules are hosted on github, we should open github issues and link to them from here in case anyone else runs into the same issues.

PR #51, opened 2018-07-10
multisig migration setup

jaycenhorton:
this is only part of the multisig upgrades. it introduces a new migration to configure the multisig wallets and transfers ownership over to them after deployment. Dont look much past the 2nd or third migration in this PR, as those ones will change in the next pr

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,6 +22,7 @@
     "ethereum-ens": "^0.7.4",
     "ethereumjs-abi": "^0.6.5",
     "ethereumjs-testrpc-sc": "^6.1.2",
+    "glob": "^7.1.2",
     "prettier": "^1.13.6",
     "solidity-coverage": "^0.5.4",
     "solium": "^1.1.7",
--- a/test/helpers/contracts.js
+++ b/test/helpers/contracts.js
@@ -12,11 +12,12 @@ function getLogs(Event, filter, additionalFilters) {
 
 const deployOrGetRootRegistry = async config => {
   const { network, artifacts, web3, deployer } = config;
-  if (
-    network === 'develop' &&
-    !(await artifacts.require('RootRegistryV0_1_0').deployed())
-  ) {
-    return deployer.deploy(artifacts.require('RootRegistryV0_1_0'));
+  if (network === 'develop') {
+    try {
+      await artifacts.require('RootRegistryV0_1_0').deployed();
+    } catch (e) {
+      return deployer.deploy(artifacts.require('RootRegistryV0_1_0'));
+    }
   }
   const rootRegistry = await ensUtils.getENSDetails(network, artifacts, web3);
   if (rootRegistry) {
--- a/test/helpers/contracts.js
+++ b/test/helpers/contracts.js
@@ -12,7 +12,7 @@ function getLogs(Event, filter, additionalFilters) {
 
 const deployOrGetRootRegistry = async config => {
   const { network, artifacts, deployer } = config;
-  if (network === 'develop') {
+  if (network === 'develop' || network === 'test') {
     try {
       await artifacts.require('RootRegistryV0_1_0').deployed();
     } catch (e) {
--- a/test/helpers/contracts.js
+++ b/test/helpers/contracts.js
@@ -1,5 +1,6 @@
 import { encodeCall } from '../helpers/utils';
 
+const ensUtils = require('./ens');
 const { promisify } = require('util');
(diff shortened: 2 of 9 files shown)


test/helpers/utils.js
```diff
@@ -99,7 +100,32 @@ const giveEth = (toAccount, percentage, fromAccounts) => {
   });
 };
 
+const onlyWhitelisted = async (config, ifWhiteListed) => {
```
pcardune: this doesn't need to be async. And it doesn't need to take a callback function.

test/helpers/utils.js
```diff
@@ -99,7 +100,32 @@ const giveEth = (toAccount, percentage, fromAccounts) => {
...
+      );
+    }
+  } else if (network === 'develop' || network === 'test') {
+    if (!getNamedAccounts(web3).allAccounts.includes(from.toLowerCase())) {
+      throw new Error(
+        `${from} is not a whitelisted account for deploying to ${network}.`
+      );
+    }
+  }
+  return ifWhiteListed(config);
```
pcardune: How about just

```js
return config;
```

test/helpers/multisig.js
```diff
@@ -0,0 +1,48 @@
+const deployOrGetRootRegistry = require('./contracts').deployOrGetRootRegistry;
```
pcardune: You can use object destructuring here.

test/helpers/contracts.js
```diff
@@ -9,6 +10,27 @@ function getLogs(Event, filter, additionalFilters) {
   return promisify(query.get.bind(query))();
 }
 
+const deployOrGetRootRegistry = async config => {
+  const { network, artifacts, deployer } = config;
```
pcardune: You can put this object destructuring directly in the function parameter declaration and it will serve as documentation.

test/helpers/multisig.js
```diff
@@ -0,0 +1,48 @@
...
+
+const prepareMultiSigAndRoot = async config => {
+  const { network, artifacts, accounts } = config;
+  const admin0 = accounts[0];
+  const admin1 = accounts[1];
```
pcardune: use array destructuring

we should copy over the eslint config from the monorepo.

test/helpers/multisig.js
```diff
@@ -0,0 +1,48 @@
...
+      multiAdmin.address,
+      '0_1_0',
+      multiAdmin.address
+    );
+    await rootRegistry.setVersion(
+      'MultiSigWallet',
+      multiSigWallet.address,
+      '0_1_0',
+      multiSigWallet.address
+    );
```
pcardune: Can you add a README.md file somewhere that explains which contracts are in which registry?

package.json
```diff
@@ -22,6 +22,7 @@
     "ethereum-ens": "^0.7.4",
     "ethereumjs-abi": "^0.6.5",
     "ethereumjs-testrpc-sc": "^6.1.2",
+    "glob": "^7.1.2",
```
pcardune: did this just get missed in a previous PR?

test/helpers/utils.js
jaycenhorton: I dont follow, Im passing a function which needs to be executed if the account is whitelisted, and then passing the config into that function

test/helpers/utils.js
jaycenhorton: ? it's taking a function of which to execute, as long as the account is able to do it. Can you propose a more clear solution?

test/helpers/multisig.js
jaycenhorton: sure, creating a ticket

package.json
jaycenhorton: no, it worked fine, but provided an nnoying eslint error without adding it. is there a better eslint rule solution? tech dont need it in deps

test/helpers/utils.js
pcardune: use if statements. Which is better:

```js
const ifCondition = async (doSomethingConfig, doSomething) => {
  someCondition = ...;
  if (someCondition) {
    doSomething(doSomethingConfig);
  }
};
ifCondition("hello", console.log);
```

or

```
const isConditionTrue = () => {
  const someCondition = ...;
  return someCondition;
}
if (isConditionTrue) {
  console.log("hello");
}
```

You've written the first version, I'm suggesting the second.

PR #52, opened 2018-07-10
multi sig migration files for root and contract registry  upgrades 

jaycenhorton:


test/helpers/contracts.js
```diff
@@ -224,7 +225,138 @@ const upgradeToContract = async (
...
 };
 
+const deployOrGetProxy = async (
+  config,
+  existingProxyAddr,
+  contractRegistry,
+  multiAdmin,
+  deployParams
+) => {
+  const { artifacts } = config;
```
pcardune: Whenever I see destructuring of arguments as the first line in a function, I move the destructing to the function parameter declaration.

pcardune: In fact, whenever I have functions that take this many parameters, I use object destructuring for all of them so it's more obvious what's what at the function call sites.

jaycenhorton: I dont folllow the first comment

pcardune: Which is better:

```js
function log(config) {
  const {message, level, error, stack} = config;
  ...
}
```

or

```js
function log({message, level, error, stack}) {
  ...
}
```

My comment is that the second is always preferable to the first.

jaycenhorton: ah yes, thats clear. will modify on a future pr

PR #53, opened 2018-07-12
add all multisig migrations

jaycenhorton:
Needs some tuning, but it works.

### DIFF
--- a/migrations/3_deploy_nori.js
+++ b/migrations/3_deploy_nori.js
@@ -1,101 +1,115 @@
 /* globals artifacts web3 */
-const ENS = require('ethereum-ens');
-const { upgradeToContract } = require('../test/helpers/contracts');
+const bluebird = require('bluebird');
+const {
+  deployOrGetRootRegistry,
+  upgradeAndTransferToMultiAdmin,
+} = require('../test/helpers/contracts');
 const getNamedAccounts = require('../test/helpers/getNamedAccounts');
+const utils = require('../test/helpers/utils');
+const {
+  nori,
+  participantRegistry,
+  crc,
+  participant,
+  supplier,
+  verifier,
+  fifoCrcMarket,
+} = require('../test/helpers/contractConfigs');
 
-const ContractRegistryV0_1_0 = artifacts.require('ContractRegistryV0_1_0');
-const RootRegistryV0_1_0 = artifacts.require('RootRegistryV0_1_0');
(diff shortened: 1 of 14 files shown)


migrations/3_deploy_nori.js
```diff
@@ -1,101 +1,115 @@
...
+const MultiAdmin = artifacts.require('MultiAdmin');
+
+const upgradeAndMigrateContracts = (
+  config,
+  adminAccountAddress,
+  contractsToUpgrade,
+  multiAdmin,
+  registry
+) =>
+  bluebird.mapSeries(contractsToUpgrade, async contractConfig => {
```
pcardune: This is how you would do this without bluebird...
```js
for (contractConfig of contractsToUpgrade) {
  const configuredContract = await contractConfig(multiAdmin, registry);
  ...
}
```

migrations/3_deploy_nori.js
```diff
@@ -1,101 +1,115 @@
...
+    const upgrade = () =>
+      upgradeAndTransferToMultiAdmin(
+        config,
+        configuredContract.contractName,
+        registry,
+        [configuredContract.initParamTypes, configuredContract.initParamVals],
+        { from: adminAccountAddress },
+        multiAdmin
+      );
+    return utils.onlyWhitelisted(config, upgrade);
```
pcardune: This would be so much easier to read as:

```js
if (utils.isWhitelisted(config)) {
  upgradeAndTransferToMultiAdmin(
    config,
    ...
  );
}
```

test/helpers/contractConfigs.js
```diff
@@ -0,0 +1,83 @@
...
+
+module.exports = {
+  nori,
+  participantRegistry,
+  crc,
+  participant,
+  supplier,
+  verifier,
+  fifoCrcMarket,
+};
```
pcardune: This is a nice file.

migrations/3_deploy_nori.js
jaycenhorton: I think I had a for of loop first, but eslint was screaming at me for putting it here. I'm sure bluebirds doing the same thing under the hood though, and so maybe there's an eslint config I can change instead?

test/helpers/contractConfigs.js
jaycenhorton: its neat, but unfortunately a little heavy when using it with production since it has to fetch the latest participant registry for each contract

PR #54, opened 2018-07-14
add helpers to more closely mock test env to prod env

jaycenhorton:
Adds a couple test helpers so that I can more closely mock a real environment when writing tests. These are consumed in a subsequent pr that contains contracts and tests for new fifo/crc functions to get sale totals

### DIFF
--- a/migrations/0_initial_migration.js
+++ b/migrations/0_initial_migration.js
@@ -1,5 +1,6 @@
-module.exports = function deploy(deployer) {
+module.exports = function deploy(deployer, network) {
   deployer.then(async () => {
+    global.network = network;
     // yeah we won't be using this...
   });
 };
--- a/test/helpers/contractConfigs.js
+++ b/test/helpers/contractConfigs.js
@@ -17,6 +17,15 @@ const participantRegistry = async (multiAdmin, registry) => ({
   initParamVals: [registry.address, multiAdmin.address],
 });
 
+const contractRegistry = async (multiAdmin = null, rootRegistry) => ({
+  contractName: 'ContractRegistry',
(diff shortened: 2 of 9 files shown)


migrations/0_initial_migration.js
```diff
@@ -1,5 +1,6 @@
-module.exports = function deploy(deployer) {
+module.exports = function deploy(deployer, network) {
   deployer.then(async () => {
+    global.network = network;
```
pcardune: Is adding this global really necessary? I don't see where it gets used. If it really is necessary, please add a comment on the line above explaining why and where it gets used.

test/behaviors/BasicCommodity.js
```diff
@@ -0,0 +1,167 @@
+// ENABLED IN NEXT PR
```
pcardune: how about just delete this file for this PR?

test/helpers/contractConfigs.js
```diff
@@ -1,83 +1,126 @@
...
-  contractName: 'Participant',
-  initParamTypes: ['address', 'address', 'address'],
-  initParamVals: [
-    registry.address,
-    await registry.getLatestProxyAddr.call('ParticipantRegistry'),
-    multiAdmin.address,
-  ],
-});
+const noriConfig = async (root, artifacts) =>
+  contractRegistryConfig(root, artifacts).then(async contractRegistry => ({
```
pcardune: It's a bit odd to mix and match promise/async like this. Why not:

```js
const contractRegistry = await contractRegistryConfig(root, artifacts);
return {
  ...
};
```

PR #57, opened 2018-07-18
[DO NOT MERGE-- example only]

jaycenhorton:
This is an interesting way to get some really dynamic behavior using the new data fields in contract functions. 

This changes the way sales are created/canceled, by using the authorizeOperator's data field. 

- Create the sale/cancel transaction outside chain
- invoke authorizeOperator, pass in the data
- use execute_call in the market, to execute whichever you want to execute

Problem for this particular feature is that we need explicit canceling/creating so that we can more fluently check and clear permissions/allwoances/balances in the crc contract. There might be a way to do that using this approach, but its not obvious to me if so

Ill probably just create a `revokedOperatorForCommodity` 820 function so that when you revoke operator it calls a completely different function (cancelSale). 

any thoughts on this?

### DIFF
--- a/contracts/market/FifoTokenizedCommodityMarket.sol
+++ b/contracts/market/FifoTokenizedCommodityMarket.sol
@@ -59,8 +59,9 @@ contract FifoTokenizedCommodityMarket is StandardTokenizedCommodityMarket, IEIP7
     if (preventCommodityOperator) {
       revert();
     }
-    require(_executeCall(address(this), 0, operatorData));
+    require(_executeCall(address(this), 0, operatorData)); // use operator as to param?
   }
+
   function _executeCall(address to, uint256 value, bytes data) private returns (bool success) {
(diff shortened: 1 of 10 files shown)


contracts/market/FifoTokenizedCommodityMarket.sol
```diff
@@ -45,29 +45,27 @@ contract FifoTokenizedCommodityMarket is StandardTokenizedCommodityMarket, IEIP7
...
-    //todo jaycen can we figure out how to do this passing in a CommodityLib.Commodity struct (I was having solidity errors but it would be ideal -- might be possible using eternal storage, passing hash of struct and then looking up struct values <-- would be VERY cool)
-    createSale(
-      tokenId, 
-      1, 
-      2, 
-      from, 
-      value, 
-      userData
-    );
+    require(_executeCall(address(this), 0, operatorData)); // use operator as to param?
```
jaycenhorton: and here

PR #66, opened 2018-08-13
[in progress] Add FIFO queue contract for FifoCommodityMarket

rlfarman:


contracts/market/FifoQueue.sol
```diff
@@ -0,0 +1,57 @@
...
+
+  function pop() public returns (uint256 tokenId) {
+    require(
+      last >= first,
+      "There are no items in the queue"
+    );
+
+    for(uint i = last; i < first; i = i.sub(1)) {
+      tokenId = peek();
+      if (tokenId == 0x0) {
```
farmanrl: Here's where we check to see if the element in the mapping still exists (ie: the element hasn't been removed after it was pushed).

contracts/market/FifoQueue.sol
```diff
@@ -0,0 +1,57 @@
...
+  mapping(uint256 => uint256) fifoQueue;
+  uint256 first = 1;
+  uint256 last = 0;
+
+  function push(uint256 tokenId) public {
+    last = last.add(1);
+    fifoQueue[last] = tokenId;
+  }
+
+  function pop() public returns (uint256 tokenId) {
```
farmanrl: `pop()` works by iterating through the length of the FifoQueue. It peeks at the first element, checking to see if it has been removed. If it has, we move the front of the queue. If it hasn't, we delete that item from the queue, move the front of the queue, and then return the tokenId.

contracts/market/FifoQueue.sol
```diff
@@ -0,0 +1,56 @@
...
+  function peek() public view returns (uint256 tokenId) {
+    require(
+      last >= first,
+      "There are no items in the queue"
+    );
+    tokenId = fifoQueue[first];
+    return tokenId;
+  }
+
+  function remove(uint256 tokenId) public {
```
jaycenhorton: So if I'm reading this right:

Create sale(tokenID 0): [0] = 0
Create sale(tokenID 1): [1] = 1
Create sale(tokenID 2): fifoQueue[2] = 2
 
And then when the trade is initiated
buy() : [0] = 0
buy(): ???

I'm not seeing where first is then changes to the next "index" of the mapping. Also, what happens when a user removes a sale in the middle of the queue, (rather than when a buy is invoked which wants to always grab  the first index being removed each time)?

jaycenhorton: something seems to be missing, but maybe I'm not looking in the right spots

farmanrl: Right, so creating a sale will insert a new tokenId in fifoQueue[last], and popping will delete  fifoQueue[first] and return the tokenId that was stored there, iterating `first` by 1.

When we remove an element, we search through the mapping from first to last and delete the match. This leaves a hole in our mapping, however, and to fill that gap and preserve ordering we would need to shift all following elements to the left, which is costly. So instead, we leave the gap, and in `pop()` we check to see if the element we peeked is equal to 0. If it is, we move the front of the queue and check again. So we skip over any deleted entries.

So `first` is iterated when we `pop()`, `last` is iterated when we push.

I think I'll change `peek()` to do the empty check, and `buy()` to use `pop()` instead of `tokenid = peek()` and then `remove(tokenId)`. That might make things a little more clear.

jaycenhorton: But Im not seeing where `pop` is called

jaycenhorton: Oh, I see your comment, Im guessing this is more a "which should I use" pr?

contracts/market/FifoQueue.sol
```diff
@@ -0,0 +1,56 @@
+
+pragma solidity ^0.4.24 ;
+import "../../node_modules/zeppelin-solidity/contracts//math/SafeMath.sol";
+
+contract FifoQueue {
```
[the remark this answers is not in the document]
farmanrl: I anticipated we might want FIFO behavior for more than our commodity markets.

contracts/market/FifoQueue.sol
farmanrl: There is a bug here, since `peek()` doesn't have the behavior of checking zero equality. I'm submitting this change now.

jaycenhorton: I thought there might be. This is good progress though, very creative and I like where it's headed

PR #67, opened 2018-08-16
add reasons for all the revert and require statements

jaycenhorton:
No changes in here except some typos and the addition of reason strings for all the reverts and require.


jaycenhorton:
[Add "reason strings" to all reverts so we can use them in truffle 5 (beta at truffle@next)](https://trello.com/c/gOBIZ4LJ/477-add-reason-strings-to-all-reverts-so-we-can-use-them-in-truffle-5-beta-at-trufflenext)

PR #68, opened 2018-08-21
 refactor to use erc820 submodule

jaycenhorton:
moving the EIP820 into a git submodule better allows us to stay in sync with the latest standards as well as to contribute directly to it.

This isn't quite done, still a bit more cleanup to do. Additionally, I think it might be best to move the actual submodule into this directory rather than the root of the nori monorepo


jaycenhorton:
Submodule integration successful, now just needs some clean up

jaycenhorton:
Ok, I think this is ready for review

package.json
```diff
@@ -18,16 +18,20 @@
     "eslint-config-airbnb-base": "12.1.0",
     "eslint-config-prettier": "^2.9.0",
     "eslint-plugin-import": "^2.13.0",
+    "eth-contract-class": "^0.0.12",
```
farmanrl: A few of these packages don't seem to be used anywhere

jaycenhorton: theyre used in the submodule

farmanrl: bitmoji

PR #69, opened 2018-08-22
[In progress] add ability to pause contracts

jaycenhorton:
Not ready yet, but putting this here before switching contexts to a COMET related task. Since this one is going to require base modifications to the registry (when we make that pausable), I'm going to wait to merge this + #67 (and maybe #66 if @farmanrl  finishes before I return to this?). 

Will merge all of those together and do a single version bump + ropsten migration.  There shouldn't be any dependencies that are waiting on these three PRs.


jaycenhorton:
Additional note: the Nori Token is the only contract requiring this type of implementation. All others will just require a change to the pase contract (swapping UnstructuredOwnable for Pausable, and then adding the modifier to each function in the base). The reason the Nori contract reqs an implementation like this is because soon the 777 folder will be submoduled for the official repo (WIP), and we don't want to propose this type of pausability in the official standard.

jaycenhorton:

PR #70, opened 2018-08-25
call toggles using admin

jaycenhorton:
This adds a helper function and changes to some tests that don't do much different other than call those functions using the multiAdmin contract. (this is a requisite fix for the subsequent PR I will submit which adds the remaining pausable contracts (and onlyOwner modifiers when the function is still callable by admin instead of pausing)


jaycenhorton:

PR #71, opened 2018-08-25
add pausable contract funcs and onlyOwner

jaycenhorton:
adds pausable and onlyOwner modifiers to the remaining contract functions.


jaycenhorton:

PR #76, opened 2018-09-22
add the first version of the risk mitigation account idea

jaycenhorton:
This adds the risk mitigation account contract

Currently it only allows tokens to be restricted and deposited into the account, but it will be trivial to withdraw them once we figure out the real-world logic surrounding that idea.

This WIP implementation takes a first stab at calculating a restricted/unrestricted balance using the following logic:

* assume a CRC is minted and verified with a 50% score.
* the CRC is listed, and sold, in the market
* The supplier receives 50 tokens
* the restricted account increments by 50 tokens
* WIP the supplier should be able to submit some proof to withdraw the restricted balance

There are more contingencies not accounted for here, such as the logic around what happens when a CRC is found to remove less than the attested amount and how the restricted tokens can be used to make up for those by purchasing new ones. That said, it should at least give us enough to think more about concept.

For now, best to just assign 100 as the verification score, and then everything will work as it was previously (since 100% of the tokens will be unrestricted)


migrations/6_deploy_test_contracts.js
```diff
@@ -19,17 +19,17 @@ module.exports = function deploy(deployer, network, accounts) {
...
-    );
-    await crcMarket.initialize(
-      registry.address,
-      [
-        await registry.getLatestProxyAddr.call('CRC'),
-        await registry.getLatestProxyAddr.call('Nori'),
-      ],
-      accounts[0]
-    );
+    // const crcMarket = await deployer.deploy(
```
farmanrl: Any reason why this, and the other parts of the testing suite for SelectableCrcMarket, are commented out?

jaycenhorton: Yeah, I talked out the selectable market contracts as they're not relevant to the main app. I left them commented out as a note to myself to make sure to copy the tests over to the atomic-market repo as they're still relevant there. This still needs some clean up before merging, I just opened it to give an idea of how it might work 😁

jaycenhorton:
ok, if tests pass this is ready

PR #77, opened 2018-10-09
use the official 777 repo

jaycenhorton:
Although this looks like a lot of changes, a lot is largely unchanged and this is more or less just a refactor. 

The biggest change is that this should more-or-less finalize the NORI token as it grabs the latest eip777 version, and cleans up quite a bit of old todos


jaycenhorton:
there still appears to be an issue with the change in how solc links files (works locally, but there seems to need to be some config changes to the circle image)

Will look into this tomorrow

jaycenhorton:
[move eip777 contracts to a gitsubmodule so we can contribute to the offical standard](https://trello.com/c/vTV99ahB/479-move-eip777-contracts-to-a-gitsubmodule-so-we-can-contribute-to-the-offical-standard)

PR #102, opened 2019-08-11
rewrite contracts from scratch; rewrite upgrade pattern checkpoint

jaycenhorton:
This PR is a ground-up rewrite. 

The original contracts were written long ago with entirely different design patterns in mind. This is an attempt to remove all the unnecessary noise that exists from those old designs. 

# Things that will be removed include:

## Marketplace contracts: 
these are no longer necessary based on the latest design of the marketplace. Since we are retiring CRCs and minting CRCs post verification + sale (at least for the first transaction), there is no need for a complex atomic swap marketplace

## Commodity contracts:
the commodity contracts were created before there was standardization of the ERC 721 NFT. This allows us to drop custom logic and instead just use extremely simple overriding logic to disable transfers post minting

## RootRegistry and ContractRegistry
These were a complex idea that was necessary due to how we wrote out custom migrations to hijack truffle migrations and use an entirely custom upgrade framework. This is no longer necessary as zeppelin os appears to have achieved maturity. In addition, ENS subdomains can be used in the future to achieve something similar.

## Upgrade contracts
See notes about removing custom upgrade logic throughout this message. In addition to removing custom upgrade contracts (Proxy, UnstructuredUpgradeabilityProxy, etc.), the custom ownable contracts, and complex usage of ERC 820 (introspection registry) and ERC 777 (advanced token standard for use with atomic swaps) will likely be removed as well.

## Various others
Other contracts (such as the participant contracts, the NORI token-- which might now instead be a simple ERC 20 instead of an ERC 777--, and the CRC contract) will also be drastically simplified. Custom lifecycle contracts such as the Pausable contract will be replaced with an official openzeppelin implementation. Custom ENS contracts will also be replaced by official implementations. 

# Misc changes

- Along with removing old implementations, this PR includes the scrapping of the custom upgrade pattern which was necessary at the time but can be deprecated in favor of a well-maintained upgrade framework (zeppelin os). This means a rewrite of the migrations to include a drastically dumbed down version will eventually be useful. Until then upgrades will be managed manually. 

- Tests are rewritten

- This also includes an upgrade to web3@1, truffle@5, and solidity@0.5


contracts/CRC.sol
```diff
@@ -0,0 +1,53 @@
+pragma solidity ^0.5.0;
+
+import "@openzeppelin/contracts-ethereum-package/contracts/token/ERC721/ERC721Full.sol";
+import "@openzeppelin/contracts-ethereum-package/contracts/token/ERC721/ERC721Pausable.sol";
+
+contract CRC is Initializable, ERC721, ERC721Enumerable, ERC721Metadata, ERC721Pausable {
```
jaycenhorton: an example of a dead-simple rewrite of the CRC contract. Verification data can be included in the `tokenURI` param of the `mintAndRetire` function. 

From 1,000+ custom lines of solidity to just 50.

jaycenhorton:
@rlfarman  This is ready for review. It will not be ready for merge however until I remove any breaking changes in the monorepo.

jaycenhorton:
[Deploy CRC](https://trello.com/c/HTMRQaGx/1042-deploy-crc)

PR #136, opened 2021-12-14
WIP migrate nccr to erc 1155; add fifo contract

jaycenhorton:


contracts/v1/Certificate.sol
```diff
@@ -0,0 +1,114 @@
...
+  /**
+   * @dev returns the removal IDs and the amounts of the sources
+   */
+  function sources(uint256 certificateId) public view returns (Source[] memory) {
+    return _sources[certificateId];
+  }
+
+  /**
+   * @dev mints the certificate (using a batch of certificate sources)
+   * @param to The supplier address
```
amiecorso, line 55: `to` is actually the buyer address right?

PR #137, opened 2021-12-17
Support contracts V0 alongside new on-chain market contracts NO-1380

jaycenhorton:
- integrates legacy contracts with new contracts
- stop using the deprecate openzeppelin cli tool
- migrate to the openzeppelin hardhat upgrades plugin
- migrate to hardhat
- stop using truffle


deploy/0_deploy_contracts.ts
```diff
@@ -0,0 +1,165 @@
...
+        ethernal.push({
+          name: 'Certificate',
+          address: certificateInstance.address,
+        }),
+        ethernal.push({
+          name: 'FIFOMarket',
+          address: fifoMarketInstance.address,
+        }),
+      ]);
+      console.log('Registered contracts in Ethernal', buyer);
```
amiecorso, line 115: why second `buyer` arg to this log?

test/FIFOMarket.test.ts
```diff
@@ -0,0 +1,47 @@
...
+    it('should purchase removals and create a certificate', async () => {
+      const {
+        contracts: { NORI, Removal, Certificate, FIFOMarket },
+      } = await setupTest();
+      expect(NORI.address).to.be.properAddress;
+      expect(Removal.address).to.be.properAddress;
+      expect(Certificate.address).to.be.properAddress;
+      expect(FIFOMarket.address).to.be.properAddress;
+    });
+  });
```
amiecorso, line 46: should we include a test about what happens if there aren't enough removals to fulfill the order?  or is it just too obvious that the transaction will revert and it's not interesting?

amiecorso (approved):
just left two little comments but otherwise approving!  can't claim to understand this 100% but I definitely get the gist and took a look at every file.

hardhat.config.ts
```diff
@@ -0,0 +1,109 @@
...
+task('accounts', 'Prints the list of accounts', async (taskArgs, hre) => {
+  const accounts = await hre.ethers.getSigners();
+  accounts.forEach((account) => {
+    console.log(account.address);
+  });
+});
+
+task(
+  'ethernal:reset',
+  'Prints the list of accounts',
```
scottynomad, line 25: stale docs

scottynomad (approved):

test/FIFOMarket.test.ts
jaycenhorton: we should definitely add tests for that!

For now I'm opting to handle those incrementally as we work through more requirements

PR #147, opened 2022-01-26
Update FIFOContract to revert when queue is empty, add more test cases

amiecorso:
When testing checkout on staging, a completely empty marketplace was still minting a certificate... Indicating something wrong. 
This PR adds a couple new tests and proposes some fixes for bugs in the FIFOMarket.

Certificate minting was happening because the FIFOMarket wasn't reverting when the queue was completely empty because the for loop that has all the other reversion statements in it simply never ran at all.... so the loop would get skipped, the certificate would get minted, and then nothing would get sent to the supplier or nori wallet because the `amounts` array was empty, thus leaving the contract with the buyer's NORI balance.

To handle this case, I added a check for emptiness on the `amounts` array (could have used `ids` or `suppliers` too) instead of making an initial check to `numberOfTonnesInQueue` because a call to that function seems unnecessarily gas intensive. If `amounts` is empty, the for loop never ran, so we revert bc the market is empty.

Next, I added a test that tried to make a purchase that spanned and used up 3 different removals.  This revealed a bug that helped me realize that the `EnumerableUpgradableSet` isn't the right data structure to be using for the FIFOMarket `_queue`, so I'm trying out a different implementation of that (see contract).
EnumerableUpgradableSet isn't the right option for two reasons:
1) It doesn't make any guarantee of ordering... So the thing you inserted at index i won't necessarily always stay at index i as the set is mutated, making this inappropriate for a true FIFO queue where order matters.
2) relatedly, the bug I discovered was in the code that removes used-up removals from the queue.  First, there was a classic issue of trying to index into a mutable data structure that you're changing at the same time... such that by the time your loop is at index 2 (for example), you've been removing things from your queue and it only has 2 items in it, making index 1 the highest available index and it throws.
In the case of a true queue, this would be solvable by always removing from index 0, since that is the head (like if this was being implemented with a linked list or something).  But because we're using a (weird) SET, and not a queue, you can't solve it this way because of the aforementioned issue with changing indexes.  Such that as you remove removals from the set, if you always pull from index 0, there's no guarantee that you're even removing the right removal.  Additionally, always removing from 0 still seemed to throw even in some cases where the set was not yet empty.

There may be a better way to do this, but I've experimented with using an int->int map that is serving as a queue by way of tracking the current head and tail of the queue and using the map keys as "indices".  I briefly looked at dynamic arrays in solidity but wasn't sure they'd behave the way I want them to behave... maybe something to look into more and to compare the gas costs with this proposed implementation! But I wanted to propose this option too... the tests pass.


scottynomad (commented):
Nice find @amiecorso !  Do you have a lead on adjusting the linter settings?  It's super hard to review like this. :/

.vscode/settings.json
```diff
@@ -0,0 +1,69 @@
+{
```
amiecorso, line 1: added the same settings config to this repo so that it matches what's going on in the monorepo and we don't get linter whiplash if working directly in submodule?

amiecorso:
I think I took care of the linter noise!

contracts/FIFOMarket.sol
```diff
@@ -50,14 +50,26 @@ contract FIFOMarket is
...
 
-  function numberOfTonnesInQueue() public view returns (uint) {
-    uint tonnesInQueue = 0;
-    for(uint i = 0; i < _queue.length(); i++) {
-      tonnesInQueue += _removal.balanceOf(address(this), _queue.at(i));
+  function _queueLength() private view returns (uint256) {
+    return _queueNextInsertIndex - _queueHeadIndex;
+  }
+
+  function numberOfTonnesInQueue() public view returns (uint256) {
```
scottynomad, line 69: NRTs?

contracts/FIFOMarket.sol
```diff
@@ -87,65 +100,81 @@ contract FIFOMarket is
...
+      );
+      address supplier = _removal.vintage(_queue[i]).supplier;
+      if (remainingAmountToFill < removalAmount) {
+        ids[i] = _queue[i];
         amounts[i] = remainingAmountToFill;
         suppliers[i] = supplier;
         remainingAmountToFill = 0;
-      } else if(remainingAmountToFill >= removalAmount) {
-        if(i == _queue.length() - 1 && remainingAmountToFill > removalAmount){
+      } else if (remainingAmountToFill >= removalAmount) {
```
scottynomad, line 125: Can we remove the `if` here?  It's true by definition given the preceding `if` clause.  This has the potential to cause bugs / confusion later.

contracts/FIFOMarket.sol
```diff
@@ -87,65 +100,81 @@ contract FIFOMarket is
...
+            remainingAmountToFill > removalAmount
+        ) {
           revert("FIFOMarket: Not enough supply");
         }
-        ids[i] = _queue.at(i);
+        ids[i] = _queue[i];
         amounts[i] = removalAmount;
         suppliers[i] = supplier;
         remainingAmountToFill -= removalAmount;
       } else {
```
scottynomad, line 136: Isn't this unreachable?

scottynomad (approved):
I think your changes to the data structure make sense.  I'm struggling to read the big nested if in `tokensReceived` so apologies if some of those comments don't make sense.  Maybe we can figure out a way to structure this a bit more clearly.

contracts/FIFOMarket.sol
amiecorso: nope - only if the above if statement leads to the revert, otherwise it'll execute

contracts/FIFOMarket.sol
```diff
@@ -96,55 +100,62 @@ contract FIFOMarket is
...
     uint256 remainingAmountToFill = certificateAmount;
-    uint256[] memory ids = new uint256[](_queue.length());
-    uint256[] memory amounts = new uint256[](_queue.length());
-    address[] memory suppliers = new address[](_queue.length());
-    for (uint256 i = 0; i < _queue.length(); i++) {
-      uint256 removalAmount = _removal.balanceOf(address(this), _queue.at(i));
-      address supplier = _removal.vintage(_queue.at(i)).supplier;
+
+    address recipient = abi.decode(userData, (address)); // todo handle the case where someone invokes this function without operatorData
+    require(recipient == address(recipient),"FIFOMarket: Invalid address");
```
jaycenhorton, line 107: noticing some inconsistency in formatting-- can you make sure prettier ran?

contracts/FIFOMarket.sol
```diff
@@ -25,13 +23,13 @@ contract FIFOMarket is
...
-  using EnumerableSetUpgradeable for EnumerableSetUpgradeable.UintSet;
-
   IERC1820RegistryUpgradeable private _erc1820;
   Removal private _removal;
   Certificate private _certificate;
   NORI private _nori;
-  EnumerableSetUpgradeable.UintSet private _queue;
+  mapping(uint256 => uint256) private _queue;
+  uint256 private _queueHeadIndex;
+  uint256 private _queueNextInsertIndex;
```
jaycenhorton, line 32: why do we need to store both `_queueHeadIndex` and `_queueNextInsertIndex`?

contracts/FIFOMarket.sol
```diff
@@ -78,16 +82,16 @@ contract FIFOMarket is
     bytes memory
   ) public override returns (bytes4) {
     for (uint256 i = 0; i < ids.length; i++) {
-      _queue.add(ids[i]);
+      _queue[_queueNextInsertIndex] = (ids[i]);
+      _queueNextInsertIndex++;
```
jaycenhorton, line 86: is it possible to just push this?

contracts/FIFOMarket.sol
```diff
@@ -96,55 +100,62 @@ contract FIFOMarket is
...
-        revert("FIFOMarket: Not enough supply");
       }
+
       if (remainingAmountToFill == 0) {
         break;
       }
     }
+    // In a completely empty queue, the for loop never runs.
+    // Catch this case here without the gas-heavy numberOfNrtsInQueue call.
+    require(amounts[0] > 0, "FIFOMarket: Not enough supply");
```
jaycenhorton, line 144: might be better to check this before the for loop?

contracts/FIFOMarket.sol
amiecorso: discussed in OH about how we're using a map not array

contracts/FIFOMarket.sol
amiecorso: same about map vs array

contracts/FIFOMarket.sol
amiecorso: actually, with this different implementation of the queue I think I can change the way this check is made. the queue is empty if the head index and the next insert index are the same.  I'll add this check before the loop, I think it's probably a tiny bit gas cheaper that way and clearer

PR #148, opened 2022-01-27
Added updated LockedNori with tests.

scottynomad:
Knows TODOs:
* [x] Revoke vesting balances without needing an allowance from the recipient (defaultOperators)
* ~[ ] Pass a NORI wallet address in the initializer as the destination for the clawed back lNORI~ 
* [x] Enforce no transfer / send of LockedNori that is still locked
* [x] Expose `grantTo` publicly to cover the post-launch restricted tokens case when finding occurs gradually.
* [x] Enforce RBAC on 777 functions (mint, burn, etc..)
* [x] Update deploy script
* [x] Expand docs


contracts/LockedNORI.sol
```diff
@@ -40,127 +63,339 @@ contract LockedNORI is ERC777Upgradeable, ERC20PresetMinterPauserUpgradeable, ER
...
+
+  /**
+   * @dev Allow a user to burn a number of wrapped tokens and withdraw the corresponding number of _underlying tokens.
+   */
+  function withdrawTo(address account, uint256 amount)
+    public
+    virtual
+    returns (bool)
+  {
+    TokenGrant storage grant = grants[account];
```
jaycenhorton, line 130: does this need to check that they can claim the amount specified or is that happening somehow elsewhere?

e.g.,

```solidity
require(grant.originalAmount - grant.claimedAmount >= amount, "LockedNORI: insufficient remaining token amount");
```

contracts/LockedNORI.sol
```diff
@@ -40,127 +63,339 @@ contract LockedNORI is ERC777Upgradeable, ERC20PresetMinterPauserUpgradeable, ER
...
-    uint256 cliff2Time,
-    uint256 vestCliff1Amount,
-    uint256 vestCliff2Amount,
-    uint256 unlockCliff1Amount,
-    uint256 unlockCliff2Amount
-  ) external onlyRole(TOKEN_GRANTER_ROLE) whenNotPaused {
+    bytes calldata userData,
+    bytes calldata operatorData
+  ) internal returns (bool) {
+    (address recipient, uint256 startTime) = abi.decode(userData, (address, uint256));
```
jaycenhorton, line 103: can this be destructured in a way similar to what you're doing [here](https://github.com/nori-dot-eco/contracts/pull/148/files#diff-f7b5af71dc8daef37251508c0bf68958b31af1e685419b93673f97c9ded59340R151)?

contracts/LockedNORI.sol
jaycenhorton: also, should it be updating the `grant.grantAmount` ?

contracts/LockedNORI.sol
```diff
@@ -40,127 +63,339 @@ contract LockedNORI is ERC777Upgradeable, ERC20PresetMinterPauserUpgradeable, ER
...
     grant.exists = true;
-    grant.vestingSchedule.totalAmount = amount;
-    grant.vestingSchedule.startTime = startTime;
-    grant.vestingSchedule.endTime = vestEndTime;
-    // grant.vestingSchedule.addCliff(cliff1Time, vestCliff1Amount);
-    // grant.vestingSchedule.addCliff(cliff2Time, vestCliff2Amount);
+
+    if (params.vestEndTime > params.startTime) {
+      require(params.vestCliff1Amount >= params.unlockCliff1Amount
+        && params.vestCliff2Amount >= params.unlockCliff2Amount,
```
jaycenhorton, line 167: should this be `||`?

contracts/LockedNORI.sol
```diff
@@ -40,127 +63,339 @@ contract LockedNORI is ERC777Upgradeable, ERC20PresetMinterPauserUpgradeable, ER
...
+    uint256 startTime,
+    uint256 vestEndTime,
+    uint256 unlockEndTime,
+    uint256 cliff1Time,
+    uint256 cliff2Time,
+    uint256 vestCliff1Amount,
+    uint256 vestCliff2Amount,
+    uint256 unlockCliff1Amount,
+    uint256 unlockCliff2Amount
+  ) public whenNotPaused onlyRole(TOKEN_GRANTER_ROLE) {
```
jaycenhorton, line 212: are we currently setting `TOKEN_GRANTER_ROLE` somewhere?

jaycenhorton: thinking this should be `external`

jaycenhorton: along with several other fns

contracts/LockedNORI.sol
```diff
@@ -40,127 +63,339 @@ contract LockedNORI is ERC777Upgradeable, ERC20PresetMinterPauserUpgradeable, ER
...
     TokenGrant storage grant = grants[account];
-    return grant.grantAmount;
+    return grant.originalAmount - grant.grantAmount;
   }
 
+  /**
+   * @dev vestedBalanceOf: Vested balance less any claimed amount
+   */
   function vestedBalanceOf(address account) public view returns (uint256) {
+    return _vestedBalanceOf(block.timestamp, account);
```
jaycenhorton, line 273: we're pretty inconsistent with this pattern of using an internal `_functionName`)

contracts/LockedNORI.sol
```diff
@@ -40,127 +63,339 @@ contract LockedNORI is ERC777Upgradeable, ERC20PresetMinterPauserUpgradeable, ER
...
     TokenGrant storage grant = grants[account];
     if (grant.exists) {
-      // Past the end date user can claim any remaining wrapped tokens
-      return
-        grant.vestingSchedule.availableAmount(block.timestamp) -
-        grant.claimedAmount;
+      if (grant.vestingSchedule.startTime > 0) {
+        // If any tokens have been revoked then the schedule (which doesn't get updated) could
+        // return more than the total grant amount.
+        return
```
jaycenhorton, line 289: this is subjective, but I personally find multiple-return functions harder to read since I have to look for a return value in more than one place

contracts/LockedNORI.sol
```diff
@@ -40,127 +63,339 @@ contract LockedNORI is ERC777Upgradeable, ERC20PresetMinterPauserUpgradeable, ER
...
+          MathUpgradeable.min(grant.grantAmount, grant.vestingSchedule.availableAmount(atTime)) - grant.claimedAmount;
```
jaycenhorton, line 290: can you break down why we need to take the minimum of these two instead of just the latter?

contracts/LockedNORI.sol
```diff
@@ -40,127 +63,339 @@ contract LockedNORI is ERC777Upgradeable, ERC20PresetMinterPauserUpgradeable, ER
...
-        wrappedBalance
-      );
-      require(amount <= availableBalance, "Withdrawl amount unavailable");
+  /**
+   * @dev getGrant: Returns all governing settings on a grant.
+   */
+  function getGrant(address account)
+    public
+    view
+    returns (
```
jaycenhorton, line 339: I wonder if we should just return the struct?

contracts/LockedNORI.sol
```diff
@@ -40,127 +63,339 @@ contract LockedNORI is ERC777Upgradeable, ERC20PresetMinterPauserUpgradeable, ER
...
+      grant.vestingSchedule.cliffs[0].amount,
+      grant.vestingSchedule.cliffs[1].amount,
+      grant.lockupSchedule.cliffs[0].amount,
+      grant.lockupSchedule.cliffs[1].amount,
+      grant.claimedAmount,
+      grant.originalAmount
+    );
+  }
+
+  function _isTokenGranter(address account) internal view returns (bool) {
```
jaycenhorton, line 371: what is this and why is it needed in addition to `hasRole(TOKEN_GRANTER_ROLE)`

contracts/LockedNORI.sol
```diff
@@ -40,127 +63,339 @@ contract LockedNORI is ERC777Upgradeable, ERC20PresetMinterPauserUpgradeable, ER
...
     uint256 amount
   ) internal override {
-    if (to == address(0)) {
-      TokenGrant storage grant = grants[from];
-      grant.claimedAmount += amount;
+    bool isNotMinting = from != address(0); // if it's not minting, then we check balances
+    // this is a burn / unwrap
+    if (isNotMinting) {
+      uint256 availableBalance = unlockedBalanceOf(from);
+      if(!hasRole(DEFAULT_ADMIN_ROLE, _msgSender())){
```
jaycenhorton, line 393: I added this for revoking, but I wonder if it's the best approach

contracts/LockedNORI.sol
```diff
@@ -171,39 +406,93 @@ contract LockedNORI is ERC777Upgradeable, ERC20PresetMinterPauserUpgradeable, ER
...
+    virtual
+    override(ERC20Upgradeable, ERC777Upgradeable)
+    returns (bool)
+  {
     return ERC777Upgradeable.transferFrom(holder, recipient, amount);
   }
-}
\ No newline at end of file
+
+  //   function _burn(
```
jaycenhorton, line 490: rm?

contracts/LockedNORI.sol
scottynomad: Good idea. Fixed.

contracts/LockedNORI.sol
scottynomad: You're right.  Good catch.

contracts/LockedNORI.sol
scottynomad: We were only assigning that role in the tests. I updated the initialized to give it to the admin as well.  Also switched to external along with others where it made sense.

contracts/LockedNORI.sol
scottynomad: I just updated `unlockedTokenBalance` to follow this pattern.  Is it worth doing it throughout?

scottynomad: Applied the pattern to `revokeUnvestedTokens` as well.

jaycenhorton: I don't have a strong opinion

contracts/LockedNORI.sol
scottynomad: Yeah, I certainly agree for long functions. I'm not opposed to updating these to use a local variable. I guess readability is top priority in this kind of code!

contracts/LockedNORI.sol
scottynomad: In order to change the behavior of the lockup to the way Paul wants it to work -- unlock continues at the original rate after a revocation, I changed the implementation to shrink only `grant.grantAmount` (leaves the schedules untouched). So if `vestedBalanceOf` was being called after a revocation but before the end date the schedule amount would be greater than the total grant amount.

All that said that behavior is only relevant for unlocking not vesting and I can see why it's confusing.  I'll update it to shrink both `grant.grantAmount` and `grant.vestingSchedule.amount` and remove this check.  It's more explicit that way.

contracts/LockedNORI.sol
scottynomad: removed

contracts/LockedNORI.sol
scottynomad: I defined a struct for the return type at least to document it better.

contracts/LockedNORI.sol
scottynomad: I looked into inverting this logic but it makes it more confusing to rely on `remainingBalance` given how the schedule library is designed.

contracts/LockedNORI.sol
scottynomad: I think we need it due to the operatorSend in `revokeUnvestedTokens` but should probably be tracking granter role rather than admin.  That also means all granters need to be operators I guess?

scottynomad: I added an `it.todo` to test this.

jaycenhorton (approved):

PR #150, opened 2022-01-28
use send function to deposit + grant tokens in single tx

jaycenhorton:
NO 1096


scottynomad (commented):
If I'm reading this correctly, this is making the bundling of the granting parameters into a send the only way to set up a grant?  

We still have the one usecase (post-launch restricted tokens) where we need to be able to set up a grant without funding it and then the tokens are trickled in with each transaction.

contracts/LockedNORI.sol
```diff
@@ -163,8 +223,16 @@ contract LockedNORI is ERC777Upgradeable, ERC20PresetMinterPauserUpgradeable {
       emit UnvestedTokensRevoked(atTime, from, quantityRevoked);
       // TODO: destination address for clawed back tokens should be a role
```
scottynomad, line 224: What do you think about this TODO?  Worth doing? I guess it's more flexible this way.  Maybe we should just burn the lNORI and `operatorSend` the actual NORI?

test/LockedNORI.test.ts
```diff
@@ -1,337 +1,428 @@
...
+      'TokenGrantCreated'
+    );
+    return { nori, lNori, hre };
+  }
+);
+
+describe('LockedNori', () => {
+  // todo
+  // it.todo('test that the admin cannot revoke vested tokens')
+  // todo should revoking redeem the NORI too?
```
scottynomad, line 160: I'm thinking so now.

scottynomad (approved):
Looks good, the tests are certainly more readable with this approach setting the next block timestamp.

PR #152, opened 2022-01-29
Removals.sol emit `BatchMinted` event, set admin approval on minting, list while minting

amiecorso:
Corresponding monorepo PR: https://github.com/nori-dot-eco/nori/pull/2363

NO-1401

This PR makes several updates to `Removal.sol` and adds some corresponding tests.
Emits a `BatchMinted` event with an array of tokenIds as a return value which can be used to assign the correct tokenId to the removal entities that were minted.

There is also the addition of `_setApprovalForAll` and a call to it in `mintBatch` that sets the Nori admin address as an approved operator for all of those supplier's tokens.  May be able to improve this to be a one-time thing at contract deployment instead of an every-time call, which will make minting a little more gas expensive.

Finally, there is another piece of data passed to `mintBatch` in the call data, a boolean which determines whether to list the freshly minted removal tokens.  This allows for atomically listing while minting.


scottynomad (commented):
Looks good!  This is a good addition!

contracts/Removal.sol
```diff
@@ -47,11 +49,24 @@ contract Removal is ERC1155PresetMinterPauserUpgradeable, ERC1155SupplyUpgradeab
   }
 
   /**
-   * @dev mints all of the removal vintages for an issuance
+   * @dev returns the token ids for a set of removals given each one's vintage and the data that was provided
+   * in its mint transaction
+   */
+  function tokenIdsForRemovals(bytes[] memory parcelIdentifiers, uint256[] memory removalVintages) public view returns (uint256[] memory) {
```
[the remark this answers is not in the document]
amiecorso: called!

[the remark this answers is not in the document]
amiecorso: (using a require so the behavior can't end up being undefined in the case the arrays aren't the same length)

contracts/Removal.sol
```diff
@@ -26,12 +26,14 @@ contract Removal is
...
     address supplier;
     uint16 vintage;
+    bytes32 uniqueId;
     // todo: location
     // todo: methodology
     // todo: supplier name
   }
 
   mapping(uint256 => Vintage) private _vintages;
+  mapping(bytes32 => uint256) private _vintageTokenIdMap;
```
jaycenhorton, line 36:
```suggestion
  mapping(bytes32 => uint256) private _tokenIdToVintage;
```

contracts/Removal.sol
```diff
@@ -43,12 +45,56 @@ contract Removal is
...
+   * @dev returns the token ids for a set of removals given each one's vintage and the data that was provided
+   * in its mint transaction
+   */
+  function tokenIdsForRemovals(
+    bytes32[] memory parcelIdentifiers,
+    uint256[] memory removalVintages
+  ) public view returns (uint256[] memory) {
+    require(
+      parcelIdentifiers.length == removalVintages.length,
+      "parcelIdentifers and removalVintages must be same length"
```
jaycenhorton, line 64:
```suggestion
      "Removal: ..."
```

Although it seems weird to have a revert in a public view function.

contracts/Removal.sol
```diff
@@ -43,12 +45,56 @@ contract Removal is
...
+  }
+
+  /**
+   * @dev See {IERC1155-setApprovalForAll}.
+   */
+  function setApprovalForAllAsAdmin(
+    address owner,
+    address operator,
+    bool approved
+  ) public virtual onlyRole(DEFAULT_ADMIN_ROLE) {
```
jaycenhorton, line 82:
```suggestion
  ) public virtual onlyRole(DEFAULT_ADMIN_ROLE) whenNotPaused {
```

contracts/Removal.sol
```diff
@@ -57,22 +103,29 @@ contract Removal is
     bytes memory data
   ) public override {
     // todo require vintage is within valid year range and doesn't already exist
+    (address market, bytes32 parcelId, bool listNow) = abi.decode(
+      data,
+      (address, bytes32, bool)
```
scottynomad, line 108: Can we define a struct to describe this 3-tuple?  Similar to this:  https://github.com/nori-dot-eco/contracts/blob/master/contracts/LockedNORI.sol#L251

scottynomad (commented):
Looks good, just a minor comment and :+1: to Jaycen's

contracts/Removal.sol
amiecorso: this is actually mapping from information about the vintage (but not from an actual vintage struct because that can't be a map key) TO the tokenId.. so I'll update the name to `_vintageToTokenIdMap`

contracts/Removal.sol
amiecorso: I think you're right - I did a little research and it sounds like while this would work as expected in Remix or using a local ganache chain, it won't actually revert on a public network.  I pulled this out and instead added to the documentation to make it clear that these arrays should be the same length because they correspond to the same group of removals.

scottynomad (commented):
Looks good!  Some error handling suggestions in here ...

contracts/Removal.sol
```diff
@@ -64,16 +107,37 @@ contract Removal is
...
+      data,
+      (BatchMintRemovalsData)
+    );
+
     uint256[] memory ids = new uint256[](vintages.length);
     for (uint256 i = 0; i < vintages.length; i++) {
+      bytes32 uniqueIdHash = keccak256(
+        abi.encodePacked(decodedData.uniqueId, vintages[i])
+      );
+
```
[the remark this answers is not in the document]
amiecorso: moot

contracts/Removal.sol
```diff
@@ -42,20 +50,55 @@ contract Removal is
...
+   * @dev returns the token ids for a set of removals given each one's vintage and the identifying data that was provided
+   * in its mint transaction. Arguments should be arrays of the same length as they correspond to the same removals.
+   * @param uniqueIdentifiers The unique identifier that was passed at the time of minting each removal.
+   * @param removalVintages The vintage (year) of each removal.
+   */
+  function tokenIdsForRemovals(
+    bytes32[] memory uniqueIdentifiers,
+    uint256[] memory removalVintages
+  ) public view returns (uint256[] memory) {
+    uint256[] memory ids = new uint256[](removalVintages.length);
```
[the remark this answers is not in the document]
amiecorso: moot

contracts/Removal.sol
```diff
@@ -42,20 +50,55 @@ contract Removal is
...
+    for (uint256 i = 0; i < removalVintages.length; i++) {
+      ids[i] = _vintageToTokenIdMap[
```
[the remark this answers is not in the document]
amiecorso: moot - tearing this out in favor of using events to assign the correct tokenId to a removal entity at time of minting

contracts/Removal.sol
```diff
@@ -42,20 +50,55 @@ contract Removal is
...
+        keccak256(abi.encodePacked(uniqueIdentifiers[i], removalVintages[i]))
+      ];
+    }
+    return ids;
+  }
+
+  /**
+   * @dev See {IERC1155-setApprovalForAll}.
+   */
+  function setApprovalForAllAsAdmin(
```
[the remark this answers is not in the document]
amiecorso: renamed

contracts/Removal.sol
```diff
@@ -88,6 +125,7 @@ contract Removal is
   ) public override {
     // todo require _to is a known market contract
     super.safeBatchTransferFrom(_from, _to, _ids, _amounts, _data);
+    // todo should this emit an event too?
```
scottynomad, line 128: I think that would be a good idea!

contracts/Removal.sol
```diff
@@ -74,6 +105,12 @@ contract Removal is
     }
     _latestTokenId = ids[ids.length - 1] + 1;
     super.mintBatch(to, ids, amounts, data);
+    emit BatchMinted(ids);
+
+    setApprovalForAll(to, _msgSender(), true); // todo look at vesting contract for potentially better approach
+    if (decodedData.list) {
+      safeBatchTransferFrom(to, decodedData.marketAddress, ids, amounts, data);
```
scottynomad, line 112: I wonder if we consider just baking the market address in during initialization?    I like the flexibility of being able to specify it but we don't have a strong usecase for that yet.  And we'd need to add the whitelisting mechanism for those market addresses.

scottynomad (approved):
Nice! This is much cleaner than the other approach.

contracts/Removal.sol
amiecorso: @jaycenhorton whatchya think?

contracts/Removal.sol
```diff
@@ -31,6 +31,21 @@ contract Removal is
...
+
+  /**
+   * @notice Emitted on successful minting of a batch of removals.
+   */
+  event BatchMinted(uint256[] tokenIds);
+
+  /**
+   * @notice Emitted on successful listing of a batch of removals.
+   */
+  event BatchListed(uint256[] tokenIds);
```
jaycenhorton, line 47: What was wrong with using the built-in events? Not opposed to using these just want to make sure added gas cost is defensible

test/Removal.test.ts
```diff
@@ -0,0 +1,162 @@
...
+    await hre.deployments.fixture(); // ensure you start from a fresh deployments
+    return getDeployments({ hre });
+  }
+);
+
+describe('Removal', () => {
+  describe('Minting removals', () => {
+    it('should mint a batch of removals without listing any', async () => {
+      const { fifoMarket, removal } = await setupTest();
+      const { supplier } = await hardhat.getNamedAccounts();
```
jaycenhorton, line 18: You can use

```ts
const supplier = hre.namedAccounts.supplier;
```

With the newly extended environment (see extendEnvironment) which awaits this promise and the signers promise once so we don't have to do this every time

test/Removal.test.ts
```diff
@@ -0,0 +1,162 @@
...
+
+      const removalBalances = ['100', '200', '300', '400'];
```
jaycenhorton, line 20: Use formatTokenAmount instead

test/Removal.test.ts
```diff
@@ -0,0 +1,162 @@
...
+        supplier,
+        removalBalances.map((balance) =>
+          hardhat.ethers.utils.parseUnits(balance)
+        ),
+        removalVintages,
+        packedData
+      );
+
+      const transactionReceipt = await transactionResponse.wait();
+      const batchMintedEvent = transactionReceipt?.events?.filter((event) => {
```
jaycenhorton, line 39: Why wouldn't these events fire? Wouldn't we want this to throw if they didn't fire since if they didn't we' could end up with a false positive in this test run?

test/Removal.test.ts
```diff
@@ -0,0 +1,162 @@
...
```
jaycenhorton, line 85: https://github.com/nori-dot-eco/contracts/pull/152/files#r809678906

test/Removal.test.ts
```diff
@@ -0,0 +1,162 @@
...
+        return event.event === 'BatchMinted';
+      })[0];
+      const tokenIds = (batchMintedEvent?.args?.tokenIds as BigNumber[]).map(
+        (tokenId) => tokenId.toNumber()
+      );
+
+      const balances = await Promise.all(
+        tokenIds.map((tokenId) => {
+          return removal.totalSupply(tokenId);
+        })
```
jaycenhorton, line 95: If the above tokenIds are an empty array this test seems like it could cause a false positive pass

test/Removal.test.ts
```diff
@@ -0,0 +1,162 @@
...
+          hardhat.ethers.utils.parseUnits(balance)
+        ),
+        removalVintages,
+        packedData
+      );
+
+      const transactionReceipt = await transactionResponse.wait();
+      const batchMintedEvent = transactionReceipt?.events?.filter((event) => {
+        return event.event === 'BatchMinted';
+      })[0];
```
jaycenhorton, line 133: https://github.com/nori-dot-eco/contracts/pull/152/files#r809678906

contracts/Removal.sol
amiecorso: nothing wrong just didn't know they were there! using them now

test/Removal.test.ts
amiecorso: nice! replaced this everywhere relevant

amiecorso:
Still think something's not right with the Coveralls settings because this is saying I increased coverage and we're above 90% but it's still blocking me from merging.

Also, in OH yesterday, I know we talked about introducing a unique parcel identifier to the process of minting removals such that we can potentially enforce a no-double-minting on the chain itself (as well as potentially other benefits).  Wondering if I can make a ticket to do that in a separate PR so that big ones aren't hanging out for so long.  I really need to move on to the vesting UI and this vintage minting stuff keeps eating a lot of branch management attention

gasReporterOutput.json
```diff
@@ -0,0 +1,57 @@
+·----------------------------------------------------------------------|---------------------------|-------------|-----------------------------·
```
jaycenhorton, line 1: Don't commit this file. Maybe add to gitignore

jaycenhorton (commented):
Remove the gas report json and this will lgtm

Any idea why gas reporter isn't tracking removal or FIFO method calls?

jaycenhorton (commented):
Maybe gas reporter code check only can report diffs since the report json file you committed definitely has data for it?

gasReporterOutput.json
amiecorso: removed

amiecorso:
gas reporter is looking good... but it's still coveralls that won't let me merge.  Is there a setting you have permission to adjust in there?

jaycenhorton (approved):

PR #162, opened 2022-02-09
Remove legacy contracts `NCCR` and `Nori_V0`

amiecorso:
NO-1527
monorepo PR: https://github.com/nori-dot-eco/nori/pull/2397

This PR removes deprecated contracts.  We leave the NCCR artifact and the addresses in `contracts.json` so that if needed we can create new test certificates manually for practicing/testing the migration of old certificate data to L2.


jaycenhorton:
variant of this now included in master

PR #165, opened 2022-02-10
Adds a pausable revokeOperator method.  NO-1096

scottynomad:


contracts/LockedNORI.sol
```diff
@@ -138,7 +138,7 @@ contract LockedNORI is
     uint256 amount,
     bytes calldata userData,
     bytes calldata operatorData
-  ) external override {
+  ) external override whenNotPaused {
```
jaycenhorton, line 141: this is redundant since `_beforeTokenTransfer` uses `whenNotPaused`

scottynomad: Hmm, yeah you're right.  I was thinking there was a path through _depositFor that would still mutate state and create a schedule but as long as it also sent tokens the `_beforeTokenTransfer` hook would be called.

contracts/LockedNORI.sol
```diff
@@ -675,6 +675,16 @@ contract LockedNORI is
     return ERC777Upgradeable.authorizeOperator(operator);
   }
 
+  /**
+   * @dev This is a wrapper, minting is only supported via `send`
+   *
+   * See {ERC20-_mint}.
+   *
+   */
+  function mint(address, uint256) public virtual override {
```
jaycenhorton, line 684: I believe with my latest this is also redundant as the public `mint` function is now longer visible by default and so this just exposes it for no reason

(mint was previously public when we were inheriting from the ERC20 preset)

contracts/LockedNORI.sol
```diff
@@ -734,4 +744,30 @@ contract LockedNORI is
...
+   *
+   * Emits {Burned} and {IERC20-Transfer} events.
+   */
+  function operatorBurn(
+    address,
+    uint256,
+    bytes memory,
+    bytes memory
+  ) public virtual override {
+    revert();
```
jaycenhorton, line 759: I think the base branch has a better variant of this override with a revert message

contracts/LockedNORI.sol
```diff
@@ -734,4 +744,30 @@ contract LockedNORI is
...
+  }
+
+  /**
+   * @dev See {IERC777-revokeOperator}.
+   */
+  function revokeOperator(address operator)
```
jaycenhorton, line 765: this seems to definitely be missing in the base though, so might be the only thing to merge from this branch

PR #169, opened 2022-02-14
add inherited erc777 preset

jaycenhorton:
de-duplicates a lot of the work required to finish permission/logic for the NORI token by moving similar LockedNORI code into an inherited preset that both NORI and LockedNORI use


scottynomad (approved):
Looks great, nice cleanup using the preset!

PR #170, opened 2022-02-16
upgrade proxies if they exist, otherwise deploy; refactor

jaycenhorton:
TODO
- [ ] need to redeploy testnet/mainnet contracts one last time (delete oz config files and contracts.json entries, then run deployment)


scottynomad (approved):
🚀

PR #171, opened 2022-02-16
Untracked docs

amiecorso:
A bunch of untracked docs pulled in from master are muddying my other PR so going to commit them here and then diff my PR from this branch. (https://github.com/nori-dot-eco/contracts/pull/152)


scottynomad:
Can we wait till the removal of the legacy contracts is fully merged (if it's not already?)?  A bunch of these docs will just need to go away again at that point.

PR #172, opened 2022-02-16
Add bridged polygon nori

jaycenhorton:


scottynomad (approved):
Makes sense to me!  Seems logical to have the root NORI contract live on mainnet.

PR #178, opened 2022-03-02
LockedNORI: Fixes a bookkeeping bug when revoking by amount.

scottynomad:
Well spotted @amiecorso !

Also added some additional guards and tests for revocation preconditions and multiple revocation calls.


test/LockedNORI.test.ts
```diff
@@ -988,12 +1010,18 @@ describe('LockedNori', () => {
...
+      ]);
+      await hardhat.network.provider.send('evm_mine');
+
       expect(await lNori.vestedBalanceOf(employee)).to.equal(newBalance);
       expect(await lNori.balanceOf(employee)).to.equal(newBalance);
 
       expect(await lNori.totalSupply()).to.eq(newBalance);
       expect(await lNori.unlockedBalanceOf(employee)).to.eq(
-        BigNumber.from('600008888888888888888')
+        BigNumber.from('600000000000000000000')
```
jaycenhorton, line 1024: for consistency, we should prefer
```suggestion
        formatTokenAmount(600)
```
though I realize `Bignumber.from` is also being used in several places so this doesn't have to happen for this PR

test/LockedNORI.test.ts
```diff
@@ -1042,8 +1070,67 @@ describe('LockedNori', () => {
...
+      const grantDetail = await lNori.getGrant(employee);
+      expect(grantDetail.grantAmount).to.equal(grantAmount.sub(quantityToRevoke));
+      expect(grantDetail.originalAmount).to.equal(grantAmount);
     });
 
+    it('Should revoke a specific amount of unvested tokens repeatedly', async () => {
+        const { lNori, grantAmount, grant } = await setupWithGrant(
+          linearParams()
+        );
+        const { admin } = await hre.getNamedAccounts();
```
jaycenhorton, line 1084: FYI you can use the following to avoid the `await`

```suggestion
        const { admin } = hre.namedAccounts;
```

Same applies to namedSigners

test/LockedNORI.test.ts
```diff
@@ -1042,8 +1070,67 @@ describe('LockedNori', () => {
...
+          .to.emit(lNori, 'UnvestedTokensRevoked')
+          .withArgs(
+            grant.startTime + DELTA,
+            grant.recipient,
+            quantityToRevoke
+          );
+  
+        expect(await lNori.balanceOf(grant.recipient)).to.equal(newBalance);
+
+        const newBalance2 = newBalance.sub(quantityToRevoke);
```
jaycenhorton, line 1107:
```suggestion
        const postRevocationBalance = newBalance.sub(quantityToRevoke);
```

test/LockedNORI.test.ts
```diff
@@ -1042,8 +1070,67 @@ describe('LockedNori', () => {
...
+            grant.startTime + DELTA,
+            grant.recipient,
+            quantityToRevoke
+          );
+
+          expect(await lNori.balanceOf(grant.recipient)).to.equal(newBalance2);
+          expect(await lNori.totalSupply()).to.eq(newBalance2);
+    
+          // Ensures grantAmount is set correctly after a revocation
+          const grantDetail2 = await lNori.getGrant(grant.recipient);
```
jaycenhorton, line 1129:
```suggestion
          const postRevocationGrantDetails = await lNori.getGrant(grant.recipient);
```

jaycenhorton (approved):

amiecorso (approved):
v nice

PR #179, opened 2022-03-02
add vesting task to hre

jaycenhorton:


contracts/ERC777PresetPausablePermissioned.sol
```diff
@@ -23,6 +23,65 @@ contract ERC777PresetPausablePermissioned is
...
+   */
+  function batchSend(
+    address[] memory recipients,
+    uint256[] memory amounts,
+    bytes[] memory userData,
+    bytes[] memory operatorData,
+    bool[] memory requireReceptionAck
+  ) public {
+    require(
+      recipients.length == amounts.length,
```
scottynomad, line 62: would be good to assert the same length on all five list args.

contracts/ERC777PresetPausablePermissioned.sol
```diff
@@ -23,6 +23,65 @@ contract ERC777PresetPausablePermissioned is
...
+   * @param requireReceptionAck list of requirements (if true, contract recipients are required to implement
+   * ERC777TokensRecipient)
+   *
+   * Emits a {SendBatch} event.
+   *
+   * ##### Requirements:
+   *
+   * - the contract must not be paused
+   */
+  function batchSend(
```
scottynomad, line 54: Is this copied over from the OZ implementation or is this net new?  Do we want to set a limit on the batch size?  I assume otherwise we're eventually going to run into a size limit here beyond which the batch fails?

jaycenhorton: No, this is custom. Size limit is going to depend on more than just number of recipients since send can call an external contract. We'd need to set a limit some other way if we want to explore that

contracts/LockedNORI.sol
```diff
@@ -303,41 +312,57 @@ contract LockedNORI is
...
-    _revokeUnvestedTokens(from, to, atTime, 0);
-  }
-
-  /**
-   * @dev revokeUnvestedTokenAmount: Truncates a vesting grant.
    *
-   * Transfers any unvested tokens in `from`'s grant to `to`
-   * and reduces the total grant size.
+   * For revocation by amount regardless of time (must by <= unvested amount):
+   *    *atTime* should be pase.  Amount must be present.
```
scottynomad, line 328: This comment doesn't make sense ...

I think it's most straightforward to say amount is zero for a revoke by amount and amount is zero for a revoke by time.

contracts/LockedNORI.sol
```diff
@@ -349,14 +374,38 @@ contract LockedNORI is
...
   /**
-   * @dev Vested balance less any claimed amount at current block timestamp.
+   * @notice Vested balance less any claimed amount at current block timestamp.
    */
   function vestedBalanceOf(address account) external view returns (uint256) {
     return _vestedBalanceOf(account, block.timestamp);
   }
 
+  // todo document expected initialzation state
+  function initialize(BridgedPolygonNORI bridgedPolygonNoriAddress)
```
scottynomad, line 384: Can we move this back down to the original place to reduce the noise in the diff?

contracts/LockedNORI.sol
```diff
@@ -397,35 +446,21 @@ contract LockedNORI is
...
-    _ERC1820_REGISTRY.setInterfaceImplementer(
-      address(this),
-      ERC777_TOKENS_RECIPIENT_HASH,
-      address(this)
-    );
-    _grantRole(TOKEN_GRANTER_ROLE, _msgSender());
+  /**
+   * @notice Unlocked balance less any claimed amount at current block timestamp.
+   */
+  function unlockedBalanceOf(address account) public view returns (uint256) {
```
scottynomad, line 457: also this one?

contracts/LockedNORI.sol
```diff
@@ -565,41 +595,50 @@ contract LockedNORI is
     address to,
     uint256 atTime,
     uint256 amount
-  ) internal {
+  ) internal onlyRole(TOKEN_GRANTER_ROLE) {
```
scottynomad, line 598: Isn't this redundant if it's enforced on the outer public function?

scottynomad (approved):
Reviewed the contract changes ... I haven't gone into the tasks yet.

contracts/LockedNORI.sol
jaycenhorton: Sure. Moved them to fix linting errors. Will open a subsequent with these changes

contracts/LockedNORI.sol
jaycenhorton: Hmm. I don't remember making this change even. Will fix

contracts/LockedNORI.sol
jaycenhorton: Yep. Must've lost this fix in merging. Will fix

contracts/LockedNORI.sol
```diff
@@ -303,41 +312,57 @@ contract LockedNORI is
...
+   * For revocation by amount regardless of time (must by <= unvested amount):
+   *    *atTime* should be pase.  Amount must be present.
    *
-   * No change is made to balances that have vested but not yet been claimed
-   * whether locked or not.
+   * ##### Requirements:
+   * - Can only be used when the caller has the `TOKEN_GRANTER_ROLE` role
+   * - fromAccounts.length == toAccounts.length == atTimes.length
+   * - The requirements of _beforeTokenTransfer apply to this function
+   * - atTimes.length == amounts.length
```
jaycenhorton, line 334: Todo: missing other requirements for other array length mismatches

tasks/index.ts
```diff
@@ -21,24 +21,36 @@ import { TASK as NORI_TASK } from './nori';
...
+const VESTING_TASK = GET_VESTING_TASK();
 interface Task {
-  run: ActionType<unknown>;
+  run: ActionType<
+    {
+      run: ActionType<{}, any>;
+    },
+    any
+  >;
 }
```
jaycenhorton, line 35: Todo rm

tasks/index.ts
```diff
@@ -21,24 +21,36 @@ import { TASK as NORI_TASK } from './nori';
...
 
+// // todo importing currently breaks the build, so we define the interface here instead https://github.com/dethcrypto/TypeChain/issues/371
```
jaycenhorton, line 36: Todo rm

contracts/LockedNORI.sol
scottynomad: Thanks! Looks like merge problems!

PR #180, opened 2022-03-03
Fixed a bug in unlockedBalanceOf when there is no vesting schedule.

scottynomad:
@amiecorso gets all the bug bounty tokens on this one too! 😛


amiecorso (approved):

contracts/LockedNORI.sol
```diff
@@ -622,11 +630,17 @@ contract LockedNORI is
   {
     TokenGrant storage grant = _grants[account];
     uint256 balance = this.balanceOf(account);
+    uint256 vestedBalance;
+    if (_hasVestingSchedule(account)) {
+        vestedBalance = grant.vestingSchedule.availableAmount(atTime);
+    } else {
+        vestedBalance = grant.grantAmount;
+    }
```
jaycenhorton, line 638: _might_ be cleaner as

```suggestion
    uint256 vestedBalance = _hasVestingSchedule(account)
      ? grant.vestingSchedule.availableAmount(atTime)
      : grant.grantAmount;
```

jaycenhorton (approved):

contracts/LockedNORI.sol
scottynomad: Good idea.  I always forget which languages have ternary operator and which don't!

PR #182, opened 2022-03-04
NO-1096 Fixes a bug in withdrawTo causing the wrong grant to be debited when withdrawing to a different address than the grant belongs to.

scottynomad:


amiecorso (approved):
wow these are some good bugs to fix

scottynomad:
> wow these are some good bugs to fix

🤦

PR #187, opened 2022-03-09
NO-1584 LNO-05M Validates sender or operator of bpnori into lockednori is has TOKEN_GRANT_CREATOR role.

scottynomad:


amiecorso (approved):

contracts/LockedNORI.sol
```diff
@@ -223,8 +226,13 @@ contract LockedNORI is
     require(
       msg.sender == address(_bridgedPolygonNori),
       "lNORI: not BridgedPolygonNORI"
-    ); // todo verify this can only be invoked by the nori contract
-    // todo restrict such that only admin can invoke this function
+    );
+    require(
+      hasRole(TOKEN_GRANTER_ROLE, from) || hasRole(TOKEN_GRANTER_ROLE, operator),
```
jaycenhorton, line 231: I'm surprised this works

jaycenhorton (approved):

contracts/LockedNORI.sol
scottynomad: why?

.openzeppelin/unknown-80001.json
```diff
@@ -386,7 +386,7 @@
             "label": "_roleMembers",
             "offset": 0,
             "slot": "151",
-            "type": "t_mapping(t_bytes32,t_struct(AddressSet)5634_storage)",
+            "type": "t_mapping(t_bytes32,t_struct(AddressSet)5546_storage)",
```
scottynomad, line 389: @jaycenhorton does this file need to be in version control?  It seems to churn a lot?

jaycenhorton: yep-- it's used by open zeppelin to track upgrades for mumbai (80001 = mumbai)

jaycenhorton: should only change if the contracts deploy a new instance though

contracts/LockedNORI.sol
jaycenhorton: Mainly surprised because I thought I had already thought of everything :joy: 

Without the `require(msg.sender == address(_bridgedPolygonNori))` this would be problematic. But since that's restricted it should be fine :thinking:

PR #190, opened 2022-03-10
NO-1585 Pin solidity version.

scottynomad:


contracts/ERC777PresetPausablePermissioned.sol
```diff
@@ -1,5 +1,5 @@
 // SPDX-License-Identifier: MIT
-pragma solidity ^0.8.0;
+pragma solidity =0.8.10;
```
jaycenhorton, line 2: can we use 0.8.12?

scottynomad: sure

jaycenhorton (approved):

PR #192, opened 2022-03-11
NO-1592 Gas optimizations and code style recommendations

scottynomad:
- Switch to using _grantRole instead of _setup_role
- LNO-02C Remove redundant interface declaration
- SUS-01C store total cliffs
- SUS-02C loop limits
- SUS-03C remove use of signed int
- SUS-04C inefficient mapping lookup
- SUS-05C redundant lookup


scottynomad:
Hmm, their gas optimization recommendations are a wash at best if not making things worse.

contracts/LockedNORI.sol
```diff
@@ -402,18 +409,15 @@ contract LockedNORI is
...
+    _bridgedPolygonNori = bridgedPolygonNoriAddress;
     _ERC1820_REGISTRY.setInterfaceImplementer(
       address(this),
       ERC777_TOKENS_RECIPIENT_HASH,
       address(this)
     );
-    _setupRole(DEFAULT_ADMIN_ROLE, _msgSender()); // todo why doesnt grantRole work
-    _setupRole(TOKEN_GRANTER_ROLE, _msgSender()); // todo why doesnt grantRole work
-    _setupRole(PAUSER_ROLE, _msgSender()); // todo why doesnt grantRole work
+    _grantRole(TOKEN_GRANTER_ROLE, _msgSender());
```
jaycenhorton, line 420: is this intentionally excluding the PAUSER_ROLE?

contracts/ScheduleUtils.sol
```diff
@@ -59,37 +60,25 @@ library ScheduleUtils {
...
+        time >= schedule.cliffs[cliffCount - 1].time,
         "Cliffs not chronological"
       );
     }
     require(time <= schedule.endTime, "Cliffs cannot end after schedule");
     require(
-      allCliffAmounts(schedule) + amount <= schedule.totalAmount,
-      "cliff amounts exceed total"
+      schedule.totalCliffAmount + amount <= schedule.totalAmount,
+      "Cliff amounts exceed total"
```
jaycenhorton, line 75:
```suggestion
      "ScheduleUtils: Cliff amounts exceed total"
```

contracts/ScheduleUtils.sol
```diff
@@ -132,11 +122,10 @@ library ScheduleUtils {
       rampStartTime = schedule.cliffs[schedule.cliffCount - 1].time;
     }
     uint256 rampTotalTime = schedule.endTime - rampStartTime;
-    int256 rampTimeElapsed = int256(atTime) - int256(rampStartTime);
-    return
-      rampTimeElapsed <= 0
-        ? 0
-        : (rampTotalAmount * uint256(rampTimeElapsed)) / rampTotalTime;
+    if (atTime < rampStartTime) {
```
jaycenhorton, line 125: was removing the ternary recommended in favor of multi-return?

contracts/LockedNORI.sol
scottynomad: I introduce a call to `__ERC777PresetPausablePermissioned_init_unchained();` just above which sets up PAUSER and ADMIN roles consistently with how we do it in the bridged nori contract.

contracts/ScheduleUtils.sol
scottynomad: not explicitly no, it was just about not using signed ints.  I'll factor tat back out.

jaycenhorton (approved):

PR #204, opened 2022-03-22
Proposal to simplify and rename `withdrawTo` to `claim`

scottynomad:
This is just a suggestion for how we can clarify the claim functionality of LockedNORI by removing the ability to claim to some other address.

@amiecorso I know this would cause a minor change to the UI.  You weren't exposing the claim to other address in the UI anyway were you?


contracts/LockedNORI.sol
```diff
@@ -244,31 +244,33 @@ contract LockedNORI is
   }
 
   /**
-   * @notice Unwrap BridgedPolygonNORI tokens and makes them available for use in the BridgedPolygonNORI contract
+   * @notice Claim unlocked tokens.
    *
-   * @dev This function burns `amount` of wrapped tokens and withdraws them to the corresponding {BridgedPolygonNORI}
-   * tokens.
+   * @dev This function burns `amount` of {LockedNORI} and transfers `amount`
```
jaycenhorton, line 249: What's the reasoning behind using the `{contract-name}` syntax? Afaik this is used in oz repos for their custom docs tempgenrator, but maybe there's another reason to us it?

contracts/LockedNORI.sol
```diff
@@ -244,31 +244,33 @@ contract LockedNORI is
...
-   * Burns unlocked tokens from sender and sends them to *recipient*
+   * Enforcement of the availability of wrapped and unlocked tokens
+   * for the `_burn` call happens in `_beforeTokenTransfer`
    *
    * ##### Requirements:
    *
    * - Can only be used when the contract is not paused.
    */
-  function withdrawTo(address recipient, uint256 amount)
+  function claim(uint256 amount)
```
jaycenhorton, line 260: The implication is that we can only withdraw to the owner and not to a different owner or contract right?

contracts/LockedNORI.sol
scottynomad: Oh, yeah I was just trying to follow their doc styles but I guess it doesn't matter since our doc generator doesn't paint these as links to the other contracts.

scottynomad:
Closing in favor of #207

PR #205, opened 2022-03-22
Meaningful token ids for removals

amiecorso:
NO-1604

This PR updates the way `Removal` tokens are minted, such that the token ids for the batch to mint are supplied to `mintBatch`, and are now expected to encode meaningful information about the removal.

The format of a 32-byte removal token id is now:

`[1byte][1byte][--2 bytes--][--2 bytes--][--2 bytes--][----------- 20 bytes------------- ][------4 bytes------]`
  tokIdV---meth&v---vintage------country--------admin1------------ supplier address ----------------subidentifier------

The first field here is the token id version, which we can use in case we need to implement new versions of this format, or keep track of removals that may have been minted without creating additional information mappings or other things that come later down the line.
Second field is methodology and methodology version.
The rest should be pretty self explanatory.

Adds a library `RemovalUtils.sol` to contain the logic for decoding a removal token id into its component fields, and also for creating one from component fields as a pure view function.  This allows us to centralize the logic for creating and decoding removal ids on chain, while our datastore can still use a call to the contract to generate removal ids for removal entities, stamp the entities with the correct id, and then mint the entities by passing this id.

Tests n' stuff.  Left some comments for more explanation and ideas.
Not totally clear on how meaningful our gas reporting numbers are right now, but notice that gas went down for the FIFOMarket, presumably because we're avoiding that vintage struct lookup, though it went up a little bit for minting, perhaps because we have an extra map that we're updating and a require check for each token id that we didn't previously have.  Seems that keeping gas fees low for certificate purchase is the priority.


contracts/Removal.sol
```diff
@@ -25,7 +25,7 @@ contract Removal is
 
   struct Vintage {
     address supplier;
-    uint16 vintage;
+    uint256 vintage;
```
amiecorso, line 28: p sure uint256 is just as cheap if not cheaper than other int sizes because of the 256bit wordsize of the evm... a fun fact I recall about solidity that also makes my life easier here

contracts/Removal.sol
```diff
@@ -23,22 +23,29 @@ contract Removal is
...
   using SafeMathUpgradeable for uint256;
 
-  struct Vintage {
-    address supplier;
-    uint16 vintage;
-    // todo: location
-    // todo: methodology
-    // todo: supplier name
-  }
-
```
amiecorso, line 33: I think this is safe to get rid of for now because all of this information (and more) is encoded in the token id.  If we end up in a place where we need more information for a removal, maybe we add back a struct or something else like a registry approach

contracts/Removal.sol
```diff
@@ -23,22 +23,29 @@ contract Removal is
...
+  uint256 constant _PARCEL_ID_FIELD_LENGTH = 5;
+  uint256 constant _PARCEL_ID_OFFSET = 7;
+  uint256 constant _VINTAGE_FIELD_LENGTH = 2;
+  uint256 constant _VINTAGE_OFFSET = 5;
+  uint256 constant _COUNTRY_CODE_FIELD_LENGTH = 2;
+  uint256 constant _COUNTRY_CODE_OFFSET = 3;
+  uint256 constant _STATE_CODE_FIELD_LENGTH = 2;
+  uint256 constant _STATE_CODE_OFFSET = 1;
+  uint256 constant _METHODOLOGY_DATA_FIELD_LENGTH = 1;
+  uint256 constant _METHODOLOGY_DATA_OFFSET = 0;
```
amiecorso, line 43: since these are all internal constants, it's chill to declare them all here without incurring additional storage gas costs (according to the internet) because their instances will be replaced with the constant value by the optimizer.  Just much better for readability like this.

contracts/Removal.sol
```diff
@@ -23,22 +23,29 @@ contract Removal is
...
+
   uint256 private _latestTokenId;
   string public name; // todo why did I add this
+  mapping(uint256 => uint256) public orderedTokenIds;
+  mapping(uint256 => bool) private _tokenIds;
```
amiecorso, line 48: these maps are for what we talked about regarding future-proofing our ability to iterate sequentially over token ids that are now of arbitrary ordering. the second map keeps track of token ids for existence so we can prevent duplications. this seemed like the better approach than trying to call `balanceOf` for the tokenId because that function requires an `address` argument, and we don't care about who may or may not have a balance of that token, we just need to know if its id exists at all.  And we don't have access to the underlying balances mapping from this contract so we keep track ourselves. maybe there's a better way

contracts/FIFOMarket.sol
```diff
@@ -125,7 +125,7 @@ contract FIFOMarket is
     address[] memory suppliers = new address[](_queueLength());
     for (uint256 i = _queueHeadIndex; i < _queueNextInsertIndex; i++) {
       uint256 removalAmount = _removal.balanceOf(address(this), _queue[i]);
-      address supplier = _removal.vintage(_queue[i]).supplier;
+      address supplier = _removal.supplierAddressFromTokenId(_queue[i]);
```
amiecorso, line 128: Since the supplier address is encoded in the removal token id itself, we don't have to perform a struct lookup to know who to pay anymore.  hopefully gas savings... wondering if there's any gas still to be saved by defining this method here in the `FIFOMarket` contract in order to avoid the external call to the removal contract... since it's just calling a view function, maybe the gas is exactly the same?  obviously better to avoid implementing the same method here if we can help it so we're not maintaining this logic in two places.

scottynomad: I think you could drop the `supplierAddressFromTokenId` and friends into a library file like we did with `ScheduleUtils.sol` then both contracts can inline it.  Seems like an optimization we can experiment with later though.

contracts/Removal.sol
scottynomad: I think this is fine.  Some take it or leave it naming ideas:

`indexToTokenId` (easier to intuit which the ids are on)
`_tokenIdExists` (more explicit about the bool value means)

scottynomad (approved):
Nice bit packing!

Do you think there's any merit to putting the `createTokenId` logic in the contract?  I'm also wondering if for readability purposes it might be useful to add another layer of abstraction in here so you could do along the lines of:

```struct UnpackedTokenId {
  uint256 address;
  uint256 parcelId;
  uint256 vintage;
  uint256 country;
  uint256 state;
}

function createTokenId(address, parcelId, ....) public pure returns (uint256 memory) {
  // 
}

function unpackTokenId(uint256 tokenId) public pure returns (UnpackedTokenId) {
 // extract all values into an UnpackedTokenId struct
}

function vintageFromTokenId(uint256 tokenId) public view returns (uint256) {
   return unpackTokenId(tokenId).vintage;
}

... other getters in terms of unpackTokenId

```

It's probably slightly less gas efficient but makes the contract a bit more self-contained to have the packing logic there.  You can also write a very trivial test for the whole thing right in solidity ie:

```
assert unpackTokenId(createTokenId(...args)) == args
```

contracts/Removal.sol
amiecorso: love it

contracts/FIFOMarket.sol
amiecorso: did it! makes a ton of sense in this case, great opportunity to learn about solidity libraries

amiecorso:
TODO:
- move meth and version to leftmost bits
- add a token id version leftmost bit
- year > country > state > address > parcel

implement the id packing logic directly in the contract as a pure view function
generate the ids with contract calls
stamp enttites
pass those ids to mint

contracts/FIFOMarket.sol
```diff
@@ -24,6 +26,8 @@ contract FIFOMarket is
   ERC1155HolderUpgradeable,
   IERC777RecipientUpgradeable
 {
+  using RemovalUtils for *;
```
scottynomad, line 29: What's the goal of the using for '*' here?  My understanding from the docs:

https://docs.soliditylang.org/en/v0.8.13/contracts.html#using-for

Is that this would attach the methods of `RemovalUtils` to all types in scope.  I think given the way you're calling those functions in a qualified way (i.e. `RemovalUtils.supplierAddressFromRemovalId(_queue[i])`) that this line may not be needed.

contracts/Removal.sol
```diff
@@ -65,40 +53,117 @@ contract Removal is
     _setApprovalForAll(owner, operator, approved);
   }
 
+  function createRemovalId(bytes calldata removalData)
```
scottynomad, line 56: This block of functions feels very redundant but I agree with the end result that they become part of `Removal`'s public API.  The rest of them after this one might also feel tidied up if they used the `removalId.version()` style naming in their implementations.

contracts/RemovalUtils.sol
```diff
@@ -0,0 +1,214 @@
...
+    returns (UnpackedRemovalIdV0 memory)
+  {
+    uint256 version = versionFromRemovalId(removalId);
+    uint256 methodology = methodologyFromRemovalId(removalId);
+    uint256 methodologyVersion = methodologyVersionFromRemovalId(removalId);
+    uint256 vintage = vintageFromRemovalId(removalId);
+    string memory country = countryCodeFromRemovalId(removalId);
+    string memory admin1 = admin1CodeFromRemovalId(removalId);
+    address supplierAddress = supplierAddressFromRemovalId(removalId);
+    uint256 subIdentifier = subIdentifierFromRemovalId(removalId);
```
scottynomad, line 97: :+1: I think that's a more future proof name.  I would splatter lots of comments around that link it back to profile.

contracts/FIFOMarket.sol
scottynomad: I think there is a path to using using :) that might look something like:

```suggestion
  using RemovalUtils for uint256;

  function test() {
    uint256 myId = 12345;
    console.log(myId.supplierAddress());
  }
```

Where the method still takes the id as the first argument but the binding (this is very similar to `.bind()` in javascript to set `this`).

```
  function supplierAddress(uint256 removalId)
    internal
    pure
    returns (address)
  {
    return
      address(
        uint160(
          _extractValue(removalId, _ADDRESS_FIELD_LENGTH, _ADDRESS_OFFSET)
        )
      );
  }
```

scottynomad (approved):
Nice!  The library feels like it was a useful refactor.  Some fairly minor comments inline but I feel like this is basically ready to go.  🚀

contracts/Removal.sol
amiecorso: definitely feels a bit cleaned up this way.... I also removed "FromRemovalId" from the name of all these functions, which is what I think your comment above was suggesting, so that we're just writing `removalId.fieldWeWant()` and it's less redundant

contracts/FIFOMarket.sol
```diff
@@ -10,6 +10,8 @@ import "@openzeppelin/contracts-upgradeable/utils/introspection/IERC1820Registry
 import "./Removal.sol";
 import "./Certificate.sol";
 import "./BridgedPolygonNORI.sol";
+import {RemovalUtils} from "./RemovalUtils.sol";
```
jaycenhorton, line 13: is your prettier/linter working?

test/helpers/index.ts
```diff
@@ -71,3 +74,37 @@ export const setupTestEnvironment = createFixture(
     };
   }
 );
+
+export const createRemovalTokenId = async (
+  removalInstance: Removal,
+  options?: Partial<UnpackedRemovalIdV0Struct>
+): Promise<BigNumber> => {
+  const defaultRemovalData: UnpackedRemovalIdV0Struct = {
```
jaycenhorton, line 82: not sure a default value is really that useful here

test/RemovalUtils.test.ts
```diff
@@ -0,0 +1,81 @@
...
+    );
+    const removalTestHarness =
+      (await RemovalTestHarness.deploy()) as RemovalTestHarness;
+    return {
+      removalTestHarness,
+    };
+  }
+);
+
+const formatRemovalIdData = (
```
jaycenhorton, line 27: is this different than the test util you wrote?

test/Removal.test.ts
```diff
@@ -1,6 +1,6 @@
 import type { ContractInstances } from '@/test/helpers';
-import { deploy } from '@/deploy/0_deploy_contracts';
 import { expect, createFixture } from '@/test/helpers';
+import { deploy } from '@/deploy/0_deploy_contracts';
```
jaycenhorton, line 3: is this used?

test/FIFOMarket.test.ts
```diff
@@ -212,13 +229,17 @@ describe('FIFOMarket', () => {
...
 
       const removalBalances = [];
-      const vintages = [];
-      const tokenIds = [];
+      let tokenIds = [];
       for (let i = 0; i <= 20; i++) {
         removalBalances.push(hre.ethers.utils.parseUnits('50'));
-        vintages.push(2018);
-        tokenIds.push(i);
+        tokenIds.push(
```
jaycenhorton, line 235: is it worth turning this into a util? A previous test in this file has similar setup logic but doesn't set up using the same pattern (`should purchase removals and mint a certificate for a small purchase spanning several removal`)

contracts/RemovalUtilsTestHarness.sol
```diff
@@ -0,0 +1,62 @@
...
+/**
+ * @dev Testbed contract for testing RemovalUtils library.
+ *
+ * Not intended for deployment in productionish environments.
+ */
+
+contract RemovalTestHarness {
+  using RemovalUtils for uint256;
+
+  function createRemovalId(bytes calldata removalData)
```
jaycenhorton, line 15: can you add natspec comments for each fn so we don't have to worry about it right before the next audit?

contracts/RemovalUtils.sol
```diff
@@ -0,0 +1,191 @@
...
+    bytes32 extractedCode = bytes32(
+      _extractValue(removalId, _COUNTRY_CODE_FIELD_LENGTH, _COUNTRY_CODE_OFFSET)
+    );
+    bytes memory bytesArray = new bytes(2);
+    bytesArray[0] = extractedCode[30];
+    bytesArray[1] = extractedCode[31];
+    return string(bytesArray);
+  }
+
+  function admin1Code(uint256 removalId) internal pure returns (string memory) {
```
jaycenhorton, line 158: what is `admin1`?

contracts/RemovalUtils.sol
```diff
@@ -0,0 +1,191 @@
...
+  uint256 methodology;
+  uint256 methodologyVersion;
+  uint256 vintage;
+  string country;
+  string admin1;
+  address supplierAddress;
+  uint256 subIdentifier;
+}
+
+uint256 constant _BITS_PER_BYTE = 8;
```
jaycenhorton, line 17: are we using this leading underscore, all caps constant style elsewhere?

contracts/Removal.sol
```diff
@@ -65,40 +53,85 @@ contract Removal is
...
    * amounts: [100 * (10 ** 18), 10 * (10 ** 18), 50 * (10 ** 18)] <- 100 tonnes, 10 tonnes, 50 tonnes in standard erc20 units (wei)
-   * vintages: [2018, 2019, 2020]
    * token id 0 URI points to vintage information (e.g., 2018) nori.com/api/removal/0 -> { amount: 100, supplier: 1, vintage: 2018, ... }
    * token id 1 URI points to vintage information (e.g., 2019) nori.com/api/removal/1 -> { amount: 10, supplier: 1, vintage: 2019, ... }
    * token id 2 URI points to vintage information (e.g., 2020) nori.com/api/removal/2 -> { amount: 50, supplier: 1, vintage: 2020, ... }
    * @param to The supplier address
    * @param amounts Each removal's tonnes of CO2 formatted as wei
-   * @param vintages The year for each removal
+   * @param ids The token ids to use for this batch of removals.
+   *            The id itself encodes the supplier's ethereum address, a parcel identifier, the vintage, country code, state code, methodology identifer, and methodology version
```
jaycenhorton, line 114: weird whitespacing

contracts/Removal.sol
```diff
@@ -65,40 +53,85 @@ contract Removal is
...
+    returns (UnpackedRemovalIdV0 memory)
+  {
+    return removalId.unpackRemovalId();
+  }
+
+  function version(uint256 removalId) public pure returns (uint256) {
+    return removalId.version();
+  }
+
+  function methodology(uint256 removalId) public pure returns (uint256) {
```
jaycenhorton, line 76: seems weird to repeat this stuff with this syntactic sugar vs just inlining it or renaming the functions in the library. Can you justify this?

contracts/Removal.sol
```diff
@@ -22,38 +23,25 @@ contract Removal is
...
+  uint256 private _tokenIdCounter;
   string public name; // todo why did I add this
+  mapping(uint256 => uint256) public indexToTokenId;
+  mapping(uint256 => bool) private _tokenIdExists;
 
   function initialize() public virtual initializer {
     super.initialize("https://nori.com/api/removal/{id}.json");
     __ERC1155Supply_init_unchained();
-    _latestTokenId = 0;
+    _tokenIdCounter = 0;
```
jaycenhorton, line 41: remind me, is there still value in a nested tokenid? can we also just encode this? A bit dusty on our previosus conversations.

contracts/FIFOMarket.sol
amiecorso: yeah it actually auto-corrects to this

test/RemovalUtils.test.ts
amiecorso: yeah, this utility just encodes the removal data as bytes to pass it to `createRemovalId`, whereas the other test utility is more involved and includes default convenience values so that we can easily generate a pre-made removal id to use in other places where the focus of the test isn't on the valid creation of the id.

test/Removal.test.ts
amiecorso: yeah it's used in `setupTest` throughout the file

contracts/RemovalUtils.sol
amiecorso: all caps yes... at least I see that for constants in `LockedNORI`.  leading underscore just indicating the privacy of the variable

test/helpers/index.ts
amiecorso: I think I disagree.. otherwise we'd have to pass values for all of these things everywhere we use this utility, when we don't actually care what the value of most of them is

contracts/Removal.sol
amiecorso: per our discussion, we removed everything except `createRemovalId` and `unpackRemovalId` which gets you all the data back at once as a struct

contracts/Removal.sol
amiecorso: as discussed, left a todo next to this to reconsider whether there's a more appropriate data structure to use to meet our needs on this one, but we're going to leave the sequential index ---> removal id map there for now

contracts/RemovalUtils.sol
amiecorso: confirmed that this is what OZ uses for internal constants.  going to mark all of those constants as internal, which is the default visibility level for all state variables

amiecorso: oh wait, since these are top level variables, they're not even part of a contract... so in that sense that are private in that they're not going to be visible to anything but whatever's defined in this file, into which they will be inlined as constants.  so i'm just going to leave this naming convention

contracts/RemovalUtilsTestHarness.sol
scottynomad: Not relevant for this test harness contract I don't think fwiw

test/Removal.test.ts
scottynomad: I removed this in my PR.  I can help you adjust accordingly.

scottynomad (approved):
Updates look good!  :+1:

PR #207, opened 2022-03-22
Explicitly disable send and transfer operations on LockedNORI

scottynomad:


contracts/LockedNORI.sol
```diff
@@ -734,4 +743,24 @@ contract LockedNORI is
     }
     return balance;
   }
+
+  function _beforeOperatorChange(address, uint256) internal override pure {
+      revert('lNORI: operator actions disable');
```
jaycenhorton, line 748:
```suggestion
      revert('lNORI: operator actions disabled');
```

jaycenhorton (approved):

PR #217, opened 2022-04-21
Prevent removals being burned with a value of 0

rlfarman:
Reduces the ids and amounts passed to `_removal.burnBatch(...)` to limit the amount of gas spent on `batchTransfer` events with a value of 0.


contracts/FIFOMarket.sol
```diff
@@ -147,27 +148,41 @@ contract FIFOMarket is
...
         break;
       }
     }
 
+    uint256[] memory batchedIds = new uint256[](numberOfRemovals);
+    uint256[] memory batchedAmounts = new uint256[](numberOfRemovals);
+
+    for (
+      uint256 i = _queueHeadIndex;
+      i < _queueHeadIndex + numberOfRemovals;
```
rlfarman, line 162: We actually shouldn't need to iterate over the whole list again, assuming _queueHeadIndex is correct, we can just iterate over a slice of the list indicated by `numberOfRemovals`

contracts/FIFOMarket.sol
```diff
@@ -147,27 +148,41 @@ contract FIFOMarket is
...
+      i++
+    ) {
+      batchedIds[i] = ids[i];
+      batchedAmounts[i] = amounts[i];
+    }
+
     bytes memory encodedCertificateAmount = abi.encode(certificateAmount);
-    _certificate.mintBatch(recipient, ids, amounts, encodedCertificateAmount);
-    for (uint256 i = 0; i < ids.length; i++) {
-      if (amounts[i] == 0) {
```
rlfarman, line 159: This is no longer necessary since we're chopping off all items from the array that would otherwise have a value of 0

scottynomad (approved):

jaycenhorton (approved):

contracts/Certificate.sol
```diff
@@ -73,7 +73,7 @@ contract Certificate is
     // todo require _sources[_latestTokenId][n] doesnt exist
     for (uint256 i = 0; i < removalIds.length; i++) {
       if (removalAmounts[i] == 0) {
-        break;
+        revert("Certificate: Removal amount 0");
```
jaycenhorton, line 76: we should prefer [custom errors](https://blog.soliditylang.org/2021/04/21/custom-errors/) moving forward

contracts/FIFOMarket.sol
```diff
@@ -154,27 +155,41 @@ contract FIFOMarket is
         suppliers[i] = supplier;
         remainingAmountToFill -= removalAmount;
       }
-
+      numberOfRemovals++;
```
jaycenhorton, line 158: this is not actionable (not worth sacrificing readability), but leaving this here for reference 
image

jaycenhorton (approved):

PR #218, opened 2022-04-22
Prevent large arrays from being initialized in tokensRecieved

rlfarman:
Testing a gas-saving theory by preventing larger than necessary arrays from ever being initialized. This means we spend more time looping through code (and violate our DRY principles), but might end up saving some gas as a result.
Just testing a curiosity of mine, although curious how this would compare to my other PR at optimization for scale.

### DIFF
--- a/contracts/FIFOMarket.sol
+++ b/contracts/FIFOMarket.sol
@@ -171,6 +171,9 @@ contract FIFOMarket is
     bytes memory encodedCertificateAmount = abi.encode(certificateAmount);
     _certificate.mintBatch(recipient, ids, amounts, encodedCertificateAmount);
     for (uint256 i = 0; i < ids.length; i++) {
+      if (_amounts[i] == 0) {
+        break;
+      }
       if (amounts[i] == _removal.balanceOf(address(this), _queue[i])) {
         _queueHeadIndex++;
       }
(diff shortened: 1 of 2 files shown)


contracts/FIFOMarket.sol
```diff
@@ -118,22 +118,16 @@ contract FIFOMarket is
...
+    // TODO: verify this can only be invoked by the nori contract
     require(
       msg.sender == address(_bridgedPolygonNori),
       "FIFOMarket: This contract can only receive BridgedPolygonNORI"
     );
 
-    uint256[] memory ids = new uint256[](_queueLength());
-    uint256[] memory amounts = new uint256[](_queueLength());
-    address[] memory suppliers = new address[](_queueLength());
+    uint256 numberOfRemovals = 0;
```
rlfarman, line 127: Loop through the array of removals just to find the right size for the arrays above.

contracts/FIFOMarket.sol
```diff
@@ -142,23 +136,51 @@ contract FIFOMarket is
...
         }
+        remainingAmountToFill -= removalAmount;
+      }
+      numberOfRemovals++;
+      if (remainingAmountToFill == 0) {
+        break;
+      }
+    }
+
+    uint256[] memory ids = new uint256[](numberOfRemovals);
```
rlfarman, line 147: Then, repeat a simplified version of the algorithm to assign the amounts + ids.

PR #220, opened 2022-04-22
Market contract shows next for sale

amiecorso:
NO-1661

Adds a view function to the market contract that returns the removal id that is next in the queue.  Note that none of the work for reserving removals has been done yet, so depending on the implementation of that, this implementation may need to change to filter out reserved removals as appropriate.


rlfarman (approved):

rlfarman:
Maybe worth adding a test for if the queue has been completely sold through

contracts/FIFOMarket.sol
```diff
@@ -71,6 +71,7 @@ contract FIFOMarket is
     return _queueNextInsertIndex - _queueHeadIndex;
   }
 
+  // todo do we want this to report NRTs in either or wei? because removal token balances are denominated in wei
```
jaycenhorton, line 74: I think wei as we do this for every unit-- doing otherwise might just be confusing as then two possible units are returned from different functions

test/FIFOMarket.test.ts
```diff
@@ -530,6 +526,99 @@ describe('FIFOMarket', () => {
...
+          vintage: 2019,
+        }),
+        createRemovalTokenId(removal, {
+          supplierAddress: supplier,
+          vintage: 2020,
+        }),
+      ]);
+      const removalBalance1 = '5';
+      const removalBalance2 = '5';
+      const removalBalance3 = '5';
```
jaycenhorton, line 581: we should just store these as bignumbers here and then you can `.add` below  for total price and not need to `parseUnits` below that in `mintBatch`

test/FIFOMarket.test.ts
```diff
@@ -530,6 +526,99 @@ describe('FIFOMarket', () => {
...
+      const purchaseAmount = '10'; // purchase first two removals
+      const fee = '1.5';
+      const totalPrice = (Number(purchaseAmount) + Number(fee)).toString();
+
+      const list = true;
+      const packedData = hre.ethers.utils.defaultAbiCoder.encode(
+        ['address', 'bool'],
+        [fifoMarket.address, list]
+      );
+      await Promise.all([
```
jaycenhorton, line 591: doesn't need `promise.all`

contracts/FIFOMarket.sol
amiecorso: ok cool, makes the most sense to me

contracts/FIFOMarket.sol
```diff
@@ -79,6 +80,12 @@ contract FIFOMarket is
     return nrtsInQueue;
   }
 
+  function nextRemovalForSale() public view returns (uint256) {
+    // todo if our queue ever becomes an array and not a map, we need to define behavior for empty queue
```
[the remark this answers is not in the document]
amiecorso: it'll depend on the implementation of our queue, which is highly up in the air right now as we grapple with the implementation of efficiently managing reserved removals.  However, it won't be hard either way.  Basically we'll just say that if the queue (however it's implemented) is empty, we return 0 (a token Id that no removal will ever have and the closest thing to null in solidity).  Because this is a map and the default value for a nonexistent entry is 0, we can do this in one line, but if that changes we'll just have to be careful we don't revert.

test/FIFOMarket.test.ts
```diff
@@ -530,6 +526,99 @@ describe('FIFOMarket', () => {
...
+      const list = true;
+      const packedData = hre.ethers.utils.defaultAbiCoder.encode(
+        ['address', 'bool'],
+        [fifoMarket.address, list]
+      );
+      const removalId = await createRemovalTokenId(removal, {
+        supplierAddress: supplier,
+      });
+
+      await Promise.all([
```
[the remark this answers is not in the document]
amiecorso: haha yeah this might have been from copy pasting more complex tests and paring it down without removing that

test/FIFOMarket.test.ts
amiecorso: there were literally so many unnecessary cases in here haha

test/FIFOMarket.test.ts
amiecorso: gonna leave this per our discussion because it's fiddly either way

contracts/FIFOMarket.sol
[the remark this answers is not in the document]
amiecorso: added the length check now anyway

jaycenhorton (approved):

PR #221, opened 2022-04-23
Market contract implements priority restricted supply

amiecorso:
NO-1467

Implements priority buyer restricted supply.
New variable for keeping track of the threshold below which we do not sell to normal buyers and require the buyer to be allowlisted.  Only the `DEFAULT_ADMIN_ROLE` can update this threshold, which is initialized to 0 on deployment.

New role `ALLOWLIST_ROLE` for designating allowlisted buyers.  The address that deploys the contract is given both `DEFAULT_ADMIN_ROLE` and `ALLOWLIST_ROLE` upon initialization.  The role admin for `ALLOWLIST_ROLE` is the `DEFAULT_ADMIN_ROLE`.

If supply is at or below the priority restricted threshold, the buyer must have the `ALLOWLIST_ROLE` in order to make a purchase.

`nextRemovalForSale` is parameterized to indicate whether it should include or exclude supply that is priority restricted. Behavior is all the same if inventory is above this threshold, but if it falls below this threshold, an argument of `true` will display the next removal for sale while an argument of `false` will exclude this reserved supply and return `0`, the value indicating no removal for sale next.

new public variable `totalSupply` keeps a running total of current inventory, because we cannot be recomputing inventory with every purchase now that we have to check inventory to see if we're below the threshold.  This value is updated whenever removals are listed or purchased.  Depending on how the implementation of `tokensReceived` changes in the future, there may be cleaner ways to handle this information.

new view function `totalUnrestrictedSupply` give the unrestricted supply only.  This would be used to determine whether out of stock on the normal buyer's interface.

`numberOfNrtsInQueue` now called `numberOfNrtsInQueueComputed` and actually computes the value based on the current state of the queue.  We should decide if we want to put a mechanism on the contract to reset the `totalSupply` variable with a fresh computation by calling this function... though this should never be necessary and would indicate a real problem with the logic of maintaining `totalSupply` if those two numbers ever became out of sync.

Tests and some test refactors to clean up the market contract tests as they grow in number and complexity.


test/FIFOMarket.test.ts
```diff
@@ -12,12 +12,40 @@ import {
 const setupTestLocal = async (
```
amiecorso, line 12: this function now takes an optional array of removal amounts that will be used to initialize inventory in the market contract upon setup.  there isn't net granular control over these removals such as which suppliers they belong to, what the vintages are etc. but that can be added later as we need more control over these aspects of the test cases.

test/FIFOMarket.test.ts
```diff
@@ -526,99 +779,6 @@ describe('FIFOMarket', () => {
       );
     });
   });
-
-  describe('nextRemovalForSale', () => {
```
amiecorso, line 530: moved this into another describe block

contracts/FIFOMarket.sol
```diff
@@ -65,25 +77,53 @@ contract FIFOMarket is
...
+  function numberOfNrtsInQueueComputed() public view returns (uint256) {
     uint256 nrtsInQueue = 0;
     for (uint256 i = _queueHeadIndex; i < _queueNextInsertIndex; i++) {
       nrtsInQueue += _removal.balanceOf(address(this), _queue[i]);
     }
     return nrtsInQueue;
   }
 
-  function nextRemovalForSale() public view returns (uint256) {
+  function totalUnrestrictedSupply() public view returns (uint256) {
```
rlfarman, line 106: Any reason you opted for a `totalSupply` class variable, instead of just reusing `numberOfNrtsInQueue`?
Such that `totalUnrestrictedSupply = numberOfNrtsInQueue - priorityRestrictedTreshold`

contracts/FIFOMarket.sol
```diff
@@ -65,25 +77,53 @@ contract FIFOMarket is
...
+    emit PriorityRestrictedThresholdSet(threshold);
   }
 
   function _queueLength() private view returns (uint256) {
     return _queueNextInsertIndex - _queueHeadIndex;
   }
 
-  // todo do we want this to report NRTs in either or wei? because removal token balances are denominated in wei
-  function numberOfNrtsInQueue() public view returns (uint256) {
+  function numberOfNrtsInQueueComputed() public view returns (uint256) {
```
rlfarman, line 98: I don't see a good reason to use this class method in tandem with a class variable. They both appear to be keeping track of the total number of NRTs in the queue, without regard to whether they're restricted or not.

contracts/FIFOMarket.sol
```diff
@@ -65,20 +77,55 @@ contract FIFOMarket is
     );
     _queueHeadIndex = 0;
     _queueNextInsertIndex = 0;
+    priorityRestrictedThreshold = 0;
+    totalSupply = 0;
+    _grantRole(DEFAULT_ADMIN_ROLE, _msgSender());
+    _grantRole(ALLOWLIST_ROLE, _msgSender()); // todo, do this here or have to do it explicitly after intialization?
```
jaycenhorton, line 83: AFAIK these all have to happen elsewhere unless we redeploy proxy (not opposed to re-deploying this contract)

jaycenhorton: This would likely change with the new [reinitializer](https://forum.openzeppelin.com/t/release-candidate-for-contracts-4-6-open-review-period/27075) feature OZ contracts 4.6 adds

contracts/FIFOMarket.sol
```diff
@@ -65,20 +77,55 @@ contract FIFOMarket is
...
+  function nextRemovalForSale(bool includePriorityRestrictedSupply)
+    public
+    view
+    returns (uint256)
+  {
+    // todo if our queue ever becomes an array and not a map, we need to define behavior for empty queue
+    // instead of trying to index into it (returning 0 if empty)
+    uint256 nextRemovalId = _queue[_queueHeadIndex];
+    if (!includePriorityRestrictedSupply) {
+      if (totalSupply <= priorityRestrictedThreshold) {
```
jaycenhorton, line 122:
```suggestion
    if (!includePriorityRestrictedSupply && totalSupply <= priorityRestrictedThreshold) {
```

test/FIFOMarket.test.ts
```diff
@@ -12,12 +12,42 @@ import {
 const setupTestLocal = async (
   {
     buyerInitialBPNoriBalance = formatTokenAmount(1_000_000),
-  }: { buyerInitialBPNoriBalance: BigNumberish } = {
+    nrtsToListAmounts = [],
+  }: {
+    buyerInitialBPNoriBalance?: BigNumberish;
+    nrtsToListAmounts?: number[];
```
jaycenhorton, line 18: maybe change to `numberOfNrtsToList`

test/FIFOMarket.test.ts
```diff
@@ -12,12 +12,42 @@ import {
...
+    );
+    const removalBalances = nrtsToListAmounts.map((amount) =>
+      hre.ethers.utils.parseUnits(amount.toString())
+    );
+
+    const packedData = hre.ethers.utils.defaultAbiCoder.encode(
+      ['address', 'bool'],
+      [fifoMarket.address, true]
+    );
+    await Promise.all([
```
jaycenhorton, line 47: unnecessary `promise.all`

test/FIFOMarket.test.ts
```diff
@@ -28,48 +58,346 @@ const setupTestLocal = async (
   return {
     hre,
     contracts,
+    removal,
+    fifoMarket,
     ...rest,
+    listedRemovalIds: tokenIds,
```
jaycenhorton, line 64: `...rest` should be last

test/FIFOMarket.test.ts
```diff
@@ -28,48 +58,346 @@ const setupTestLocal = async (
...
+          expect(
+            await fifoMarket.getRoleMemberCount(await fifoMarket[role]())
+          ).to.eq(1);
+        });
+      });
+    });
+  });
+  describe('role access', () => {
+    describe('roles', () => {
+      describe('DEFAULT_ADMIN_ROLE', () => {
```
jaycenhorton, line 94: since this is printed in the `foreach` below, should we remove this describe? That way tou can use multiple roles in the array below

jaycenhorton: if you don't want to `forEach` the roles, we should probably remove it from the object used below

test/FIFOMarket.test.ts
```diff
@@ -28,48 +58,346 @@ const setupTestLocal = async (
...
+              `AccessControl: account ${namedAccounts[
+                accountWithoutRole
+              ].toLowerCase()} is missing role ${roleId}`
+            );
+          });
+        });
+      });
+      describe('ALLOWLIST_ROLE', () => {
+        [
+          {
```
jaycenhorton, line 139: I would either write a second test (e.g., a not allowed test) for this array and the previous one or just remove the forEach for now.

test/FIFOMarket.test.ts
jaycenhorton: or, rather than using a `startingVintage` below, just accept an array like:
```ts
{vintage: number, amountToList: number}[]
```

contracts/FIFOMarket.sol
amiecorso: Deriving the value each time is O(n) wrt the size of the queue so it's simply not an option to do every time.  I think we need to keep the running total.  I was thinking of leaving this computed version of the method on the contract in for now to use in tests and to be able to detect any discrepancy that might arise between the truth and the running total

contracts/FIFOMarket.sol
amiecorso: that `totalSupply` class variable is necessary for other reasons and provides the same information as `numberOfNrtsInQueueComputed`.  I figure using it here makes it so that we don't have to change anything if we ever call this method internally in such a way that the efficiency matters.

test/FIFOMarket.test.ts
amiecorso: ok, I pulled out this `forEach` and the `roles` describe block to simplify this.  we can add a loop back in if we find ourselves adding additional tests for a given role

test/FIFOMarket.test.ts
amiecorso: ok cool, we're eventually going to need an abstraction around this other removal data, so even though it's really not necessary as part of this PR I'm refactoring this to take an array of objects, each of which can specify the details of a given removal that will be seeded into the marketplace.  everything but the amount field is optional, so that for tests where the supplier and vintage don't matter you don't always have to specify everything, these things are just assigned defaults in the setup function.

does allow me to further refactor one test whose setup had been left manual because it has to specify different suppliers

contracts/FIFOMarket.sol
amiecorso: Gonna summarize what I understand here from our convos and my research:
- When working with proxies, initializers only get called ONCE ever, unless you completely redeploy the proxy.  For this reason, making a change like this to the contract forces us to re-deploy the proxy in order for these roles to actually be granted - it's not enough to upgrade the implementation. 
- In this case, we're going to have to redeploy the proxy anyway because we're going to make breaking changes to the contract's storage anyway and everything is under development anyway, so we will just do this and it's fine.
- In fact, there's no other (secure) way to allow granting of these roles if we don't at LEAST establish a `DEFAULT_ADMIN_ROLE` during initialization. The incumbent initializer never granted that role to any address, and this contract also doesn't have a notion of "owner" or other privileged status that we could use to secure something like an external role-granting function.  So we HAVE to redeploy the proxy and grant at least that `DEFAULT_ADMIN_ROLE` role to start taking advantage of the access control features inherited by this contract.
- no other changes are required because we can use `grantRole`, `revokeRole` etc. which are already inherited by this contract to manage role access once the deployer at least has the `DEFAULT_ADMIN_ROLE`.  We might as well allowlist the deployer at the same time, but we can allowlist other addresses using `grantRole` as needed.

jaycenhorton: addendum to the first bullet: OZ contracts 4.6 introduces a reinitialize function that allows for re-initializing contracts. Until then, all the above holds true :100:

jaycenhorton (approved):

PR #226, opened 2022-04-28
Fix bug with indexes in market contract

amiecorso:
I noticed a bug in the way we use indexes in the market contract.  The collectors that we use as we decided what will fulfill an order should always be 0-indexed, but we were using the current queue head as the starting index to add into those collector arrays.

None of our existing tests detected this because we were always seeding tests and then buying from the very front of the queue (such that the queue head WAS 0, and thus the various arrays ended up being index-aligned inadvertently).

I added a test that makes two different purchases, to make sure things work correctly when we're buying from deeper in the queue than the very beginning.


rlfarman (approved):

test/FIFOMarket.test.ts
```diff
@@ -560,6 +560,76 @@ describe('FIFOMarket', () => {
...
+        hre.ethers.utils.parseUnits(totalPrice),
+        hre.ethers.utils.hexZeroPad(buyer, 32)
+      );
+      await bpNori
+        .connect(hre.namedSigners.buyer)
+        .send(
+          fifoMarket.address,
+          hre.ethers.utils.parseUnits(totalPrice),
+          hre.ethers.utils.hexZeroPad(buyer, 32)
+        );
```
jaycenhorton, line 597:
```suggestion
      const purchaseNrts = async  () => await bpNori
      .connect(hre.namedSigners.buyer)
      .send(
        fifoMarket.address,
        hre.ethers.utils.parseUnits(totalPrice),
        hre.ethers.utils.hexZeroPad(buyer, 32)
      );
      await purchaseNrts(); // deplete some of the stock (ids 0,1,2)
      await purchaseNrts(); // purchase more removals (ids 3,4,5-- tests non-zero-indexed purchases in the queue)
```

PR #227, opened 2022-04-28
Attempt at implementing set-based, round-robin-style market, allows efficient removal reservation

amiecorso:
NO-1721


contracts/FIFOMarket.sol
```diff
@@ -136,9 +155,14 @@ contract FIFOMarket is
...
-      _queue[_queueNextInsertIndex] = (ids[i]);
-      _queueNextInsertIndex++;
+      totalActiveSupply += removalAmount;
+      totalNumberActiveRemovals += 1;
+      address supplierAddress = ids[i].supplierAddress();
+      _activeSupply[supplierAddress].add(ids[i]);
+      if (!_activeSuppliersUnordered[supplierAddress]) {
+        _activeSuppliersUnordered[supplierAddress] = true;
+        _activeSuppliersOrdered.push(supplierAddress);
+      }
```
amiecorso, line 165: if this supplier doesn't yet or currently have any active supply, we have to add them to the collections we use to keep track of the active sellers

contracts/FIFOMarket.sol
```diff
@@ -182,29 +202,42 @@ contract FIFOMarket is
       "Market: This contract can only receive BridgedPolygonNORI"
     );
 
-    uint256[] memory ids = new uint256[](_queueLength());
-    uint256[] memory amounts = new uint256[](_queueLength());
-    address[] memory suppliers = new address[](_queueLength());
+    uint256[] memory ids = new uint256[](totalNumberActiveRemovals);
+    uint256[] memory amounts = new uint256[](totalNumberActiveRemovals);
+    address[] memory suppliers = new address[](totalNumberActiveRemovals);
```
amiecorso, line 207: the `totalNumberActiveRemovals` keeps track of the total number of removal ids that are for sale in the marketplace, and is basically the new "queue length" except for the fact that these ids are spread across multiple collections and so we can't get a simple `.length` on anything.  We still need this number in order to know the max number of iterations our main loop could take before concluding that there's not enough supply to fill the order

contracts/FIFOMarket.sol
```diff
@@ -255,4 +322,19 @@ contract FIFOMarket is
       AccessControlEnumerableUpgradeable.supportsInterface(interfaceId) ||
       ERC1155ReceiverUpgradeable.supportsInterface(interfaceId);
   }
+
+  function _incrementCurrentSupplierIndex() private {
+    _currentSupplierIndex =
+      (_currentSupplierIndex + 1) %
+      _activeSuppliersOrdered.length;
+  }
```
amiecorso, line 330: yay for solidity modulo

PR #228, opened 2022-04-28
Create codeql-analysis.yml

aqualen:
NO-1707


aqualen:
Flying blind here, but willing to git it a try.  Does CodeQL add value to Solidity repos?  Will it do anything?  TBD.

jaycenhorton:
> Flying blind here, but willing to git it a try. Does CodeQL add value to Solidity repos? Will it do anything? TBD.

No, but it does scan or js/ts code for vulnerabilities

For solidity, we would need something like [this slither action](https://github.com/marketplace/actions/slither-action)

jaycenhorton (approved):

PR #237, opened 2022-05-07
Set `holdbackPercentage` for removal tokens

amiecorso:
NO-1742

Creates a map from token id --> holdbackPercentage on a per-removal basis, which will be used to determine what percentage of the sale from a removal needs to be placed in the supplier's `LockedNORI` instance for insurance.

The holdback percentage must be set while the removal tokens are minted, but can be updated at a later point using `setHoldbackPercentageForRemoval`


jaycenhorton:
@scottynomad and I had a brief conversation about what to do when we need to bump the solidity version again. Not sure if we came to a solid conclusion. Hardhat does support multiple versions, but that doesn't fix the issue of OZ having different versions or the fact that we may not want to deploy different compiler versions of the NORI contracts and derived contracts

Let me stew on this

contracts/Removal.sol
```diff
@@ -77,6 +77,29 @@ contract Removal is
...
   }
 
+  function batchGetHoldbackPercentage(uint256[] calldata removalIds)
+    public
+    view
+    returns (uint256[] memory)
+  {
+    uint256[] memory holdbackPercentages = new uint256[](removalIds.length);
+    for (uint256 i = 0; i < removalIds.length; i++) {
+      require(_tokenIdExists[removalIds[i]], "removal id doesn't exist");
```
jaycenhorton, line 87: This should probably be a call to a non-batched variant of this fn

contracts/Removal.sol
```diff
@@ -77,6 +77,29 @@ contract Removal is
...
+    }
+    return holdbackPercentages;
+  }
+
+  function batchSetHoldbackPercentage(
+    uint256[] calldata removalIds,
+    uint256 holdbackPercentage
+  ) external {
+    for (uint256 i = 0; i < removalIds.length; i++) {
+      require(_tokenIdExists[removalIds[i]], "removal id doesn't exist");
```
jaycenhorton, line 98: https://github.com/nori-dot-eco/contracts/pull/237/files#r867383892

PR #238, opened 2022-05-07
2/5 - `SupplierVestingNORI.sol` for restricted supplier tokens and deployment stuff [DO NOT REVIEW]

amiecorso:
NO-1744

As discussed with Jaycen, we want to actually create a separate solidity file for the supplier vesting contract, which does nothing but inherit from what is now called `VestingNORIPreset`.  `LockedNORI` now also inherits from `VestingNORIPreset` (and we might considering renaming `LockedNORI` to `VestingNORI` or something, but didn't want to create too much noise in this PR.

Also hooks up deployment infrastructure and adds an instance of this vesting contract to the market contract, since it will need to know about the contract to route its restricted tokens and potentially create grants on the fly for suppliers that don't have one yet.

A bit of linting noise in here too but it looks like what we want


tasks/vesting.ts
```diff
@@ -688,8 +686,9 @@ const GET_BLOCKCHAIN_SUBTASK = {
   ): Promise<ParsedGrants> => {
     const totalSupply = await lNori.totalSupply();
     hre.log(`Total supply: ${totalSupply}`);
-    const rawBlockchainGrants: LockedNORI.TokenGrantDetailStructOutput[] =
```
amiecorso, line 691: I was running into an issue with this line after cleaning and building locally after my changes that was upset that `LockedNORI` is only a type but was being used as a namespace here... so I deleted the type for now but maybe there's a way around this that I couldn't find. (tried adjusting the import and it just autocorrects to `import type {...`

utils/contracts.ts
```diff
@@ -158,3 +182,16 @@ export const getFIFOMarket = async ({
     hre,
     signer,
   });
+
+export const getScheduleTestHarness = async ({
```
amiecorso, line 186: I also had to add this to satisfy the compiler in `getContractsFromDeployments` which was expecting a value for `ScheduleTestHarness` in addition to the other contracts

PR #239, opened 2022-05-08
3/5 - Route restricted portion of supplier payment to `SupplierVestingNORI` contract [DO NOT REVIEW]

amiecorso:
NO-1745

(may choose to include no 1746 in this same PR, which would add support for automated grant creation either while removal tokens are minted or when the supplier would first be paid...)

Looks up the right holdback percentage on a per-removal basis and uses this to route the correct amount to the restricted token contract for the supplier being paid.

little linting noise


test/LockedNORI.test.ts
```diff
@@ -448,7 +448,19 @@ describe('LockedNori', () => {
 
   describe('role access', () => {
     // it.todo('only the token granter role can deposit tokens')
+
     describe('roles', () => {
+      describe('DEFAULT_ADMIN_ROLE', () => {
```
amiecorso, line 453: the initializer of this contract grants the `TOKEN_GRANTER_ROLE` to whoever deploys the contract, but I couldn't explicitly find in the various initializer calls that the deployer gets the `DEFAULT_ADMIN_ROLE`, which they need to be able to grant the token granter role to the market contract... So just making sure that happens, I could take this out

PR #240, opened 2022-05-12
Tweak `CertificateCreated` event indexing

amiecorso:
When you index an argument to a smart contract event, the purpose is to make the event filterable in logs by that argument. When that argument is indexed and it's a dynamic type, the hash of that argument is what actually gets stored in the logs (so that you can efficiently filter on some input).  However, this means that the original value is lost/not stored in the logs.  For our purposes, we actually don't need to index the `removalIds` and `amounts` for the creation of the certificate because these are not values we need to filter an event search on.  Rather, we want to be able to get the original data for these arguments, thus we do not actually want to index them.
(`amounts` already wasn't indexed because you can only have up to 3 indexed arguments to an event... or 4 technically but the name of the event is indexed by default and uses one of those up)

bit of lint noise, and a type cleanup that's been giving me problems


scottynomad (approved):
LGTM!

PR #244, opened 2022-05-19
[NO-1721] [WIP] Implement round robin approach with a doubly-linked list

rlfarman:


test/FIFOMarket.test.ts
```diff
@@ -339,7 +339,9 @@ describe('FIFOMarket', () => {
         expect(await fifoMarket.numberOfActiveNrtsInMarketComputed()).to.equal(
           BigNumber.from(0)
         );
-        expect(await fifoMarket.totalActiveSupply()).to.equal(BigNumber.from(0));
```
rlfarman, line 342: Everything in this file (as of this commit, at least) is just formatting noise

test/Removal.test.ts
```diff
@@ -51,7 +52,16 @@ describe('Removal', () => {
         formatTokenAmount(balance)
       );
       const expectedMarketSupply = 1000;
-      const tokenIds = [10, 11, 12, 13];
+      const { supplier } = hre.namedAccounts;
```
rlfarman, line 55: This block of code changed to prevent issues with supplier addresses with value 0x0. Everything else is formatting noise

amiecorso: mmm yeah, this one's gotten me before

contracts/FIFOMarket.sol
```diff
@@ -242,6 +283,51 @@ contract FIFOMarket is
     _removal.burnBatch(address(this), batchedIds, batchedAmounts);
   }
 
+  // TODO batch version of this?
+  function reserveRemoval(uint256 removalId) external returns (bool) {
```
jaycenhorton, line 287: One thing I keep thinking about is whether or not a function like this should actually just transfer the removal back to the supplier. Have we thought through this or discussed this already?

contracts/FIFOMarket.sol
```diff
@@ -16,6 +16,11 @@ import "hardhat/console.sol"; // todo
 
 // todo emit events
 
+struct RoundRobinOrder {
```
jaycenhorton, line 19: any reason you defined this outside the scope of the contract?

contracts/FIFOMarket.sol
```diff
@@ -242,6 +283,51 @@ contract FIFOMarket is
...
+  function reserveRemoval(uint256 removalId) external returns (bool) {
+    address supplierAddress = removalId.supplierAddress();
+    EnumerableSetUpgradeable.UintSet storage supplierSet = _activeSupply[
+      supplierAddress
+    ];
+    require(
+      supplierSet.contains(removalId),
+      "Market: removal not in active supply"
+    );
+    supplierSet.remove(removalId);
```
jaycenhorton, line 296: Since [`.remove`](https://docs.openzeppelin.com/contracts/4.x/api/utils#EnumerableSet-remove-struct-EnumerableSet-UintSet-uint256-) returns true if the value was removed (and therefore present) the following should be sufficient:
```suggestion
    require(
      supplierSet.remove(removalId) == true,
      "Market: removal not in active supply"
    );
```

contracts/FIFOMarket.sol
```diff
@@ -242,6 +283,51 @@ contract FIFOMarket is
...
+  function unreserveRemoval(uint256 removalId) external returns (bool) {
+    address supplierAddress = removalId.supplierAddress();
+    EnumerableSetUpgradeable.UintSet storage supplierSet = _activeSupply[
+      supplierAddress
+    ];
+    require(
+      _reservedSupply.contains(removalId),
+      "Market: removal not in reserved supply"
+    );
+    _reservedSupply.remove(removalId);
```
jaycenhorton, line 319: this might actually just return false if it doesnt exist. I'd suggest wrapping it in a `require(... == true, "error")`

jaycenhorton, line 319: Since [`.remove`](https://docs.openzeppelin.com/contracts/4.x/api/utils#EnumerableSet-remove-struct-EnumerableSet-UintSet-uint256-) returns true if the value was removed (and therefore present) the following should be sufficient:
```suggestion
    require(
      _reservedSupply.remove(removalId) == true,
      "Market: removal not in reserved supply"
    );
    ```

File: contracts/FIFOMarket.sol [hunk 7]
```diff
@@ -27,18 +32,23 @@ contract FIFOMarket is
...
   uint256 private _noriFee;
   uint256 public priorityRestrictedThreshold;
-  uint256 public totalSupply;
+  uint256 public totalNumberActiveRemovals;
+  uint256 public totalActiveSupply;
+  uint256 public totalReservedSupply;
+  uint256 public activeSupplierCount;
+  address private _currentSupplierAddress;
+  mapping(address => RoundRobinOrder) private _suppliersInRoundRobinOrder;
+  EnumerableSetUpgradeable.UintSet private _reservedSupply;
```
jaycenhorton, line 50: could a [bitmap](https://hiddentao.com/archives/2018/12/10/using-bitmaps-for-efficient-solidity-smart-contracts) be a better solution for this?

[OZ has a util for bitmaps here](https://docs.openzeppelin.com/contracts/4.x/api/utils#BitMaps) 

I would suggest this happens in a separate PR but I'd like to hear your thoughts or at least a spike ticket created

contracts/FIFOMarket.sol
```diff
@@ -91,41 +102,54 @@ contract FIFOMarket is
...
 
-  function _queueLength() private view returns (uint256) {
-    return _queueNextInsertIndex - _queueHeadIndex;
-  }
-
-  function numberOfNrtsInQueueComputed() public view returns (uint256) {
-    uint256 nrtsInQueue = 0;
-    for (uint256 i = _queueHeadIndex; i < _queueNextInsertIndex; i++) {
-      nrtsInQueue += _removal.balanceOf(address(this), _queue[i]);
+  function numberOfActiveNrtsInMarketComputed() public view returns (uint256) {
```
jaycenhorton, line 105:
```suggestion
  function numberOfActiveNrtsInMarketComputed() external view returns (uint256) {
```

contracts/FIFOMarket.sol
```diff
@@ -136,9 +160,17 @@ contract FIFOMarket is
   ) public override returns (bytes4) {
     for (uint256 i = 0; i < ids.length; i++) {
       uint256 removalAmount = _removal.balanceOf(address(this), ids[i]);
-      totalSupply += removalAmount;
-      _queue[_queueNextInsertIndex] = (ids[i]);
-      _queueNextInsertIndex++;
+      totalActiveSupply += removalAmount;
+      totalNumberActiveRemovals += 1;
+      address supplierAddress = ids[i].supplierAddress();
+      _activeSupply[supplierAddress].add(ids[i]);
```
jaycenhorton, line 166:
```suggestion
      require(_activeSupply[supplierAddress].add(ids[i]) == true, "...");
```
https://docs.openzeppelin.com/contracts/4.x/api/utils#EnumerableSet-add-struct-EnumerableSet-UintSet-uint256-

contracts/FIFOMarket.sol
```diff
@@ -253,4 +339,72 @@ contract FIFOMarket is
...
+        nextSupplierAddress: _currentSupplierAddress
+      });
+      // Update the previous supplier to point to the new supplier as next
+      _suppliersInRoundRobinOrder[
+        _suppliersInRoundRobinOrder[_currentSupplierAddress]
+          .previousSupplierAddress
+      ].nextSupplierAddress = supplierAddress;
+      // Update the current supplier to point to the new supplier as previous
+      _suppliersInRoundRobinOrder[_currentSupplierAddress]
+        .previousSupplierAddress = supplierAddress;
```
jaycenhorton, line 373: I dont understand how this works-- is there a description somewhere on how this updates state with multiple suppliers?

contracts/FIFOMarket.sol
rlfarman: Thanks for the suggestion, this was from Amie's original PR and I was very much planning on changes in a follow-up PR. Will look into this approach.

contracts/FIFOMarket.sol
rlfarman: Nope! Fixed

contracts/FIFOMarket.sol
rlfarman: No, I'll go ahead and add some documentation to `FIFOMarket.md`

contracts/FIFOMarket.sol
```diff
@@ -136,9 +160,17 @@ contract FIFOMarket is
   ) public override returns (bytes4) {
     for (uint256 i = 0; i < ids.length; i++) {
       uint256 removalAmount = _removal.balanceOf(address(this), ids[i]);
-      totalSupply += removalAmount;
-      _queue[_queueNextInsertIndex] = (ids[i]);
-      _queueNextInsertIndex++;
+      totalActiveSupply += removalAmount;
+      totalNumberActiveRemovals += 1;
+      address supplierAddress = ids[i].supplierAddress();
```
jaycenhorton, line 165: since `_addActiveSupplier` mutates similar state but is not called if they are already an active supplier, I'm thinking this should probably be extracted to something like `_updateActiveSupply`

contracts/FIFOMarket.sol
```diff
@@ -182,29 +210,43 @@ contract FIFOMarket is
...
-    address[] memory suppliers = new address[](_queueLength());
+    uint256[] memory ids = new uint256[](totalNumberActiveRemovals);
+    uint256[] memory amounts = new uint256[](totalNumberActiveRemovals);
+    address[] memory suppliers = new address[](totalNumberActiveRemovals);
     uint256 numberOfRemovals = 0;
-    for (uint256 i = _queueHeadIndex; i < _queueNextInsertIndex; i++) {
-      uint256 removalAmount = _removal.balanceOf(address(this), _queue[i]);
-      address supplier = _queue[i].supplierAddress();
+    for (uint256 i = 0; i < totalNumberActiveRemovals; i++) {
+      uint256 removalId = _activeSupply[_currentSupplierAddress].at(0); // grab head of this supplier's queue
```
jaycenhorton, line 218: Should this be `.at(i)`?

contracts/FIFOMarket.sol
```diff
@@ -242,6 +283,49 @@ contract FIFOMarket is
...
+    totalNumberActiveRemovals -= 1;
+    uint256 removalBalance = _removal.balanceOf(address(this), removalId);
+    totalActiveSupply -= removalBalance;
+    totalReservedSupply += removalBalance;
+    // If this is the last removal for the supplier, remove them from active suppliers
+    if (supplierSet.length() == 0) {
+      _removeActiveSupplier(supplierAddress);
+    }
+    // todo any checks on whether this id was already in there?
+    return _reservedSupply.add(removalId); // returns true if the value was added to the set, that is, if it was not already present
```
jaycenhorton, line 305:
```suggestion
    require(_reservedSupply.add(removalId), "...");
    return true;
```

contracts/FIFOMarket.sol
jaycenhorton: nope!

contracts/FIFOMarket.sol
```diff
@@ -242,6 +283,49 @@ contract FIFOMarket is
...
+    );
+    totalNumberActiveRemovals += 1;
+    uint256 removalBalance = _removal.balanceOf(address(this), removalId);
+    totalActiveSupply += removalBalance;
+    totalReservedSupply -= removalBalance;
+    // If the supplier has previously been removed from the active suppliers, add them back
+    if (supplierSet.length() == 0) {
+      _addActiveSupplier(supplierAddress);
+    }
+    return supplierSet.add(removalId); // returns true if the value was added to the set, that is, if it was not already present
```
jaycenhorton, line 326: needs to be required as well

contracts/FIFOMarket.sol
jaycenhorton, line 305: also needs a require

contracts/FIFOMarket.sol
jaycenhorton, line 326: also needs a require

test/FIFOMarket.test.ts
```diff
@@ -8,47 +8,101 @@ import {
...
+  totalAmountOfSupply: number;
+  totalAmountOfSuppliers: number;
+  totalAmountOfRemovals: number;
+}
+
+const parseNumberToBigNumber = (numberToParse: number) =>
+  hre.ethers.utils.parseUnits(numberToParse.toString()).toString();
+
+const getTotalAmountOfSupply = (removals: RemovalDataForListing[]) =>
+  removals.reduce((sum, removal) => (sum += removal.amount), 0);
```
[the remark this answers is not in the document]
jaycenhorton: (and nori-dot-com)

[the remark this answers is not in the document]
rlfarman: Was developing in the contracts repo in isolation so that probably explains it

contracts/FIFOMarket.sol
```diff
@@ -27,18 +27,32 @@ contract FIFOMarket is
   IERC777RecipientUpgradeable
 {
   using RemovalUtils for uint256;
+  using EnumerableSetUpgradeable for EnumerableSetUpgradeable.UintSet;
+
+  /**
+   * @notice Keeps track of order of suppliers by address using a psuedo doubly-linked list.
+   */
+
```
jaycenhorton, line 35:
```suggestion
```

docs/FIFOMarket.md
```diff
@@ -1,178 +1,111 @@
...
-
-### supportsInterface
+### \_addActiveSupplier
 
 ```solidity
-function supportsInterface(bytes4 interfaceId) public view virtual returns (bool)
+function addActiveSupplier(address, supplierAddress) external
 ```
 
+_Called when a new supplier is added to the marketplace. If the first supplier, initializes a cicularly doubly-linked list, where initially the first supplier points to itself as next and previous. When a new supplier is added, at the position of the current supplier, update the previous pointer of the current supplier to point to the new supplier, and update the next pointer of the previous supplier to the new supplier._
```
jaycenhorton, line 97: These are autogenerated by `hardhat docgen`, we shouldn't edit these files manually as they'll be overwritten

test/FIFOMarket.test.ts
```diff
@@ -1,6 +1,8 @@
 import type { BigNumberish, ContractReceipt } from 'ethers';
 import { BigNumber } from 'ethers';
 
+import type { FIFOMarket, Removal } from '../typechain-types';
```
jaycenhorton, line 4:
```suggestion
import type { FIFOMarket, Removal } from '@/typechain-types';
```

test/FIFOMarket.test.ts
```diff
@@ -9,46 +11,99 @@ import {
...
+}
+
+interface RemovalDataFromListing {
+  listedRemovalIds: BigNumber[];
+  totalAmountOfSupply: number;
+  totalAmountOfSuppliers: number;
+  totalAmountOfRemovals: number;
+}
+
+const parseNumberToBigNumber = (numberToParse: number): string =>
```
jaycenhorton, line 27: does the samething as `formatTokenAmount`

test/FIFOMarket.test.ts
```diff
@@ -9,46 +11,99 @@ import {
...
+  hre.ethers.utils.parseUnits(numberToParse.toString()).toString();
+
+const getTotalAmountOfSupply = (removals: RemovalDataForListing[]): number =>
+  removals.reduce((sum, removal) => sum + removal.amount, 0);
+
+const getTotalAmountOfSuppliers = (removals: RemovalDataForListing[]): number =>
+  removals.reduce(
+    (supplierSet, removal) => supplierSet.add(removal.supplier),
+    new Set()
+  ).size ?? 1;
```
jaycenhorton, line 37: why is the default 1?

test/FIFOMarket.test.ts
```diff
@@ -9,46 +11,99 @@ import {
...
+
+const getTotalAmountOfRemovals = (removals: RemovalDataForListing[]): number =>
```
jaycenhorton, line 39: this seems oddly gratuitous :joy:

test/FIFOMarket.test.ts
```diff
@@ -9,46 +11,99 @@ import {
...
+  const listedRemovalIds = await Promise.all(
+    removalDataToList.map((removalData, index) => {
+      return createRemovalTokenId(removal, {
+        supplierAddress: removalData.supplier ?? supplier,
+        vintage: removalData.vintage ?? defaultStartingVintage + index,
+      });
+    })
+  );
+  const removalBalances = removalDataToList.map((removalData) =>
+    hre.ethers.utils.parseUnits(removalData.amount.toString())
```
jaycenhorton, line 58:
```suggestion
    formatTokenString(removalData.amount.toString())
```

test/FIFOMarket.test.ts
```diff
@@ -91,7 +149,31 @@ describe('FIFOMarket', () => {
         });
       }
     });
+    it('correctly intializes state variables', async () => {
```
jaycenhorton, line 152: can you add a todo here or extend this test so it tests for all initial state

`// todo test remaining initialization values` would be sufficient

test/FIFOMarket.test.ts
```diff
@@ -376,8 +480,146 @@ describe('FIFOMarket', () => {
...
+      const [
+        totalActiveSupply,
+        totalNumberActiveRemovals,
+        activeSupplierCount,
+      ] = await Promise.all([
+        fifoMarket.totalActiveSupply(),
+        fifoMarket.totalNumberActiveRemovals(),
+        fifoMarket.activeSupplierCount(),
+      ]);
+
```
jaycenhorton, line 505: I think it would be useful to also expect that `totalAmountOfSupply`, `totalAmountOfRemovals` and `totalAmountOfSuppliers` are expected values since this could return a false positive when `totalActiveSupply` is 0 due to initial values

test/FIFOMarket.test.ts
```diff
@@ -497,7 +744,7 @@ describe('FIFOMarket', () => {
           .toString()
       );
     });
-    it('should purchase removals and mint a certificate for a large purchase spanning many removals', async () => {
+    it('mint a certificate with multiple removals per supplier in round robin order and update state variables', async () => {
```
jaycenhorton, line 747:
```suggestion
    it('should mint a certificate with multiple removals per supplier in round robin order and update state variables', async () => {
```

test/FIFOMarket.test.ts
```diff
@@ -561,7 +810,7 @@ describe('FIFOMarket', () => {
           .toString()
       );
     });
-    it('should purchase removals and mint a certificate for a small purchase spanning several removals after another purchase has already been made', async () => {
+    it('mint a certificate with multiple removals after another purchase has already been made and update state variables', async () => {
```
jaycenhorton, line 813:
```suggestion
    it('should mint a certificate with multiple removals after another purchase has already been made and update state variables', async () => {
```

test/FIFOMarket.test.ts
```diff
@@ -637,7 +888,7 @@ describe('FIFOMarket', () => {
           .toString()
       );
     });
-    it('should correctly pay suppliers when multiple different suppliers removals are used to fulfill an order', async () => {
+    it('correctly pay suppliers when multiple different suppliers removals are used to fulfill an order', async () => {
```
jaycenhorton, line 891:
```suggestion
    it('should correctly pay suppliers when multiple different suppliers removals are used to fulfill an order', async () => {
```

I think the changed string would better be used as a describe loop, but it/test declarations are better read when they indicate that something is the expected but not guaranteed behavior

test/FIFOMarket.test.ts
```diff
@@ -741,6 +993,9 @@ describe('FIFOMarket', () => {
           .toString()
       );
     });
+    it('TODO: should not use reserved supply to fulfill an order', async () => {});
+    it('TODO: should use previously reserved supply to fulfill an order after it has been unreserved', async () => {});
```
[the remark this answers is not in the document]
jaycenhorton: is `it.todo` not valid syntax?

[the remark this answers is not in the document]
rlfarman: Not when I tried :\

test/FIFOMarket.test.ts
rlfarman: Because this would return 0 if removals don't have a supplier defined (which is most of them)

test/FIFOMarket.test.ts
rlfarman: Now that's what I call functional programming!

test/FIFOMarket.test.ts
jaycenhorton: why dont they?

test/FIFOMarket.test.ts
rlfarman: Wrote this
```
      expect(totalAmountOfSupply).to.be.greaterThan(0);
      expect(totalNumberActiveRemovals).to.be.greaterThan(0);
      expect(activeSupplierCount).to.be.greaterThan(0);
```

Is that what you had in mind?

jaycenhorton: that works

test/FIFOMarket.test.ts
rlfarman: Because the function that creates the removals on chain is set up to assign supplier to a default address if none is defined, and I decided to rely on that outstanding functionality, rather than explicitly assign a supplier to each and all of these removals (which would have been the default anyway).

I think I might take another pass at some of this testing infrastructure when I start writing more tests, but didn't want to mess around too much in one PR.

jaycenhorton (approved):

PR #245, opened 2022-05-22
add first custom error and refactor things to make analyzing changes easier

jaycenhorton:
This PR does the following two things:
1. adds a custom error
2. refactors some things to make it easier to analyze changes (like the above) to solidity files using tenderly (which is admittedly still a bit of a pain).

Here is the analysis of custom errors vs revert strings using tenderly:

| Before (using a revert string) | After (using a custom error) |
| ------------- | ------------- |
| image | image  |

I wrote the following script to analyze the change using tenderly (this hardhat script is gitignored)

```ts
import hre from 'hardhat';

import { formatTokenAmount } from '../utils/units';
import * as contracts from '../contracts.json';

const go = async () => {
  const { ethers, namedAccounts } = hre;
  const removalBalances = [1, 1].map((balance) => formatTokenAmount(balance));
  const tokenIds = [1, 1];
  const listNow = false;
  const packedData = ethers.utils.defaultAbiCoder.encode(
    ['address', 'bool'],
    [contracts.localhost.FIFOMarket.proxyAddress, listNow]
  );
  const Removal = await ethers.getContractFactory(
    'Removal' as unknown as ContractNames
  );
  const removal = await Removal.attach(
    contracts.localhost.Removal.proxyAddress
  );
  try {
    const tx = await removal.mintBatch(
      namedAccounts.supplier,
      removalBalances,
      tokenIds,
      packedData,
      { gasLimit: 1_000_000 }
    );
    console.log({ tx });
  } catch (error) {
    console.log(error);
  }
};
go();

```

I then ran this script before and after the changes using

```
TRACE=true LOG_HARDHAT_NETWORK=true npx hardhat node --no-deploy

TRACE=true LOG_HARDHAT_NETWORK=true npx hardhat deploy --network localhost

TRACE=true LOG_HARDHAT_NETWORK=true npx hardhat run --network localhost scripts/tenderly.ts # runs the above script

find ./deployments -mindepth 1 ! -regex '^./deployments/localhost_9001\(/.*\)?' -delete  # annoying workaround to get tenderly to export local transactions consistently (deletes all but the localhost_9001 deployments folder)

tenderly export --debug 0x0f1f622eb4adb26e312d40c0a0a68a4935c3a57139b4adbb983e26f225d2847d
```

---

**Conclusion: we should definitely refactor all errors to use custom errors**


contracts/Removal.sol
```diff
@@ -101,8 +103,9 @@ contract Removal is
       (BatchMintRemovalsData)
     );
     for (uint256 i = 0; i < ids.length; i++) {
-      require(!_tokenIdExists[ids[i]], "Token id already exists"); // todo can the duplicate token id be reported here?
-
+      if (_tokenIdExists[ids[i]]) {
+        revert TokenIdExists(); // todo can the duplicate token id be reported here?
```
jaycenhorton, line 107: the only significant solidity change

test/Removal.test.ts
```diff
@@ -1,166 +1,214 @@
...
-      for (const [tokenId, balance] of balances.entries()) {
-        expect(balance).to.equal(removalBalances[tokenId].toString());
-      }
+        for (const [tokenId, balance] of balances.entries()) {
+          expect(balance).to.equal(removalBalances[tokenId].toString());
+        }
+      });
+    });
+    describe('error', () => {
+      describe('TokenIdExists', () => {
```
jaycenhorton, line 171: this is the only new test in this file-- the rest was just refactored

PR #247, opened 2022-05-24
use batch send technique in fifo market

jaycenhorton:
### Gas usage for purchasing removals using `bpNori.send`

| Current | 1 supplier, 1 removal, all purchased | 1 supplier, 100 removals, all purchased  | 2 suppliers, 100 removals each (200 removals total), all purchased  |
| -------------------- |:-------------:| -----:| -----:|
| Current             | [384,057](https://dashboard.tenderly.co/jaycenhorton/nori/local-transactions/16b88e80-413c-45b9-87df-5613dddbaf88/gas-usage) | [8,507,354](https://dashboard.tenderly.co/jaycenhorton/nori/local-transactions/887e07d1-a90e-408e-bf88-0fbc81de6857/gas-usage) | [16,733,699](https://dashboard.tenderly.co/jaycenhorton/nori/local-transactions/46ac8e19-455a-48c5-9f18-50e7d40804f8/gas-usage) |
| #246 (minimal) | [405,512](https://dashboard.tenderly.co/jaycenhorton/nori/local-transactions/b4ec1fc6-8618-42ec-88f7-f19b6ca06f02/gas-usage) | [7,884,831](https://dashboard.tenderly.co/jaycenhorton/nori/local-transactions/3cefe11e-56c3-4a0b-aff5-5af6724d23d9/gas-usage) | [15,495,245](https://dashboard.tenderly.co/jaycenhorton/nori/local-transactions/2b697e6c-50b0-4379-8f99-4100ea117582/gas-usage) |
| This version     | [587,985](https://dashboard.tenderly.co/jaycenhorton/nori/local-transactions/76f05d53-55a2-4de6-bcd3-65f44268de9e/gas-usage) | [6,560,848](https://dashboard.tenderly.co/jaycenhorton/nori/local-transactions/810548a0-b2d1-4807-bafd-798ad9bad7d5/gas-usage) | [12,621,635](https://dashboard.tenderly.co/jaycenhorton/nori/local-transactions/f69ab8c9-61bb-4fa9-8369-473f9b12439c/gas-usage) |

The script I used for the tenderly exporting was:
```ts
import hre from 'hardhat';
import type { BigNumber, BigNumberish } from 'ethers';

import { formatTokenAmount } from '../utils/units';
import type { setupTest } from '../test/helpers';
import { mockDepositNoriToPolygon } from '../test/helpers';
import { createRemovalTokenId } from '../utils/removal';
import type {
  BridgedPolygonNORI,
  FIFOMarket,
  NORI,
  Removal,
} from '../typechain-types';
import { getContract } from '../utils/contracts';

const setupTestLocal = async (
  {
    buyerInitialBPNoriBalance = formatTokenAmount(1_000_000),
    removalDataToList = [],
    numberOfSuppliers = 1,
  }: {
    buyerInitialBPNoriBalance?: BigNumberish;
    numberOfSuppliers?: 1 | 2;
    removalDataToList?: {
      amount: number;
      vintage?: number;
      supplier?: string;
    }[];
  } = {
    buyerInitialBPNoriBalance: formatTokenAmount(1_000_000),
    removalDataToList: [],
    numberOfSuppliers: 1,
  }
): Promise<
  Awaited<ReturnType<typeof setupTest>> & { listedRemovalIds: BigNumber[] }
> => {
  let tokenIds: BigNumber[] = [];
  let tokenIds2: BigNumber[] = [];
  const { namedAccounts, namedSigners, ethers } = hre;
  const { supplier, buyer, unassigned0 } = namedAccounts;
  const defaultStartingVintage = 2016;
  const removal = await getContract<Removal>({
    hre: hre as any,
    signer: namedSigners.admin,
    contractName: 'Removal',
  });
  const nori = await getContract<NORI>({
    hre: hre as any,
    signer: namedSigners.admin,
    contractName: 'NORI',
  });
  const bpNori = await getContract<BridgedPolygonNORI>({
    hre: hre as any,
    signer: namedSigners.admin,
    contractName: 'BridgedPolygonNORI',
  });
  const fifoMarket = await getContract<FIFOMarket>({
    hre: hre as any,
    signer: namedSigners.admin,
    contractName: 'FIFOMarket',
  });

  tokenIds = await Promise.all(
    removalDataToList.map((removalData, index) => {
      return createRemovalTokenId({
        removalInstance: removal,
        options: {
          supplierAddress: removalData.supplier ?? supplier,
          vintage: removalData.vintage ?? defaultStartingVintage + index,
        },
        hre: hre as any,
      });
    })
  );
  const removalBalances = removalDataToList.map((removalData) =>
    ethers.utils.parseUnits(removalData.amount.toString())
  );
  const packedData = ethers.utils.defaultAbiCoder.encode(
    ['address', 'bool'],
    [fifoMarket.address, true]
  );
  await removal.mintBatch(supplier, removalBalances, tokenIds, packedData);
  if (numberOfSuppliers === 2) {
    tokenIds2 = await Promise.all(
      removalDataToList.map((removalData, index) => {
        return createRemovalTokenId({
          removalInstance: removal,
          options: {
            supplierAddress: removalData.supplier ?? supplier,
            vintage:
              removalData.vintage ?? defaultStartingVintage + (index + 100),
          },
          hre: hre as any,
        });
      })
    );
    await removal.mintBatch(
      unassigned0,
      removalBalances,
      tokenIds2,
      packedData
    );
  }

  await mockDepositNoriToPolygon({
    hre: hre as any,
    contracts: {
      BridgedPolygonNORI: bpNori,
      NORI: nori,
    },
    amount: buyerInitialBPNoriBalance,
    to: buyer,
    signer: namedSigners.buyer,
  });
  return {
    listedRemovalIds: tokenIds,
    removal,
    fifoMarket,
    bpNori,
  } as any;
};

const go = async () => {
  const { ethers, namedAccounts, namedSigners } = hre;
  const buyerInitialBPNoriBalance = formatTokenAmount(1_000_000);
  const numberOfRemovalsToCreate = 100;
  const amount = 50;
  const numberOfSuppliers = 1;
  const removalDataToList = [
    ...Array.from({ length: numberOfRemovalsToCreate }).keys(),
  ].map((_) => {
    return { amount };
  });
  const { bpNori, fifoMarket } = await setupTestLocal({
    buyerInitialBPNoriBalance,
    removalDataToList,
    numberOfSuppliers,
  });
  const { buyer } = namedAccounts;
  const purchaseAmount = String(
    numberOfRemovalsToCreate * amount * numberOfSuppliers
  ); // purchase all supply
  const fee = String(Math.floor(Number(purchaseAmount) * 0.15));
  const totalPrice = (Number(purchaseAmount) + Number(fee)).toString();
  const tx = await bpNori
    .connect(namedSigners.buyer)
    .send(
      fifoMarket.address,
      ethers.utils.parseUnits(totalPrice),
      ethers.utils.hexZeroPad(buyer, 32)
    );
  console.log({ tx });
  await tx.wait();
};
go();

```


config/networks.ts
```diff
@@ -20,7 +20,7 @@ const hardhat: NetworksUserConfig['hardhat'] = {
 const localhost: NetworkUserConfig = {
   blockGasLimit: 20_000_000,
   initialBaseFeePerGas: 1,
-  gasPrice: 1,
+  gasPrice: 2,
```
amiecorso, line 23: why are there two different gas prices for hardhat vs localhost?

contracts/FIFOMarket.sol
```diff
@@ -227,19 +252,38 @@ contract FIFOMarket is
...
         _queueHeadIndex++;
       }
-      totalSupply -= batchedAmounts[i];
-      uint256 noriFee = (batchedAmounts[i] / 100) * _noriFee;
-      uint256 supplierFee = batchedAmounts[i];
-      _bridgedPolygonNori.transfer(_noriFeeWallet, noriFee);
-      _bridgedPolygonNori.transfer(suppliers[i], supplierFee);
+      totalSupply -= batchedAmounts[i]; // todo test gas optimization with caching of these vars
+    }
+    for (uint256 i = 0; i < highestFeeIndex; i++) {
```
amiecorso, line 271: this loop iterates over the number of *unique* suppliers whose removals went into fulfilling an order (because there's only one fee for any given supplier no matter how many removals they contributed).  however the `suppliers` array that you're indexing with `i` here on line 272 may contain duplicate suppliers because a supplier address is pushed onto that array for every removal that comprises the order event if multiple come from the same supplier.  I think this needs to be fixed so that when you're populating these `fees` and `feeRecipients` arrays for the batched send you're ONLY looping over the fee structs for the order (which would involve making them indexable by index somehow).  or you create an enumerable set of *unique* suppliers for the order and iterate over that, grabbing the fee struct for each address.

contracts/FIFOMarket.sol
```diff
@@ -39,6 +47,8 @@ contract FIFOMarket is
   uint256 private _noriFee;
   uint256 public priorityRestrictedThreshold;
   uint256 public totalSupply;
+  mapping(uint256 => mapping(address => Fee)) feeAddressToFeeIndex;
```
amiecorso, line 50: I guess "fee" makes sense to use here but I was confused at first trying to figure out what's going on here because it's not really a "fee" that we're giving suppliers, it's their actual payment...  I think of fee as like a percentage whereas payment would be the actual value due someone.  Not sure I have a better idea though... and "fee" is nice and short.

Can we rename this mapping `orderToAddressToFeeStruct` or something?  it's not mapping to an index but rather to an entire `Fee`

contracts/FIFOMarket.sol
```diff
@@ -39,6 +47,8 @@ contract FIFOMarket is
...
+  uint256 private _orderId;
```
amiecorso, line 51: do we need to keep track of all this fee data in persistent storage? why don't we make these data structures in-memory only? (the orderid -> address -> Fee map).  then we wouldn't even need the concept of an order id because the data structures start fresh for every order and we can just have a `address -> Fee` map.  It occurred to me that it might be nice to have data about the orders available for the future but none of this is data that we can't get from looking at the `Certificate` contract and events, which can tell us how many orders there were (number of certs) AND way more information about the order such as who bought it, which actual removals went into it and for what amounts.  Since in the future we'll have a map from removalId -> noriFee, we'll be able to use that to reconstruct what the relative fee amounts were for the removals that comprised an order.  I think we can remove this from persistent storage, which should save hella gas since storage writes are so expensive.

contracts/FIFOMarket.sol
```diff
@@ -182,10 +190,15 @@ contract FIFOMarket is
       "Market: This contract can only receive BridgedPolygonNORI"
     );
 
+    uint256 certificateAmount = (amount * 100) / (100 + _noriFee); // todo calculate noriFee in loop to allow dynamic fees per removal
+    uint256 remainingAmountToFill = certificateAmount;
     uint256[] memory ids = new uint256[](_queueLength());
     uint256[] memory amounts = new uint256[](_queueLength());
     address[] memory suppliers = new address[](_queueLength());
     uint256 numberOfRemovals = 0;
+    Fee memory noriFee = Fee(_noriFeeWallet, 0, 0, true); // todo throw if supplier address == nori fee address
```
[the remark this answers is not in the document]
jaycenhorton: I actually went down a rabbit hole on this one yesterday and tried implementing per token royalties using [EIP 2891](https://docs.openzeppelin.com/contracts/4.x/api/token/common#ERC2981). I ultimately ended up dropping that from this PR as it would have introduced additional complexity and I wasn't even sure if we actually did want to introduce per-removal fee variables. Per-removal variable fees was actually pretty hard to figure out as well since we actually need to send the an amount of tokens + the fee when calling `bpNori.send`. The only way to do that would be to pre-calculate the fee and pass it in as data but that would also ultimately be prone to errors/bugs due to the race conditions it would introduce off-chain

contracts/FIFOMarket.sol
```diff
@@ -182,10 +190,15 @@ contract FIFOMarket is
...
+    feeAddressToFeeIndex[_orderId][_noriFeeWallet] = noriFee;
+    uint256 highestFeeIndex = noriFee.index;
```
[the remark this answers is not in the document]
jaycenhorton: The reason I didn't was to make it explicit that the nori fee index was the first index. I'm indifferent to either approach

contracts/FIFOMarket.sol
```diff
@@ -206,15 +219,27 @@ contract FIFOMarket is
...
+        highestFeeIndex,
+        true
+      );
+      feeAddressToFeeIndex[_orderId][supplier] = supplierFee;
+      noriFee.fee += (amounts[i] / 100) * _noriFee;
       if (remainingAmountToFill == 0) {
         break;
       }
     }
+    feeAddressToFeeIndex[_orderId][_noriFeeWallet] = noriFee;
```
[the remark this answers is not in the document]
jaycenhorton: on line 200 `noriFee` is initialized in the base state, on line 239 it's aggregated the total fee amount from the loop above

contracts/FIFOMarket.sol
jaycenhorton: good catch-- I like the enumerable set approach as it might end up cleaning up things a little bit

contracts/FIFOMarket.sol
jaycenhorton: oops yeah forgot to rename this after playing around with different data structures. `orderToAddressToFeeStruct` makes sense to me

contracts/FIFOMarket.sol
jaycenhorton: Can't declare mappings in memory in solidity :shrug:

amiecorso: oh wow, very interesting! and a bummer.  explainer makes sense: https://ethereum.stackexchange.com/questions/25282/why-conceptually-cant-mappings-be-local-variables

contracts/FIFOMarket.sol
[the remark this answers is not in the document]
amiecorso: ah gotcha

contracts/FIFOMarket.sol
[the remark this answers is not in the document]
amiecorso: i see, yeah not a big deal either way. reading it at first, it was harder to grok that the fee indices were a fresh concept for every order than maybe it would have been if that was set to 0 but I think it's good to leave it how it is

contracts/FIFOMarket.sol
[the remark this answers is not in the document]
amiecorso: oh right, I remember thinking about this problem and ran into the same issue about having to know the total overall fee percentage for the order to be able to calculate how many NRTs are actually getting purchased v.s. what's going to fees.  Not only is it super challenging to implement from a race condition perspective (one idea I had would be that that part of the data you pass in to send are exactly how many NRTs you want to buy, and then you submit a payment that includes an upper bound on the fees, and then you get refunded whatever doesn't go to fees) - but I feel like we should never do this because it's effectively charging different buyers different prices for NRTs, which doesn't make any good business sense at all to me.

scottynomad:
Can we close this one too?

PR #252, opened 2022-06-02
make test setup functionality more reusable; disable ethernal in tests

jaycenhorton:


test/Certificate.test.ts
```diff
@@ -1,97 +1,15 @@
...
-    hre,
-    contracts,
-    listedRemovalIds: tokenIds,
-    removal,
-    fifoMarket,
-    ...rest,
-  };
-};
-
-describe('Certificate', () => {
```
amiecorso, line 69: why is this all deleted?

amiecorso (approved):

test/Certificate.test.ts
jaycenhorton: just moved the test to bpNori tests since it is a function of bpNORI.send 

Was going to rewrite certificate tests as part of the scope of the certificate untransferability pr

PR #253, opened 2022-06-02
[NO-1772] make certificate non-transferable

jaycenhorton:
Pretty sure this should handle all transfers beyond minting and so this should just need tests to get it over the finish line


contracts/ERC1155PresetPausableNonTransferrable.sol
```diff
@@ -0,0 +1,96 @@
...
+  {
+    _grantRole(CERTIFICATE_OPERATOR_ROLE, _msgSender());
+  }
+
+  // slither-disable-next-line naming-convention, solhint-disable-next-line func-name-mixedcase
+  function __ERC1155PresetPausableNonTransferrable_init(string memory uri_)
+    internal
+    onlyInitializing
+  {
+    // todo verify all inherited initializers are called
```
amiecorso, line 44: did a little reading on `init` v.s. `init_unchained` and from what I understand, if we're using the unchained initializers we just need to make sure that our contract calls the unchained initializer for each unique parent contract in the inheritance tree?

test/ERC1155PresetPausableNonTransferrable.test.ts
```diff
@@ -0,0 +1,289 @@
...
+      await expect(
+        mockERC1155PresetPausableNonTransferrable.mintBatch(
+          hre.namedAccounts.buyer,
+          [0],
+          [0],
+          createBatchMintData({ amount: 1, hre })
+        )
+      ).to.emit(
+        mockERC1155PresetPausableNonTransferrable,
+        'BeforeTokenTransfer'
```
amiecorso, line 18: i'm having trouble understanding how this check is valuable given that the event is just emitted from the mock contract directly... I guess it's just indicating that `_beforeTokenTransfer` on the mock is being called, which should only happen if the implementation of `mintBatch` in the real parent contract calls `_beforeTokenTransfer` ? so in that sense maybe this does make sense...

test/helpers/index.ts
```diff
@@ -2,8 +2,15 @@ import type { BigNumberish, BigNumber } from 'ethers';
 import type { namedAccounts } from 'hardhat';
```
amiecorso, line 2: really nice fixture scaffolding!

contracts/ERC1155PresetPausableNonTransferrable.sol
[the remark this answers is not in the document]
jaycenhorton: correct

contracts/ERC1155PresetPausableNonTransferrable.sol
```diff
@@ -0,0 +1,96 @@
...
+    virtual
+    override(ERC1155PresetMinterPauserUpgradeable, ERC1155SupplyUpgradeable)
+  {
+    bool isNotMinting = !(from == address(0));
+    bool isNotBurning = !(to == address(0));
+    bool missingOperatorRole = !hasRole(CERTIFICATE_OPERATOR_ROLE, operator);
+    if (isNotMinting && isNotBurning && missingOperatorRole) {
+      revert ForbiddenTransferAfterMinting();
+    }
+    return super._beforeTokenTransfer(operator, from, to, ids, amounts, data);
```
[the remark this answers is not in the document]
jaycenhorton: [yeah](https://docs.soliditylang.org/en/latest/contracts.html#inheritance)

contracts/Certificate.sol
```diff
@@ -7,19 +7,21 @@ import "@openzeppelin/contracts-upgradeable/token/ERC777/IERC777RecipientUpgrade
...
 import "@openzeppelin/contracts-upgradeable/token/ERC777/IERC777RecipientUpgradeable.sol";
 import "@openzeppelin/contracts-upgradeable/utils/introspection/ERC1820ImplementerUpgradeable.sol";
-
-// import "hardhat/console.sol"; // todo
+import "./ERC1155PresetPausableNonTransferrable.sol";
 
 // todo non-transferable/approveable
 // todo disable other mint functions
+// todo whenNotPasused
+// todo setApprovalForAll should only work when called on accounts with CERTIFICATE_OPERATOR_ROLE
```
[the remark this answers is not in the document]
jaycenhorton: Agreed we should discuss this

I think that either path is possible by starting with this by default and then relinquishing the roles in the future

I left this in for now as we might want to leverage this when we migrate from our legacy certificates on L1 over to polygon and then allow users to initiate a claim.

contracts/Certificate.sol
```diff
@@ -71,30 +77,43 @@ contract Certificate is
    *  - removalIds can be used to look up vintage years, e.g. 0 -> 2018
    */
   function mintBatch(
-    address to,
+    address to, // todo array?
     uint256[] memory removalIds,
     uint256[] memory removalAmounts,
-    bytes memory data
+    bytes memory data // todo array?
```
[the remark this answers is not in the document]
jaycenhorton: not sure-- just matches consistency with our batch calls in the NORI contracts

contracts/ScheduleTestHarness.sol
```diff
@@ -8,8 +8,9 @@ import {ScheduleUtils, Schedule} from "./ScheduleUtils.sol";
  *
  * Not intended for deployment in productionish environments.
  */
-
 contract ScheduleTestHarness {
+  // todo consider TestHarness vs Mock naming convention
```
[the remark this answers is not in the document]
jaycenhorton: They are sort of different though since one is require to use a library and the other is a layer of a contract.. might leave them both but move them into a mock or test parent folder

test/ERC1155PresetPausableNonTransferrable.test.ts
jaycenhorton: this is basically just to test that our implementations of `batchMint` `safeTransferFrom` etc functions implement this. It's just a layer to explictly make sure that because those functions implement this function, then those functions don't need to test the logic of this internal individually and instead we can just test the logic of the hook on its own once

contracts/Certificate.sol
```diff
@@ -123,6 +131,28 @@ contract Certificate is
...
+   * A. The contract must not be paused
+   * B. One of the following must be true:
+   *    1. the operation is minting
+   *    2. the operation is burning
+   *    3. the operation is transferring and the operator has the `CERTIFICATE_OPERATOR_ROLE` role
+   *
+   * ##### Behaviors (todo: add behaviors and requirements to all function natspec)
+   *
+   * A. Reverts with `ForbiddenTransferAfterMinting` if none of the above requirements for requirement A are met.
+   * B. Reverts with the string `"Pausable: paused"` if the contract is paused // todo consider not inheriting pausable
```
scottynomad: A/B here are inverted wrt to the A/B in requirements above.

contracts/Certificate.sol
```diff
@@ -133,8 +163,13 @@ contract Certificate is
   )
     internal
     override(ERC1155PresetMinterPauserUpgradeable, ERC1155SupplyUpgradeable)
+    whenNotPaused
   {
-    if (from != address(0)) {
+    bool isMinting = from != address(0);
```
scottynomad: These two conditions look inverted to me.  `isMinting` is true when the from address _is_ `0`.  Likewise `isBurning` is true when to address _is_ `0`.

contracts/Certificate.sol
```diff
@@ -133,8 +163,13 @@ contract Certificate is
...
+    bool isBurning = to != address(0);
+    if (
+      (isMinting || isBurning) || !hasRole(CERTIFICATE_OPERATOR_ROLE, operator)
```
scottynomad: This doesn't entirely make sense either.  Looks like a late refactor impacted this?

If the flags were implemented as named then this would be:

`!(isMinting || isBurning || hasRole(CERTIFICATE_OPERATOR_ROLE, operator))`

PR #256, opened 2022-06-03
improve batch mint verification in tests

jaycenhorton:


jaycenhorton:
The fact that this test is failing indicates a bug that allows you to mint using a supplier addresss that is not the address of the recipient of the removal

scottynomad:
Can this be closed?

PR #258, opened 2022-06-06
[NO-1803] Sell removals in order of vintage

rlfarman:
Changes how we sell removals such that they are sold in order of vintage.
This is handled through the introduction of the RemovalQueue library, which creates a mapping of years to removals. This is used by the supplier such that we can look up the earliest year for a given supplier, and pop a removal off the queue for that year when we want to sell it.


contracts/RemovalQueue.sol
```diff
@@ -0,0 +1,37 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity ^0.8.0;
```
jaycenhorton, line 3:
```suggestion
pragma solidity =0.8.13;
```

contracts/FIFOMarket.sol
```diff
@@ -172,21 +185,21 @@ contract FIFOMarket is
...
-      address supplierAddress = ids[i].supplierAddress();
-      require(
-        _activeSupply[supplierAddress].add(ids[i]),
-        "Market: Removal already in active supply"
-      );
+      uint256 removalToAdd = ids[i];
+      uint256 removalAmount = _removal.balanceOf(address(this), removalToAdd);
+      address supplierAddress = removalToAdd.supplierAddress();
+
+      _activeSupply[supplierAddress].insertByVintage(removalToAdd);
```
jaycenhorton, line 192: needs to be asserted since `insertAfter` returns false if it fails

jaycenhorton: e.g., `require(_activeSupply[supplierAddress].insertByVintage(removalToAdd), "...")`

rlfarman: Done

contracts/RemovalQueue.sol
```diff
@@ -0,0 +1,146 @@
...
+  uint16 private constant _DEFAULT_EARLIEST_YEAR = 2**16 - 1;
+  uint16 private constant _DEFAULT_LATEST_YEAR = 0;
+
+  /**
+   * @notice Inserts a new removal to the queue.
+   * @dev The removal is added by order of vintage, such that it will be inserted before the first removal of a greater
+   * vintage that is found. If no removal with a greater vintage is found, it will be pushed to the back of the queue.
+   * @param removalToInsert new removal to insert
+   * @return bool true if success, false otherwise
+   */
```
amiecorso, line 28: doc rewrite reminder

contracts/FIFOMarket.sol
```diff
@@ -253,11 +255,12 @@ contract FIFOMarket is
         remainingAmountToFill -= removalAmount;
 
         require(
-          _activeSupply[_currentSupplierAddress].remove(removalId),
-          "Market: Removal not in active supply"
-        ); // pull it out of the supplier's queue
+          _activeSupply[_currentSupplierAddress].removeRemoval(removalId) ==
```
amiecorso, line 258: no need to compare equality

test/FIFOMarket.test.ts
```diff
@@ -614,6 +614,38 @@ describe('FIFOMarket', () => {
...
+        );
+
+      const totalNumberActiveRemovals =
+        await fifoMarket.totalNumberActiveRemovals();
+
+      // Roundabout way of showing that the removal used to fill the order was the removal with amount 5, not 1.
+      expect(totalNumberActiveRemovals.toNumber()).to.equal(
+        totalAmountOfRemovals
+      );
+    });
```
amiecorso, line 648: Test case idea: mint removals that belong to a few vintages, sell through all of a vintage's removals and into the next vintage.  see that nothing crashes. maybe try to make another purchase and see that it picks up from the next expected earliest year. (as you mention, the first case is covered by our existing tests, maybe check that the second case of making another purchase AND checking that the removals used to fulfill it are as expected is being covered)

test/helpers/index.ts
```diff
@@ -205,6 +208,7 @@ export const batchMintAndListRemovalsForSale = async ({
         removalData: {
           supplierAddress: removalData.supplierAddress ?? supplier,
           vintage: removalData.vintage ?? defaultStartingVintage + index,
+          subIdentifier: generateRandomSubIdentifier(),
```
amiecorso, line 211: love love love

amiecorso (approved):
just a few comments from our coffee date

config/networks.ts
```diff
@@ -25,6 +25,7 @@ const localhost: NetworkUserConfig = {
   ...(typeof MNEMONIC === 'string' && {
     accounts: { mnemonic: MNEMONIC },
   }),
+  allowUnlimitedContractSize: true,
```
jaycenhorton, line 28: is this still needed? enabling this could prevent us from knowing whether a contract is deployable to mainnet

test/helpers/index.ts
jaycenhorton: maybe it's time to introduce [faker](https://github.com/faker-js/faker)?

test/FIFOMarket.test.ts
```diff
@@ -678,20 +710,21 @@ describe('FIFOMarket', () => {
...
       const { bpNori, certificate, fifoMarket, hre } = await setupTestLocal({
         buyerInitialBPNoriBalance,
         removalDataToList,
       });
       const { supplier, buyer, noriWallet } = hre.namedAccounts;
 
-      const purchaseAmount = '5000'; // purchase all supply
-      const fee = '750';
+      const purchaseAmount = (amountPerRemoval * amountOfRemovals).toString(); // purchase all supply
+      const fee = (Number(purchaseAmount) * 0.15).toString();
```
jaycenhorton, line 727: use math package

test/FIFOMarket.test.ts
```diff
@@ -678,20 +710,21 @@ describe('FIFOMarket', () => {
...
       });
       const { bpNori, certificate, fifoMarket, hre } = await setupTestLocal({
         buyerInitialBPNoriBalance,
         removalDataToList,
       });
       const { supplier, buyer, noriWallet } = hre.namedAccounts;
 
-      const purchaseAmount = '5000'; // purchase all supply
-      const fee = '750';
+      const purchaseAmount = (amountPerRemoval * amountOfRemovals).toString(); // purchase all supply
```
jaycenhorton, line 726: should we just make this the following?:

```suggestion
      const purchaseAmount = (amountPerRemoval * await fifoMarket.totalNumberActiveRemovals()).toString(); // purchase all supply
```

test/FIFOMarket.test.ts
```diff
@@ -614,6 +614,38 @@ describe('FIFOMarket', () => {
...
+          removalDataToList: [
+            { amount: 1, vintage: 2015 },
+            { amount: 5, vintage: 2014 },
+          ],
+        });
+      const { buyer } = hre.namedAccounts;
+
+      const purchaseAmount = '4';
+      const fee = '.15';
+      const totalPrice = (Number(purchaseAmount) + Number(fee)).toString();
```
jaycenhorton, line 631: use math package

test/FIFOMarket.test.ts
```diff
@@ -614,6 +614,38 @@ describe('FIFOMarket', () => {
...
+        await setupTestLocal({
+          buyerInitialBPNoriBalance,
+          removalDataToList: [
+            { amount: 1, vintage: 2015 },
+            { amount: 5, vintage: 2014 },
+          ],
+        });
+      const { buyer } = hre.namedAccounts;
+
+      const purchaseAmount = '4';
```
jaycenhorton, line 629: these tests have an inconsistency across them in using strings vs numbers for these values. Might be better start as bignumbers then convert them to other formats when utilized

test/FIFOMarket.test.ts
```diff
@@ -614,6 +614,38 @@ describe('FIFOMarket', () => {
...
       );
     });
+    it('should sell removals in order of vintage regardless of minting order', async () => {
+      const buyerInitialBPNoriBalance = formatTokenAmount(1_000_000);
+      const { bpNori, fifoMarket, hre, totalAmountOfRemovals } =
+        await setupTestLocal({
+          buyerInitialBPNoriBalance,
+          removalDataToList: [
+            { amount: 1, vintage: 2015 },
+            { amount: 5, vintage: 2014 },
```
jaycenhorton, line 624: because `setupTestLocal` uses `batch` for minting in a single call, it might make this test more sound to additionally include a subsequent call to the mint removals without an earlier vintage.

contracts/RemovalQueue.sol
```diff
@@ -0,0 +1,177 @@
...
+  uint16 earliestYear;
+  uint16 latestYear;
+}
+
+library RemovalQueue {
+  using RemovalUtils for uint256;
+  using EnumerableSetUpgradeable for EnumerableSetUpgradeable.UintSet;
+
+  uint16 private constant _DEFAULT_EARLIEST_YEAR = 2**16 - 1;
+  uint16 private constant _DEFAULT_LATEST_YEAR = 0;
```
jaycenhorton, line 19: 1 thoughts for caution: unless you know what you're doing here, leave uints at uint256 so that we can do a single pass later to optimize for gas around variable sizing/ordering later. IT's counterintuitive, but sometime using uint256 is cheaper

- https://dev.to/javier123454321/solidity-gas-optimizations-pt-3-packing-structs-23f4
- https://yos.io/2021/05/17/gas-efficient-solidity/

contracts/RemovalQueue.sol
```diff
@@ -0,0 +1,177 @@
...
+
+  /**
+   * @notice Inserts a new removal to the queue.
```
jaycenhorton, line 22:
```suggestion
   * @notice Inserts a new removal into the queue.
```

contracts/RemovalQueue.sol
```diff
@@ -0,0 +1,177 @@
...
+   * @dev The removal is added to the Enumberable Set that maps to the year of its vintage.
+   * @param removalToInsert a new removal to insert.
```
jaycenhorton, line 24: missing a param for `removalQueue`

contracts/RemovalQueue.sol
```diff
@@ -0,0 +1,177 @@
...
+  ) internal returns (bool) {
+    uint16 vintageOfRemoval = removalToInsert.vintage();
+    if (isRemovalQueueEmpty(removalQueue)) {
+      removalQueue.earliestYear = vintageOfRemoval;
+      removalQueue.latestYear = vintageOfRemoval;
+    } else if (vintageOfRemoval < removalQueue.earliestYear) {
+      removalQueue.earliestYear = vintageOfRemoval;
+    } else if (vintageOfRemoval > removalQueue.latestYear) {
+      removalQueue.latestYear = vintageOfRemoval;
+    }
```
jaycenhorton, line 39: would this be the same as the following?
```solidity
   removalQueue = ({
     earliestYear: isRemovalQueueEmpty(removalQueue) || vintageOfRemoval < removalQueue.earliestYear 
       ? vintageOfRemoval 
       : removalQueue.earliestYear,
     latestYear: isRemovalQueueEmpty(removalQueue) || vintageOfRemoval > removalQueue.latestYear
       ? vintageOfRemoval 
       : removalQueue.latestYear
   })
   return removalQueue.queueByVintage[vintageOfRemoval].add(removalToInsert);
```

contracts/RemovalQueue.sol
```diff
@@ -0,0 +1,177 @@
...
+    return removalQueue.queueByVintage[vintageOfRemoval].add(removalToInsert);
```
jaycenhorton, line 40: would this be the same as the following?
```solidity
   removalQueue = ({
     earliestYear: isRemovalQueueEmpty(removalQueue) || vintageOfRemoval < removalQueue.earliestYear 
       ? vintageOfRemoval 
       : removalQueue.earliestYear,
     latestYear: isRemovalQueueEmpty(removalQueue) || vintageOfRemoval > removalQueue.latestYear
       ? vintageOfRemoval 
       : removalQueue.latestYear
   })
   return removalQueue.queueByVintage[vintageOfRemoval].add(removalToInsert);
```

contracts/RemovalQueue.sol
```diff
@@ -0,0 +1,177 @@
...
+   */
+  function removeRemoval(
+    RemovalQueueByVintage storage removalQueue,
+    uint256 removalToRemove
+  ) internal returns (bool) {
+    uint16 vintageOfRemoval = removalToRemove.vintage();
+    require(
+      removalQueue.queueByVintage[vintageOfRemoval].remove(removalToRemove),
+      "Market: failed to remove correct removal"
+    );
```
jaycenhorton, line 58: prefer custom errors

contracts/RemovalQueue.sol
```diff
@@ -0,0 +1,177 @@
...
+    // If all removals were removed, check to see if there are any updates to the struct we need to make.
+    if (isRemovalQueueEmptyForVintage(removalQueue, vintageOfRemoval)) {
+      if (removalQueue.earliestYear == removalQueue.latestYear) {
+        // If there was only one year remaining, clear the values for latest and earliest years.
+        removalQueue.earliestYear = _DEFAULT_EARLIEST_YEAR;
```
jaycenhorton, line 63: what are the implications here if the default earliest year is a year in which no future supplier will ever list a removal for again? (e.g., a couple years down the road)

contracts/RemovalQueue.sol
```diff
@@ -0,0 +1,177 @@
...
+        for (
+          uint16 currentYear = removalQueue.latestYear - 1;
+          currentYear >= removalQueue.earliestYear;
+          currentYear--
+        ) {
+          if (removalQueue.queueByVintage[currentYear].length() > 0) {
+            removalQueue.latestYear = currentYear;
+            break;
+          }
+        }
```
jaycenhorton, line 88: is this a good use case for this? https://docs.openzeppelin.com/contracts/4.x/api/utils#Arrays-findUpperBound-uint256---uint256-

contracts/RemovalQueue.sol
```diff
@@ -0,0 +1,177 @@
...
+      uint16 currentYear = removalQueue.earliestYear;
+      currentYear <= removalQueue.latestYear;
+      currentYear++
+    ) {
+      size = removalQueue.queueByVintage[currentYear].length();
+      for (i = 0; i < size; i++) {
+        currentRemoval = removalQueue.queueByVintage[currentYear].at(i);
+        totalBalance += removal.balanceOf(address(this), currentRemoval); // TODO: Use batch get for balanceOf
+      }
+    }
```
jaycenhorton, line 174: is there a reason this can't be done with something like:
```
balanceOfBatch(...address(this)[], supplierSet[supplierAddress].values()) // pseudo code, but should convey the point
```

contracts/FIFOMarket.sol
```diff
@@ -342,24 +339,22 @@ contract FIFOMarket is
...
     totalReservedSupply -= removalBalance;
     // If the supplier has previously been removed from the active suppliers, add them back
-    if (supplierSet.length() == 0) {
+    if (_activeSupply[supplierAddress].isRemovalQueueEmpty()) {
       _addActiveSupplier(supplierAddress);
     }
     require(
-      supplierSet.add(removalId),
-      "Market: Removal already in active supply"
+      _activeSupply[supplierAddress].insertRemovalByVintage(removalId),
```
jaycenhorton, line 356: prefer custom errors

contracts/FIFOMarket.sol
```diff
@@ -312,19 +312,16 @@ contract FIFOMarket is
    */
   function reserveRemoval(uint256 removalId) external returns (bool) {
     address supplierAddress = removalId.supplierAddress();
-    EnumerableSetUpgradeable.UintSet storage supplierSet = _activeSupply[
-      supplierAddress
-    ];
     require(
-      supplierSet.remove(removalId) == true,
+      _activeSupply[supplierAddress].removeRemoval(removalId),
```
jaycenhorton, line 316: prefer custom errors

contracts/FIFOMarket.sol
```diff
@@ -164,6 +158,9 @@ contract FIFOMarket is
   //   return nextRemovalId;
   // }
 
+  /**
+   * @dev Will not work if the removals are for more than one supplier!
```
jaycenhorton, line 162: should this be a TODO?

contracts/FIFOMarket.sol
```diff
@@ -172,13 +169,13 @@ contract FIFOMarket is
...
-      totalNumberActiveRemovals += 1;
-      address supplierAddress = ids[i].supplierAddress();
+      uint256 removalToAdd = ids[i];
+      uint256 removalAmount = _removal.balanceOf(address(this), removalToAdd); // TODO: Batch get removal amounts
+      address supplierAddress = removalToAdd.supplierAddress();
+
       require(
-        _activeSupply[supplierAddress].add(ids[i]),
-        "Market: Removal already in active supply"
+        _activeSupply[supplierAddress].insertRemovalByVintage(removalToAdd),
```
jaycenhorton, line 177: prefer custom errors

contracts/FIFOMarket.sol
```diff
@@ -172,13 +169,13 @@ contract FIFOMarket is
     bytes memory
   ) public override returns (bytes4) {
     for (uint256 i = 0; i < ids.length; i++) {
-      uint256 removalAmount = _removal.balanceOf(address(this), ids[i]);
-      totalActiveSupply += removalAmount;
-      totalNumberActiveRemovals += 1;
-      address supplierAddress = ids[i].supplierAddress();
+      uint256 removalToAdd = ids[i];
+      uint256 removalAmount = _removal.balanceOf(address(this), removalToAdd); // TODO: Batch get removal amounts
```
jaycenhorton, line 173: use `balanceOfBatch`

contracts/FIFOMarket.sol
```diff
@@ -253,11 +253,11 @@ contract FIFOMarket is
         remainingAmountToFill -= removalAmount;
 
         require(
-          _activeSupply[_currentSupplierAddress].remove(removalId),
-          "Market: Removal not in active supply"
-        ); // pull it out of the supplier's queue
+          _activeSupply[_currentSupplierAddress].removeRemoval(removalId),
+          "Market: Failed to remove removal from supply"
```
jaycenhorton, line 257: prefer custom error

jaycenhorton (changes requested):
Nicely done!

Requested some hopefully insignificant changes below. Would love the same thing to be done for `RoundRobinOrder` and then I think the Market contract will be in good shape as far as readability goes

config/networks.ts
rlfarman: Nope, will rm

test/FIFOMarket.test.ts
rlfarman: I don't think so, since we're trying to buy the full amount of removals it makes more sense to use the input data as our counter, just in case something goes silently wrong when listing removals (and avoids an unnecessary await). But that's just personal preference.

test/FIFOMarket.test.ts
rlfarman: Makes sense, I have a feeling testing is gonna get a major lift soon enough so I'll wait for another PR to standardize these tests.

contracts/RemovalQueue.sol
rlfarman: The default earliest year is just a way of nulling out this value in such a way where the conditionals will work elsewhere.
Here's the definition:
`uint16 private constant _DEFAULT_EARLIEST_YEAR = 2**16 - 1;`
So I'm not sure I understand your question.

contracts/RemovalQueue.sol
rlfarman: I don't think so, since we're not iterating through an array, it's a mapping. Unless I'm misunderstanding something.

contracts/RemovalQueue.sol
rlfarman: You need to get the extract the values from each enumerable set within the mapping before you could do this, so it's not as simple as what you've written.

contracts/FIFOMarket.sol
rlfarman: I need to refresh my brain but I actually don't think this applies anymore.

jaycenhorton: I'd be curious why not if so

test/FIFOMarket.test.ts
jaycenhorton: makes more sense to calculate it explicitly from `removalDataToList` then

contracts/RemovalQueue.sol
jaycenhorton: Not sure how solidity deals with struct value setting (e.g., whether the overhead is related to per property set vs if any property is set)-- do you know?

contracts/RemovalQueue.sol
jaycenhorton: works for me-- perhaps make a ticket?

contracts/RemovalQueue.sol
jaycenhorton: if the earliest year is a year far in the past, does thath have any implications on the overhead of iterations?

contracts/RemovalQueue.sol
jaycenhorton: there is significant overhead for each time you invoke `.balanceOf` (vs a single balanceOfBatch) so I'd like to see this extracted to a batch version

test/FIFOMarket.test.ts
rlfarman: I kinda disagree, since `removalDataToList` is derived from the `amountOfRemovals` in the first place. I think using that initially defined variable establishes the clearest intent for this particular function of the test. Can you help me understand what issues you're seeing with this?

contracts/RemovalQueue.sol
rlfarman: Agree, one way or another I'll get this into a balanceOfBatch

contracts/RemovalQueue.sol
rlfarman: No, but it's not about updating both, it's about for an iteration where currentYear isn't below the earliestYear or above the latestYear - you'd just update the struct with the values that are already there. So that's the cost I'm worried about (since that would be the case for most removals).

contracts/RemovalQueue.sol
rlfarman: > Using smaller-size uints such as uint8 is only more efficient when you can pack variables into the same storage slot, such as in structs. In other cases and in loops, uint256 is more gas efficient than uint8.

Since this is the case here (ealiestYear and latestYear are member of the RemovalQueueByVintage struct which is kept in storage), I think my implementation is the right call.
Really cool detail though, good callout.

rlfarman: Also, I think it's funny that you misspelled `uint256` and uint16 in your comments (uin instead of uint), because I do that 100% of the time when writing these types. Something about that `t` just makes my brain skip over it.

contracts/RemovalQueue.sol
rlfarman: No, since `earliestYear` is always update from the `_DEFAULT_EARLIEST_YEAR` to be whatever was added by a new removal if the default is detected.

contracts/RemovalQueue.sol
jaycenhorton: I see- in that case perhaps a single if statement would avoid that, right?

contracts/RemovalQueue.sol
jaycenhorton: What about in the rest of the file where the same pattern is followed and we need to declare variables as non-uint256?

contracts/RemovalQueue.sol
rlfarman: Yes, although I don't see how that ends up any cleaner than the current implementation. Which also avoids the potential for additional gas costs mentioned above.
Also not sure how struct assignment works, but would the struct variable `mapping(uint16 => EnumerableSetUpgradeable.UintSet) queueByVintage;` be reset it we didn't otherwise specify it above when reassigning the struct?
Any reason you're suggesting this? Trying to avoid `else if` statements? Curious about the rational.

contracts/RemovalQueue.sol
rlfarman: Sorry, not sure I follow. Are you saying what if we need to instantiate variables for vintages outside of storage? Then I suppose we can just use uint256, the two types can be implicitly converted.

test/FIFOMarket.test.ts
rlfarman: I see, fair enough, will change

contracts/RemovalQueue.sol
jaycenhorton: This is more what I was looking for:



| test case | Using `uint16`  | Using `uint256` |
| ------------- | ------------- | ------------- |
| should mint a certificate with some of a single removal in round robin order and update state variables  | (801165 gas) |  (823110 gas) |
| should mint a certificate with one removal per supplier in round robin order and update state variables  | (1313367 gas) | (1292108 gas)  |

contracts/RemovalQueue.sol
jaycenhorton: I don't particularly have a preference beyond preferring the option that is the most gas efficient. Can you run gas tests locally and see if either variant has higher overhead? [Instructions in the readme](https://github.com/nori-dot-eco/contracts#reporting-gas-usage-from-tests)-- would like to see something similar to [this](https://github.com/nori-dot-eco/contracts/pull/258#discussion_r897096269)

rlfarman: I get a TypeError trying to use a slightly corrected version of the above. I think this is actually why I went about this the way I did earlier.
`TypeError: Types in storage containing (nested) mappings cannot be assigned to.`
[This](https://stackoverflow.com/questions/64200059/solidity-problem-creating-a-struct-containing-mappings-inside-a-mapping) is what I looked at on Stack Overflow, which guided my own implementation.
I rebuilt this with the closest possible equivalent:
```
    removalQueue.earliestYear = isRemovalQueueEmpty(removalQueue) || vintageOfRemoval < removalQueue.earliestYear 
       ? vintageOfRemoval 
       : removalQueue.earliestYear;
     removalQueue.latestYear = isRemovalQueueEmpty(removalQueue) || vintageOfRemoval > removalQueue.latestYear
       ? vintageOfRemoval 
       : removalQueue.latestYear;
```
Which resulted in slight gas increases.

rlfarman: Screen Shot 2022-06-14 at 2 30 06 PM
Here's a diff. Current implementation in red, implementation in comment above in red.

jaycenhorton: think there's a typo in your statement. Which implementation is which?

jaycenhorton (approved):

PR #259, opened 2022-06-07
[NO-1776] - add function to retrieve active supply

jaycenhorton:


contracts/FIFOMarket.sol
```diff
@@ -133,13 +151,113 @@ contract FIFOMarket is
...
+  function activeSupply(bool includeRestricted)
+    external
+    view
+    returns (ActiveSupply memory)
+  {
+    address supplierAddress = _currentSupplierAddress;
+    ActiveSupplier[] memory suppliers = new ActiveSupplier[](
+      activeSupplierCount
+    );
+    address[] memory fifoMarketAddressArray = new address[](
```
scottynomad, line 165: why still using fifo in terminology?

rlfarman: It's all semantics at this point, but technically suppliers are still ordered in the queue by FIFO.

PR #261, opened 2022-06-09
WIP - adds the ability to simulate transactions

jaycenhorton:
Just opening up this PR for feedback on transaction simulation approaches. This is useful for many many things, but it might also be the most dynamic approach to the active supplier endpoint. An approach like this lets you simulate a transaction by executing it in a sandbox without using gas. In this way you could use an approach like this to investigate a queue for any given order or order of size `N`

For example-- want to see an individualized or global view of the suppliers which would be included in an order of size N for wallet A? This will give you exactly that without any additional on-chain logic 

This signer is also useful for things beyond the above, such as 
- simulating large transactions before attempting them 
- debugging failed transactions
- introspection of possible future state transition of any given transaction

Longer term, I'd imagine this migrating to a plugin vs living here, but since I threw this together super fast I'm just going to let it stew here for feedback.

An example using the active supplier queue for an order of size N works something like the following:

```ts
    const simulatedTx = await (bpNori as any)
      .simulate // custom functionality to invoke tenderly simulator API call against any possible contract function (like a much beefier callStatic)
      .send(
          fifoMarket.address,
          hre.ethers.utils.parseUnits(totalPrice.toString()),
          hre.ethers.utils.hexZeroPad(hre.namedAccounts.buyer, 32)
      );
    const transactionInfo = simulatedTx.simulation.data.transaction.transaction_info;
    for (const log of transactionInfo.logs) {
      if (log.raw.address.toLowerCase() === removal.address.toLowerCase()) {
        const logs = removal.interface.parseLog(log.raw);
        console.log({ logs, removalIdsPurchasedInSimulation: logs.args.ids });
      }
    }
```

Which would output something like the following for mumbai (note that mumbai has old deployments so the output is a bit buggy but that's because the existing FIFO deployment is out of sync with master)

```ts
{
  logs: LogDescription {
    eventFragment: {
      name: 'TransferBatch',
      anonymous: false,
      inputs: [...],
      type: 'event',
      _isFragment: true,
      constructor: [Function],
      format: [Function (anonymous)]
    },
    name: 'TransferBatch',
    signature: 'TransferBatch(address,address,address,uint256[],uint256[])',
    topic: '0x4a39dc06d4c0dbc64b70af90fd698a233a518aa5d07e595d983b8c0526c8f7fb',
    args: [
      '0x61819E8238A3493547Ae1131dc657D87F62833FD',
      '0x61819E8238A3493547Ae1131dc657D87F62833FD',
      '0x0000000000000000000000000000000000000000',
      [...],
      [...],
      operator: '0x61819E8238A3493547Ae1131dc657D87F62833FD',
      from: '0x61819E8238A3493547Ae1131dc657D87F62833FD',
      to: '0x0000000000000000000000000000000000000000',
      ids: [...]
    ]
  },
  removalIdsPurchasedInSimulation: [
    BigNumber { value: "0" },
    BigNumber { value: "28323940234691845125105973673580222923891498263885679352364510073369657344" },
    BigNumber { value: "0" },
    BigNumber { value: "0" }
  ]
}

```


signers/simulator.ts
```diff
@@ -0,0 +1,72 @@
...
+
+  readonly provider: Provider;
+
+  constructor(signer: Signer) {
+    super();
+    this.signer = signer;
+    this.provider = signer.provider!;
+  }
+
+  connect(provider: Provider): TenderlySimulationSigner {
```
mrdave-dev, line 22: How is this different from `this.provider`? Should we use that here instead?

jaycenhorton:
Replaced by #380

PR #265, opened 2022-06-14
Incremental improvements to gas estimation and adds a memo to vesting transactions.

scottynomad:
Also removes all use of old style gasPrice pricing and uses eip1559 throughout.


plugins/fireblocks/fireblocks-signer.ts
```diff
@@ -136,10 +136,10 @@ export class FireblocksSigner extends Signer implements TypedDataSigner {
    * @returns
    */
   async _populateTransaction(
-    transaction: Deferrable<TransactionRequest>,
-    type = 2
+    transaction: Deferrable<TransactionRequest>
   ): Promise<UnsignedTransaction> {
-    const feeData = getGasPriceSettings(this._bridge.getChainId());
+    const feeData = await getGasFeeSettings(this._bridge.getChainId());
+    console.log(feeData);
```
jaycenhorton, line 142: can we use `hre.trace/log` here?

plugins/fireblocks/fireblocks-signer.ts
```diff
@@ -279,14 +283,18 @@ export class FireblocksSigner extends Signer implements TypedDataSigner {
...
     let txHash: string;
-    let baseTx: UnsignedTransaction;
+
+    let baseTx: UnsignedTransaction = await this._populateTransaction(
+      transaction
+    );
     if (transaction.to) {
       // looks like a contract interaction
-      baseTx = await this._populateTransaction(transaction, 0);
+      if (!baseTx.gasLimit) {
```
jaycenhorton, line 292: any inclination as to why this wouldn't exist?

plugins/fireblocks/from-upstream/fireblocks-bridge.ts
```diff
@@ -23,7 +23,24 @@ export class EthersCustomBridge extends BaseBridge {
...
       },
-      gasPrice: formatUnits(transaction?.gasPrice?.toString() || '0', 'gwei'),
+      gasPrice:
+        transaction.gasPrice != undefined
+          ? formatUnits(transaction?.gasPrice?.toString() || '0', 'gwei')
+          : undefined,
+      maxFee:
+        transaction.maxFeePerGas != undefined
+          ? formatUnits(
+              (await transaction.maxFeePerGas!.toString()) || '0',
```
jaycenhorton, line 33: does this need to be awaited?

plugins/fireblocks/from-upstream/fireblocks-bridge.ts
```diff
@@ -40,6 +57,7 @@ export class EthersCustomBridge extends BaseBridge {
         contractCallData: transaction.data,
       },
     };
+    console.log(txArguments);
```
jaycenhorton, line 60: can we use hre.trace?

utils/gas.ts
```diff
@@ -1,38 +1,75 @@
...
+  const response = await fetch(url);
+  const fees: GasStationResponse = await response.json();
+  const feeLevel: GasPrice = fees[level];
+  return {
+    maxFee: Math.round(feeLevel.maxFee),
+    maxPriorityFee: Math.round(feeLevel.maxPriorityFee),
+  };
+};
+
+export async function getGasFeeSettings(
```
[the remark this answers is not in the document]
jaycenhorton: key line would probably be [this line](https://github.com/nori-dot-eco/contracts/blob/eb6ee0bfcecce904a61d9134b7feefd245fbd3be/plugins/fireblocks/index.ts#L48)

plugins/fireblocks/from-upstream/fireblocks-bridge.ts
scottynomad: nope, fixed.

utils/gas.ts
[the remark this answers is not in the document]
scottynomad: Yeah we need to do something like that.  This is actually maps on to `getFeeData` in JsonRPCProvider.  I'd like to figure out a way to be able to pass the fee estimator in without packaging it with the fireblocks signer ...  will play with this a bit more.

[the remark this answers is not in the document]
scottynomad: Thanks for the idea, I just pushed a first pass at it.  Blocked on testing it right now because the mumbai gas station api SSL cert is expired. :/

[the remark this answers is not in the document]
jaycenhorton: looks much better!

jaycenhorton (approved):

PR #267, opened 2022-06-20
WIP: Add foundry to configuration.  

aqualen:
Note that now getting fatal errors during `yarn install`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -102,3 +102,6 @@ grants.csv
 
 # tenderly
 scripts/tenderly.ts
+
+# foundry
+out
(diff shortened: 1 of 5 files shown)


scottynomad:
@aqualen I got this to work after a `yarn clean`.  It doesn't actually do anything useful at this point (other than compile our contracts) until we write some tests in forge tests. And worse it hijacks the `hardhat test` command and no longer runs the typescript tests.

I think we're going to have to leave this open as an experimental branch for now until this plugin stabilizes some more.

scottynomad:
This can be closed.

PR #269, opened 2022-06-27
refactor and fix all the things to prepare for audit

jaycenhorton:
- [NO-1528] 
- [NO-1869]
- many more pre-audit changes/fixes

[NO-1528]: https://norinauts.atlassian.net/browse/NO-1528?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[NO-1869]: https://norinauts.atlassian.net/browse/NO-1869?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


contracts/FIFOMarket.sol
```diff
@@ -536,6 +511,28 @@ contract FIFOMarket is
...
+    address supplierAddress = removalId.supplierAddress();
+    uint256 removalBalance = _removal.balanceOf(address(this), removalId);
+    if (amount == removalBalance) {
+      require(
+        _activeSupply[supplierAddress].removeRemoval(removalId) ||
+          _reservedSupply.remove(removalId),
+        "Market: removal not in supply"
+      ); // todo custom error
+    }
+    totalReservedSupply += removalBalance; // todo should rely on balanceOf?
```
amiecorso, line 527: I don't think we want to increment this variable, because releasing a removal is not the same thing as reserving it.  reserving it moves it to a different pool where it's still valid and can be moved back into the active pool later.  releasing a removal means that carbon can never be sold.

contracts/Removal.sol
```diff
@@ -284,6 +295,32 @@ contract Removal is
...
+  function release(
+    address owner,
+    uint256 removalId,
+    uint256 amount
+  ) external whenNotPaused onlyRole(RELEASER_ROLE) {
+    // todo initialBalanceOf? should be amount + released?
+    // Querying the details of a removal returns a flag showing the quantity that was invalidated.
+    // If this Removal has not been completely sold, we will not sell the invalidated amount of that Removal
+    _tokenIdToReleased[removalId] = amount;
+    burn(owner, removalId, amount);
```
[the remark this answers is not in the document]
jaycenhorton: Not sure I follow

Why wouldn't we know how much to burn/release?

contracts/FIFOMarket.sol
```diff
@@ -536,6 +511,28 @@ contract FIFOMarket is
     return true; // returns true if the value was added to the set, that is, if it was not already present
   }
 
+  // todo consider making this a generalized `withdrawRemoval`?
+  function release(uint256 removalId, uint256 amount) external returns (bool) {
+    // todo whenNotPaused
+    // todo require sender is Removal address
+    address supplierAddress = removalId.supplierAddress();
+    uint256 removalBalance = _removal.balanceOf(address(this), removalId);
+    if (amount == removalBalance) {
```
amiecorso, line 520: I see that the way this is called from within `Removal.release()` is that the full amount being released is passed to this function as well.   It seems to me that there's a difference though between 1) creating a record of the full amount of this removal that has been released (so that's what's going in the `_tokenIdToReleased` map) and 2) performing the necessary actions to remedy things, such as burning any possible unsold balance of that removal, and then, if there wasn't enough unsold balance, making the replacements to any certificates that had consumed the released removal (not in scope for this PR)

All that to say that I think this if check should actually be asking if `(amount >= removalBalance)` instead of `==`, which allows us to still pass the full released amount, recognizing that it might actually be overshooting the unsold removal balance, in which case we still want to pull the removal id out of any supply pools we have.

contracts/Removal.sol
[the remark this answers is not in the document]
amiecorso: so the issue is that `amount` represents the total amount of that removal's released carbon, regardless of whether that carbon has already been sold or not (and we use the mapping to store that information for accountability and transparency).  However, this total released value isn't necessarily the same amount that should be getting burned because it's possible that the removal has been totally or partially sold through, and that there isn't enough remaining balance on the removal to burn the full released amount (this represents the case where certificates have to be amended)

contracts/Removal.sol
```diff
@@ -284,6 +290,32 @@ contract Removal is
...
+   *
+   * - The contract must not be paused.
+   * - The caller must have the `RELEASER_ROLE`.
+   * - The `amount` for the removal must be <= the removal's initial amount
+   */
+  function release(
+    address owner,
+    uint256 removalId,
+    uint256 amount
+  ) external whenNotPaused onlyRole(RELEASER_ROLE) {
```
amiecorso, line 308: we probably want to emit events from here too

contracts/Removal.sol
```diff
@@ -90,18 +92,26 @@ contract Removal is
...
-    external
-    onlyRole(DEFAULT_ADMIN_ROLE)
-  {
-    _restrictedNori = RestrictedNORI(restrictedNORIAddress);
+  function registerContractAddresses(
+    RestrictedNORI restrictedNoriAddress_,
+    FIFOMarket marketAddress_
+  ) external onlyRole(DEFAULT_ADMIN_ROLE) {
+    _restrictedNori = RestrictedNORI(restrictedNoriAddress_);
+    _market = marketAddress_;
```
amiecorso, line 106: do we not need to cast this the way `RestrictedNORI` is above?

contracts/Removal.sol
```diff
@@ -179,14 +189,11 @@ contract Removal is
     view
     returns (uint256[] memory)
   {
-    uint256[] memory holdbackPercentages = new uint256[](removalIds.length);
-    for (uint256 i = 0; i < removalIds.length; i++) {
+    uint256 numberOFRemovals = removalIds.length;
```
amiecorso, line 192: typo `numberOFRemovals` -> `numberOfRemovals`

contracts/Removal.sol
jaycenhorton: nope!

contracts/LockedNORILib.sol
```diff
@@ -59,26 +60,26 @@ library ScheduleUtils {
     Schedule storage schedule,
     uint256 time,
     uint256 amount
-  ) internal {
+  ) public {
```
scottynomad, line 63: @jaycenhorton This change from internal to public causes the linking error because it makes this no longer a purely functional library that can be interpolated into the using contract.

contracts/Market.sol
```diff
@@ -0,0 +1,857 @@
...
+    bytes memory
+  ) external returns (bytes4) {
+    // todo revert if not removal.sol
+    // todo whennotpaused
+    for (uint256 i = 0; i < ids.length; i++) {
+      uint256 removalToAdd = ids[i];
+      address supplierAddress = RemovalIdLib.supplierAddress(removalToAdd);
+      _activeSupply[supplierAddress].insertRemovalByVintage(removalToAdd);
+      if (
+        _suppliersInRoundRobinOrder[supplierAddress].nextSupplierAddress ==
```
scottynomad, line 215: check

contracts/Market.sol
```diff
@@ -0,0 +1,857 @@
...
+    private
+    returns (
+      uint256,
+      uint256[] memory,
+      uint256[] memory,
+      address[] memory
+    )
+  {
+    uint256 remainingAmountToFill = certificateAmount;
+    uint256 numberOfActiveRemovalsInMarket = _removal
```
scottynomad, line 411: we have a view function that does this, can we use that?

contracts/Market.sol
```diff
@@ -0,0 +1,857 @@
...
+        } else if (
+          /**
+           *  If the supplier is the only supplier remaining with supply, don't bother incrementing.
+           */
+          _suppliersInRoundRobinOrder[_currentSupplierAddress]
+            .nextSupplierAddress != _currentSupplierAddress
+        ) {
+          _incrementCurrentSupplierAddress();
+        }
+      }
```
scottynomad, line 460: I'd like to see the following code refactored so that the maintenance of the queue and round robin pointers lives in separate functions that are well documented.

contracts/Market.sol
```diff
@@ -0,0 +1,857 @@
...
+    uint256 numberOfRemovalsForOrder = 0;
+    // TODO (Gas Optimization): Declare variables outside of loop
+    for (
+      uint256 i = 0;
+      i < numberOfActiveRemovalsInMarket;
+      i++ // todo ++i consistency
+    ) {
+      uint256 removalId = _activeSupply[_currentSupplierAddress]
+        .getNextRemovalForSale();
+      uint256 removalAmount = _removal.balanceOf(address(this), removalId); // todo batch
```
scottynomad, line 427: We might also considering moving most everything from here down info a helper with a signature along the lines of:

`sellQuantityFromRemoval(uint256 quantity, RemovalID removalId)`

contracts/Market.sol
```diff
@@ -0,0 +1,857 @@
...
+   * @dev Determines the removal ids and amounts to fill the given purchase quantity, sourcing only
+   * from a single supplier.
+   *
+   * @param certificateAmount The number of carbon removals to purchase.
+   * @param supplier The supplier from which to purchase carbon removals.
+   * @return numberOfRemovalForOrder The number of distinct removal token ids used to fulfill this order.
+   * @return ids An array of the removal token ids being drawn from to fulfill this order.
+   * @return amounts An array of amounts being allocated from each corresponding removal token.
+   */
+  function _allocateSupplySingleSupplier(
```
scottynomad, line 480: My suggestion above would remove most of the duplicated code here.

scottynomad (commented):
Still working through my review but wanted to get these comments out ...

contracts/Removal.sol
```diff
@@ -173,126 +234,144 @@ contract Removal is
...
-      scheduleStartTime,
-      firstId.supplierAddress(), // todo check gas costs of extracting values per line vs once using unpackRemovalIdV0
-      firstId.methodology(), // todo check gas costs of extracting values per line vs once using unpackRemovalIdV0
-      firstId.methodologyVersion() // todo check gas costs of extracting values per line vs once using unpackRemovalIdV0
-    );
-    super._mintBatch(to, ids, amounts, data);
-    setApprovalForAll(to, _msgSender(), true); // todo look at vesting contract for potentially better approach
-    if (list) {
-      safeBatchTransferFrom(to, marketAddress, ids, amounts, data);
+    uint256[] memory totals = balanceOfBatch(owners, removals.values());
```
scottynomad, line 305: This will grow linearly with the total supply in the market which is not going to work.  I think we're going to need to maintain running balances of these as we go.

https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/utils/structs/EnumerableSet.sol#L351-L354

contracts/Removal.sol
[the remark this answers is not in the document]
jaycenhorton: @amiecorso is this still an issue in the current commit?

contracts/FIFOMarket.sol
jaycenhorton: @amiecorso is this still an issue in the latest commit?

contracts/Removal.sol
jaycenhorton: Now that the only time a removal is sent to the 0 address is when releasing happens, i think it's appropriate to just rely on `TransferSingle`/`TransferBatch` where `to == address(0)`. What do you think @amiecorso ?

contracts/Certificate.sol
```diff
@@ -52,79 +73,250 @@ contract Certificate is ERC1155PresetPausableNonTransferrable {
     _disableInitializers();
   }
 
-  function initialize() external initializer {
-    // todo verify all inherited unchained initializers are called
+  function initialize() external initializerERC721A initializer {
+    // todo validate initializers
     __Context_init_unchained();
     __ERC165_init_unchained();
+    __ERC721A_init_unchained("Certificate", "NRT");
```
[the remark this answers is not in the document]
jaycenhorton: Definitely a placeholder. What should the symbol for a certificate be?

contracts/Certificate.sol
```diff
@@ -52,79 +73,250 @@ contract Certificate is ERC1155PresetPausableNonTransferrable {
...
+
+  function releaseRemoval(
+    uint256 certificateId,
+    uint256 removalId,
+    uint256 amount
+  ) external whenNotPaused {
+    if (_msgSender() != address(_removal)) {
+      revert SenderNotRemovalContract();
+    }
+    // todo emit event?
```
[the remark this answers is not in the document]
jaycenhorton: potentially redundant. [see](https://github.com/nori-dot-eco/contracts/pull/269#discussion_r930302863)

contracts/Certificate.sol
```diff
@@ -1,49 +1,70 @@
...
+    keccak256("CERTIFICATE_OPERATOR_ROLE");
 
-  /**
-   * @dev a mapping of the certificate token ID -> sources
-   */
-  mapping(uint256 => Source[]) private _sources;
+  mapping(uint256 => mapping(uint256 => uint256))
+    private _removalBalancesOfCertificate;
+
+  mapping(uint256 => EnumerableSetUpgradeable.UintSet)
```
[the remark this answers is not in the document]
jaycenhorton: interesting idea..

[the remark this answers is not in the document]
jaycenhorton: added this to the investigation todo list to track it

contracts/Certificate.sol
```diff
@@ -1,49 +1,70 @@
...
+    private _removalsOfCertificate; // todo tests that ensure this is maintained correctly
+
+  mapping(uint256 => EnumerableSetUpgradeable.UintSet)
```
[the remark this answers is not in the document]
jaycenhorton: added this to the investigation todo list to track it

contracts/Market.sol
jaycenhorton: view function removed in #300

contracts/Market.sol
jaycenhorton: tracked

contracts/Market.sol
```diff
@@ -0,0 +1,857 @@
...
+   * @param certificateAmount The number of carbon removals to purchase.
+   * @return numberOfRemovalForOrder The number of distinct removal token ids used to fulfill this order.
+   * @return ids An array of the removal token ids being drawn from to fulfill this order.
+   * @return amounts An array of amounts being allocated from each corresponding removal token.
+   * @return suppliers The address of the supplier who owns each corresponding removal token.
+   */
+  function _allocateSupplyRoundRobin(uint256 certificateAmount)
+    private
+    returns (
+      uint256,
```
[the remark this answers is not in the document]
jaycenhorton: yeah-- prevents needing to call the removal contract multiple times as it's a costly operation

contracts/Market.sol
jaycenhorton: tracked

contracts/Removal.sol
jaycenhorton: tracked

scottynomad (approved):
Let's get this merged and go from there on the other cleanups.

contracts/Certificate.sol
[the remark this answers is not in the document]
jaycenhorton: changed :+1:

contracts/Market.sol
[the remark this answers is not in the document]
jaycenhorton: Tagging @amiecorso here as I believe she wrote this

contracts/Removal.sol
amiecorso: I think that makes sense.  since those events also include the `from` being burned from, we can use that to be able to see whether it's from the market, from certificates, or from unlisted balance that belonged to the original supplier.  I do think that we will probably want to emit events from the `Certificate.releaseRemoval` function that indicate which certificate ids are having some removal id released, because that level of granularity won't be captured by the `TransferSingle`/`TransferBatch` events since the Certificate contract itself is the one that maintains the bookkeeping at that granularity.

contracts/Certificate.sol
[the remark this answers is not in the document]
amiecorso: also in favor of emitting the event as described in my response to that comment

contracts/Removal.sol
```diff
@@ -173,126 +234,144 @@ contract Removal is
...
-    uint256[] calldata removalIds,
-    uint256 holdbackPercentage
-  ) external {
-    for (uint256 i = 0; i < removalIds.length; i++) {
-      uint256 id = removalIds[i];
-      if (!_tokenIdExists[id]) {
-        revert TokenIdDoesNotExist({tokenId: id});
-      }
-      _removalIdToRemovalData[id].holdbackPercentage = holdbackPercentage;
+  // todo this function will not scale well- consider dropping it somehow
```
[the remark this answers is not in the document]
jaycenhorton: I believe it's removed in the withdraw branch-- fyi The certificate will also own a large portion of removals as well

rlfarman (commented):
Dropping my comments so far. Still much more to review

contracts/Removal.sol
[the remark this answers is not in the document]
amiecorso: discussed on a call and you're addressing in #305

contracts/FIFOMarket.sol
amiecorso: discussed on a call and you're addressing in #305

contracts/Market.sol
```diff
@@ -0,0 +1,857 @@
...
+  {
+    _priorityRestrictedThreshold = threshold;
+    emit PriorityRestrictedThresholdSet(threshold);
+  }
+
+  /**
+   * The amount of supply available for anyone to buy.
+   */
+  function totalUnrestrictedSupply() external view returns (uint256) {
+    uint256 activeSupply = _removal.cumulativeBalanceOf(address(this));
```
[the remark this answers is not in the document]
jaycenhorton: removed in #300

contracts/LockedNORILib.sol
```diff
@@ -64,21 +65,21 @@ library ScheduleUtils {
...
       require(
         time >= schedule.cliffs[cliffCount - 1].time,
-        "ScheduleUtils: Cliffs not chronological"
+        "LockedNORILib: Cliffs not chronological"
       );
     }
     require(
       time <= schedule.endTime,
-      "ScheduleUtils: Cliffs cannot end after schedule"
+      "LockedNORILib: Cliffs cannot end after schedule" // todo custom errors?
```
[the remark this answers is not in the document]
jaycenhorton: already tracked

jaycenhorton:
Added a note to #301 to address outstanding/unresolved conversations here

PR #271, opened 2022-06-27
[NO-1742] Holdback percentage and restricted token routing for removals

amiecorso:
NO-1742
NO-1745

Sets up the removal contract to assign holdback percentages to each project when minting removals.  When removals are sold, this holdback percentage is used to determine how much of any given removal's proceeds should be routed to the `RestrictedNORI` contract.


contracts/FIFOMarket.sol
```diff
@@ -301,16 +301,35 @@ contract FIFOMarket is
...
+
     // TODO (Gas Optimization): Declare variables outside of loop
     for (uint256 i = 0; i < batchedIds.length; i++) {
       if (batchedAmounts[i] == batchedBalances[i]) {
         totalNumberActiveRemovals -= 1; // removal used up
       }
       totalActiveSupply -= batchedAmounts[i];
-      uint256 noriFee = (batchedAmounts[i] / 100) * _noriFee;
-      uint256 supplierFee = batchedAmounts[i];
+      uint256 noriFee = (batchedAmounts[i] * _noriFee) / 100;
```
amiecorso, line 314: should divide after multiplying to avoid precision errors / safe math rounding to 0

PR #275, opened 2022-06-29
[NO-1742] Holdback percentages, restricted token routing, `rNORI` update to be ERC-20 compatible

amiecorso:
NO-1742
NO-1745

Updates the `RestrictedNORI` contract to be ERC-20 compatible instead of ERC-777. (necessary to do this here to allow restricted proceed routing to happen at all without ERC-777)

Sets up the removal contract to assign holdback percentages to each project when minting removals. When removals are sold, this holdback percentage is used to determine how much of any given removal's proceeds should be routed to the RestrictedNORI contract.

New version of #271  , but based off of Jaycen's erc20 branch.


contracts/FIFOMarket.sol
```diff
@@ -302,16 +302,34 @@ contract FIFOMarket is
...
+      uint256 unrestrictedSupplierFee = batchedAmounts[i];
+      if (holdbackPercentages[i] > 0) {
+        restrictedSupplierFee =
+          (unrestrictedSupplierFee * holdbackPercentages[i]) /
+          100;
+        unrestrictedSupplierFee =
+          unrestrictedSupplierFee -
+          restrictedSupplierFee;
+        _restrictedNori.mint(restrictedSupplierFee, batchedIds[i]);
+        _bridgedPolygonNori.transfer(
```
jaycenhorton, line 326: can you add a todo to extract these to a single batch call?

test/helpers/restricted-nori.ts
```diff
@@ -33,16 +30,14 @@ export const restrictRemovalProceeds = async ({
...
-        removalAmountsToRestrict[index],
-        userData
-      );
-    })
-  );
+  const promiseArray = [];
+  for (const [index, id] of removalIds.entries()) {
+    promiseArray.push(
+      rNori.mint(removalAmountsToRestrict[index], id),
+      bpNori.transfer(rNori.address, removalAmountsToRestrict[index])
```
jaycenhorton, line 37: I don't think this is doing what you think. Just because `await` is omitted doesnt mean the function wasn't called. As such, the await after the loop could end up with an array of pending/fulfilled promises.

test/helpers/index.ts
```diff
@@ -239,18 +239,26 @@ export const createBatchMintData = async ({
...
   projectId?: number;
   scheduleStartTime?: number;
+  holdbackPercentage?: number;
 }): Promise<Parameters<Removal['mintBatch']>[3]> => {
   const actualScheduleStartTime =
     scheduleStartTime ?? (await getLatestBlockTime({ hre }));
   const packedData = hre.ethers.utils.defaultAbiCoder.encode(
-    ['uint256', 'uint256', 'address', 'bool'],
-    [projectId, actualScheduleStartTime, fifoMarket.address, listNow]
+    ['uint256', 'uint256', 'uint256', 'address', 'bool'],
```
jaycenhorton, line 254: heads up that this syntax exists and could make life easier going forward

https://forum.openzeppelin.com/t/erc20-batchtransfer-functionality/6978/6

jaycenhorton (commented):
Couple comments, but shouldn't be anything major!

test/helpers/restricted-nori.ts
amiecorso: isn't this basically the same thing that's happening when you use .map and generate an array of promises?  it's just that extra promises are being added to the array? the functions are being called but nothing is being awaited serially, and then it's just a typical `Promise.all` on all those promises.

if I still have this wrong, do you know the way to do this better?

test/helpers/index.ts
amiecorso: hmm, do you mean the use of `encodeABI()` (which seems to be a web3 function)? or using multicall? (which i'm not seeing how it fits into this step)

test/helpers/restricted-nori.ts
jaycenhorton:
```ts
params = [3000, 3000, 3000];
demoAsyncForLoop = async () => {
  promises = [];
  i = 0;
  for (const x of params) {
    promises.push(timeout(x, i));
    i++;
  }
  await Promise.all(promises);
};
demoPromiseAll = async () => {
  await Promise.all(
    params.map((p, index) => {
      return timeout(p, index);
    })
  );
};
```
image

test/helpers/index.ts
jaycenhorton: Ethers equivalent is `encodeFunctionData` I believe ([docs](https://docs.ethers.io/v5/api/utils/abi/interface/)), but you're right that it wouldn't be useful since I forgot we just accept a generic data param. Given that minting isn't part of the base erc 1155 spec, I think we should probably consider just dropping the mint inheritance like we did for rNORI and just accept a struct as the parameter rather than decoding it on-chain. Would clean up our code a lot and allow us to use objects instead of encoding ([example](https://ethereum.stackexchange.com/questions/118280/how-to-call-a-contract-function-from-reactjs-that-takes-a-struct-as-an-argument))

jaycenhorton: (not in the scope of this PR, just a comment for now -- maybe add a todo though?)

test/FIFOMarket.test.ts
```diff
@@ -2,6 +2,7 @@ import type { BigNumberish, BigNumber } from 'ethers';
 import { ContractReceipt } from 'ethers';
 import { add, multiply } from '@nori-dot-com/math';
 import { splitSignature } from 'ethers/lib/utils';
+import { compareScheduleSummaryStructs } from './helpers/restricted-nori';
```
jaycenhorton, line 5:
```suggestion
import { compareScheduleSummaryStructs } from '@/helpers/restricted-nori';
```

test/helpers/index.ts
amiecorso: oh nice... yeah it would definitely be cool to be able to work with structs/objects directly instead of encoded data.  not the first time you've recommended removing the inheritance of "mint"

test/helpers/restricted-nori.ts
jaycenhorton: I see what you're trying to say now. The problem is that this is still not following a great pattern though (for loops + declaring arrays and pushing into them is imperative and non-functional and should be avoided for their side effects that they introduce).  [[more](https://github.com/MostlyAdequate/mostly-adequate-guide/blob/822af0ce4a15b64ea6470f6dee02cf6fffd39793/ch06.md#declarative-coding)]

IMO, It should just be something like this

```ts
 await Promise.all(
    removalIds.map(async (id, index) => {
      return Promise.all([
        rNori.mint(removalAmountsToRestrict[index], id),
        bpNori.transfer(rNori.address, removalAmountsToRestrict[index]),
      ]);
    })
  );
```

jaycenhorton: [another good resource](https://github.com/you-dont-need/You-Dont-Need-Loops)

jaycenhorton (approved):

test/helpers/restricted-nori.ts
amiecorso: ahh, I see.  I agree that when I wrote this something about it felt wrong, even if it's technically doing what I want it to.  I see how this also works and doesn't unnecessarily serialize anything. updated.

jaycenhorton: Thanks for seeing this one through! 💯

PR #278, opened 2022-06-30
Introduce LockedNORIV2 and test upgrade path.

scottynomad:
Follow-ups to address in another PR:
- [ ] Removal of `ContractNames` parameter from `getContractFactory` generic type.
- [ ] Use permit to streamline and re-introduce batch grant creation.
- [ ] Custom errors.


contracts/LockedNORIV2.sol
```diff
@@ -0,0 +1,788 @@
...
+  /**
+   * @notice Emitted when the underlying token contract address is updated due to migration.
+   */
+  event UnderlyingTokenAddressUpdated(address from, address to);
+
+  /**
+   * @notice This function is triggered when BridgedPolygonNORI is sent to this contract
+   *
+   * @dev Sending BridgedPolygonNORI to this contract triggers the tokensReceived hook defined by the ERC-777 standard
+   * because this contract is a registered ERC777 tokens recipient.
```
amiecorso, line 226: could probably make this language less specific now?

contracts/LockedNORIV2.sol
```diff
@@ -0,0 +1,788 @@
...
+  event TokensClaimed(
+    address indexed from,
+    address indexed to,
+    uint256 quantity
+  );
+
+  /**
+   * @notice Emitted when the underlying token contract address is updated due to migration.
+   */
+  event UnderlyingTokenAddressUpdated(address from, address to);
```
amiecorso, line 220: probably ok that none of this is indexed since we expect there to be very few of these events?

amiecorso (approved):

contracts/LockedNORIV2.sol
scottynomad: oh interesting, yeah I would think this is exceedingly rare and probably not of interest to indexers.

contracts/LockedNORIV2.sol
scottynomad: Good call I need to go back and look at these docs.

scottynomad:
The LockedNORI tests are passing now.  The stacked branches make it difficult to see the changes made to `LockedNORI.sol` and `LockedNORI.test.ts`.  Once this merges down to jaycen's branch it should be somewhat easier to see.

contracts/LockedNORIV2.sol
```diff
@@ -0,0 +1,791 @@
...
+   */
+  function tokensReceived(
+    address,
+    address,
+    address,
+    uint256,
+    bytes calldata,
+    bytes calldata
+  ) external override {
+    revert("Funding via tokensReceived callback in not supported");
```
jaycenhorton, line 233: is this upgrade a good opportunity to add custom errors?

contracts/test/ScheduleTestHarness.sol
```diff
@@ -1,7 +1,7 @@
 // SPDX-License-Identifier: MIT
 pragma solidity =0.8.15;
 
-import {ScheduleUtils, Schedule} from "./ScheduleUtils.sol";
+import {ScheduleUtils, Schedule} from "../ScheduleUtils.sol";
```
jaycenhorton, line 4: Wonder if libraries can also be upgraded? could be nice to use [solidity global time units](https://docs.soliditylang.org/en/v0.8.15/units-and-global-variables.html#time-units)

contracts/test/LockedNORIV2Helper.sol
```diff
@@ -0,0 +1,124 @@
...
+
+  function get(address lnori, address recipient)
+    public
+    view
+    returns (LockedNORIV2.TokenGrantDetail memory)
+  {
+    return LockedNORIV2(lnori).getGrant(recipient);
+  }
+
+  function assertEq(
```
jaycenhorton, line 100: :drooling_face:

contracts/test/DSTest/test.sol
```diff
@@ -0,0 +1,469 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
```
jaycenhorton, line 1: was this actually needed? [This example repo](https://github.com/gakonst/foundry-rust-template/blob/master/contracts/src/test/Contract.t.sol) makes it look like you can just import this file using `import "ds-test/test.sol";`

The foundry book says it exists at a [different import](https://book.getfoundry.sh/forge/writing-tests.html?highlight=dstest#writing-tests) though

plugins/index.ts
```diff
@@ -56,14 +56,15 @@ extendEnvironment((hre) => {
   // All live networks will try to use fireblocks
   if (Boolean(hre.config.fireblocks.apiKey) && hre.network.config.live) {
     hre.getSigners = lazyFunction(() => hre.fireblocks.getSigners);
-    // todo namedFireblocksAccounts
-    // todo namedFireblocksSigners
     hre.log('Installed fireblocks signers');
+  } else if (hre.network.name != 'hardhat') {
```
jaycenhorton, line 60: can we also include `localhost`?

test/LockedNORIUpgrade.test.ts
```diff
@@ -0,0 +1,85 @@
...
+
+    // Create state is LockedNORI v1
+
+    await helper.createSimplePastGrant(
+      lnori.address,
+      ethers.utils.parseEther('100'),
+      recipient
+    );
+    const createdGrantDetail = await helper.get(lnori.address, recipient);
+    helper.assertSimplePastGrant(lnori.address, createdGrantDetail);
```
jaycenhorton, line 49: missing await?

test/LockedNORIUpgrade.test.ts
```diff
@@ -0,0 +1,85 @@
...
+    expect(await lnori.balanceOf(recipient)).to.eq(GRANT_AMOUNT);
+
+    // Now we upgrade
+
+    const LockedNORIV2Factory = await ethers.getContractFactory('LockedNORIV2');
+    const lNoriV2 = (await upgrades.upgradeProxy(
+      lnori.address,
+      LockedNORIV2Factory
+    )) as LockedNORIV2;
+    lNoriV2.updateUnderlying(bpNori.address);
```
jaycenhorton, line 72: missing await?

test/LockedNORIUpgrade.test.ts
```diff
@@ -0,0 +1,85 @@
...
+
+    const LockedNORIFactory = await hre.ethers.getContractFactoryFromArtifact(
+      LockedNORIArtifact
+    );
+    const lnori = (await upgrades.deployProxy(
+      LockedNORIFactory,
+      [testTokenInstance.address],
+      { initializer: 'initialize(address)' }
+    )) as LockedNORI;
+    lnori.grantRole(await lnori.TOKEN_GRANTER_ROLE(), helper.address);
```
jaycenhorton, line 39: missing await?

types/global.d.ts
```diff
@@ -191,7 +196,7 @@ declare global {
     getContractFactory<
       TContractFactory extends ContractFactory = ContractFactory
     >(
-      name: ContractNames,
+      name: string,
```
jaycenhorton, line 199: I can see how this generic can be frustrating to work with. I wrote this so to add type support and intellisense support when writing contracts deployments.


It also once helped with type inference when deploying contracts so you don't have to do `as ContractName` but that part appears to have broke (for some reason ContractFactory is actually causing a type error because it's not actually imported in global.d.ts and I think that's causing ts to infer a type of `any`)

Did you run  into issues with this? If it can't be reverted back to `ContractNames`, we should just delete this and stop trying to augment the types. 

(it's also possible the type augmentation i added here isn't required anymore and inference might be working by default in the latest types of various packages)

contracts/deprecated/ERC777PresetPausablePermissioned.sol
```diff
@@ -0,0 +1,284 @@
+// SPDX-License-Identifier: MIT
```
jaycenhorton, line 1: did this contract change at all?

contracts/LockedNORIV2.sol
jaycenhorton: I think you can actually just remove this function, unset `ERC777_TOKENS_RECIPIENT_HASH` in the 1820 registry and override any of they 777 recipient stuff if needed

contracts/LockedNORIV2.sol
```diff
@@ -0,0 +1,791 @@
...
+   *
+   * @param amount uint256 Quantity of `_bridgedPolygonNori` to deposit
+   */
+  function depositFor(address recipient, uint256 amount)
+    external
+    whenNotPaused
+    returns (bool)
+  {
+    require(_grants[recipient].exists, "lNORI: Cannot deposit without a grant");
+    _bridgedPolygonNori.transferFrom(_msgSender(), address(this), amount);
```
jaycenhorton, line 249: do we want to use this instead of the permit pattern?

- using this approach: 2 serial transactions (approve + depositFor) (although it might be possible to only ever have to call `approve` once if we give it a high enough balance, erc 20 allows finfinite approvals even if the amount is > the current accounts balance)
- use erc 20 permit: 1 signature, and 1 transactions. requires figuring out typed data signing with fireblocks. Definitely safer, than infinite approvals, and fewer transactions (although signing would feel the same as a transaction in fireblocks since it still requires confirming in the app)

jaycenhorton: I think you'd actually have to also call `createGrant` using the non-permit pattern, right? 3 total transactions I think



`createGrant` -> `approve` -> `depositFor`? Is there a way to simplify that to two?



Permit could be done in 1 signature + 1 tx

contracts/LockedNORIV2.sol
```diff
@@ -0,0 +1,791 @@
...
+   * @notice Admin function to update the underlying token contract address.
+   *
+   * @dev Used in case of major migrations only.
+   */
+  function updateUnderlying(IERC20Upgradeable bridgedPolygonNori)
+    external
+    whenNotPaused
+    onlyRole(DEFAULT_ADMIN_ROLE)
+  {
+    address old = address(_bridgedPolygonNori);
```
jaycenhorton, line 465: perhaps `require(old !== new)`

test/LockedNORI.test.ts
```diff
@@ -0,0 +1,1166 @@
+import { BigNumber } from 'ethers';
```
jaycenhorton, line 1: https://www.diffchecker.com/R75nkGLJ

test/LockedNORI.test.ts
```diff
@@ -0,0 +1,1166 @@
...
+        it(`accounts with the role "${role}" can use "batchRevokeUnvestedTokenAmounts" whilst accounts without the role "${role}" cannot`, async () => {
+          const { lNori, grantAmount, grant, hre } = await setupWithGrant(
+            (parameters) => employeeParameters(parameters)
+          );
+          const { namedAccounts, namedSigners } = hre;
+          const roleId = await lNori[role]();
+          expect(await lNori.hasRole(roleId, namedAccounts[accountWithRole])).to
+            .be.true;
+          expect(
+            await lNori
```
jaycenhorton, line 377: [await order is important for this](https://ethereum-waffle.readthedocs.io/en/latest/matchers.html#emitting-events)
```suggestion
          await expect(
            lNori
```

test/LockedNORI.test.ts
```diff
@@ -0,0 +1,1166 @@
...
+          const { namedAccounts, namedSigners } = hre;
+          const roleId = await lNori[role]();
+          expect(await lNori.hasRole(roleId, namedAccounts[accountWithoutRole]))
+            .to.be.false;
+          const { grant, grantAmount } = employeeParameters({
+            hre,
+            startTime: await getLatestBlockTime({ hre }),
+          });
+          expect(
+            await lNori
```
jaycenhorton, line 419: [await order is important for this](https://ethereum-waffle.readthedocs.io/en/latest/matchers.html#emitting-events)
```suggestion
          await expect(
            lNori
```

test/LockedNORI.test.ts
```diff
@@ -0,0 +1,1166 @@
...
+      grantAmount.sub(withdrawlAmount)
+    );
+    expect(await lNori.vestedBalanceOf(investor1)).to.equal(
+      grantAmount.sub(withdrawlAmount)
+    );
+    expect(await lNori.unlockedBalanceOf(investor1)).to.equal(
+      grantAmount.sub(withdrawlAmount)
+    );
+    expect(
+      await lNori
```
jaycenhorton, line 774:
```suggestion
    await expect(
      lNori
```

contracts/LockedNORIV2.sol
scottynomad: Good point, I removed the ERC777Recipient inheritance and this function so that sends will fail due to it not being present.  It makes sense to unregister the proxy from the 1820 registry as well.

contracts/test/ScheduleTestHarness.sol
scottynomad: I remember you pointed this out before but I wasn't clear where it was relevant since this library only deals in seconds anyway.

contracts/LockedNORIV2.sol
```diff
@@ -0,0 +1,791 @@
...
+
+  /**
+   * @notice A mapping from grantee to grant
+   */
+  mapping(address => TokenGrant) private _grants;
+
+  /**
+   * @notice The BridgedPolygonNORI contract that this contract wraps tokens for
+   */
+  IERC20Upgradeable private _bridgedPolygonNori;
```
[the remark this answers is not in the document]
scottynomad: Sure, could do.  I figured it might be more self-explanatory / self-contained this way since we only interact with BridgedPolygonNORI as an ERC20 but I'm not strongly attached to that idea.

I'm aware that once I start fixing the vesting task we may want to add bulk grant creation and deposit functions in here which may change that story.

contracts/LockedNORIV2.sol
scottynomad: Yeah definitely intending to revisit this now that we know this upgrade is working cleanly.

test/LockedNORIUpgrade.test.ts
```diff
@@ -0,0 +1,85 @@
+import { upgrades } from 'hardhat';
+import { expect, setupTest } from '@/test/helpers';
+import { LockedNORIV2 } from '@/typechain-types';
+// Not sure if this will exist in a clean build.  Does typechain build from artifacts or source?
+import { LockedNORI } from '../typechain-types/contracts/LockedNORI';
+import LockedNORIArtifact from '../legacy-artifacts/LockedNORI.sol/LockedNORI.json';
+import { TestToken777 } from '../typechain-types/contracts/test/TestToken777';
```
jaycenhorton, line 7:
```suggestion
import { LockedNORI } from '.@/typechain-types/contracts/LockedNORI';
import LockedNORIArtifact from '@/legacy-artifacts/LockedNORI.sol/LockedNORI.json';
import { TestToken777 } from '@/typechain-types/contracts/test/TestToken777';
```

contracts/LockedNORIV2.sol
scottynomad: Will address custom errors into a later PR.

contracts/test/LockedNORIV2Helper.sol
```diff
@@ -0,0 +1,124 @@
...
+  ) public {
+    LockedNORIV2.TokenGrantDetail memory actualGrantDetails = LockedNORIV2(
+      lnori
+    ).getGrant(expectedGrantDetails.recipient);
+    assertEq(actualGrantDetails, expectedGrantDetails);
+  }
+
+  event logNamedTokenGrant(string, LockedNORIV2.TokenGrantDetail grant);
+
+  //   struct TokenGrantDetail {
```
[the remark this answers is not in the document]
scottynomad: Nope, removed.

contracts/test/DSTest/test.sol
scottynomad: It is part of the foundry standard library but that's not landed yet.  I will plan to update the import once I get that merged in.

test/LockedNORIUpgrade.test.ts
```diff
@@ -0,0 +1,85 @@
+import { upgrades } from 'hardhat';
+import { expect, setupTest } from '@/test/helpers';
+import { LockedNORIV2 } from '../typechain-types';
+// Not sure if this will exist in a clean build.  Does typechain build from artifacts or source?
```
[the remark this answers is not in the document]
scottynomad: That seems to be how it's working. I'm going to remove this comment.

test/LockedNORIUpgrade.test.ts
scottynomad: out of practice! :)

types/global.d.ts
scottynomad: It had some trickle down effects with the renaming if I recall correctly.  Including `LockedNORI` as well as `LockedNORIV2` here causes it to clash with several other types that I'd like top leave `LockedNORI` out of.  Similar story for any solidity test contracts.  I will bookmark this to revisit when the dust settles as well.

contracts/deprecated/ERC777PresetPausablePermissioned.sol
scottynomad: Just the solidity version.

scottynomad:
See #279

PR #280, opened 2022-07-02
[NO-1835] Use custom errors wherever possible in market contracts

amiecorso:
Uses [custom errors](https://blog.soliditylang.org/2021/04/21/custom-errors/) wherever possible in the market smart contracts. Note that, as described in [this issue ](https://github.com/ethereum/solidity/issues/11278), there isn't currently good support for try/catch with custom errors.  For that reason, there are a few places where we still use revert strings, such that when one of our errors is thrown in the course of executing code that lives in the inherited contracts, such as the example below, it is properly recognized by the `catch Error(string memory reason)`.  Otherwise the custom error is not caught and resurfaced and we end up with the opaque error `revert("ERC1155: transfer to non ERC1155Receiver implementer");` in the final catch.

```
            try IERC1155ReceiverUpgradeable(to).onERC1155Received(operator, from, id, amount, data) returns (bytes4 response) {
                if (response != IERC1155ReceiverUpgradeable.onERC1155Received.selector) {
                    revert("ERC1155: ERC1155Receiver rejected tokens");
                }
            } catch Error(string memory reason) {
                revert(reason);
            } catch {
                revert("ERC1155: transfer to non ERC1155Receiver implementer");
            }
```

Another note: I've included an example of using a separate file to define shared custom errors that appear in more than one contract.  We may want to keep this pattern if it helps set a precedent for the most code reuse possible (and keeping names consistent, etc.), or we may decide it's unnecessary overhead and that all contracts should define any custom errors they use locally.


contracts/Certificate.sol
```diff
@@ -98,7 +99,7 @@ contract Certificate is ERC1155PresetPausableNonTransferrable {
     // todo is there a better way to verify that no removal amount == 0?
     for (uint256 i = 0; i < removalIds.length; i++) {
       if (removalAmounts[i] == 0) {
-        revert("Certificate: Removal amount 0");
+        revert RemovalAmount0({removalId: removalIds[i]});
```
jaycenhorton, line 102: this shouldn't be possible any more so we should just remove it

contracts/Certificate.sol
```diff
@@ -119,7 +120,7 @@ contract Certificate is ERC1155PresetPausableNonTransferrable {
     uint256,
     bytes memory
   ) public pure override {
-    revert ForbiddenFunctionCall(); // todo is this really what we want?
+    revert MintDisabled();
```
jaycenhorton, line 123: I would rather just change how this works. We should:
- stop inheriting `ERC1155PresetMinterPauserUpgradeable` in `ERC1155PresetPausableNonTransferrable`
- change the `mint` to allow minting `n` removals for one supplier
- change `mintBatch` to allow minting `n` removals for `m` suppliers
- perhaps this needs to be further extended one more dimension for projects?

contracts/Certificate.sol
```diff
@@ -8,19 +8,20 @@ import "@openzeppelin/contracts-upgradeable/token/ERC777/ERC777Upgradeable.sol";
 import "@openzeppelin/contracts-upgradeable/token/ERC777/IERC777RecipientUpgradeable.sol";
 import "@openzeppelin/contracts-upgradeable/utils/introspection/ERC1820ImplementerUpgradeable.sol";
 import "./ERC1155PresetPausableNonTransferrable.sol";
+import {MintDisabled} from "./SharedCustomErrors.sol";
```
jaycenhorton, line 11: if we're going to share this error, I think a generic `disabled function` error is better-- you can infer which function was disabled by the call itself

contracts/FIFOMarket.sol
```diff
@@ -176,10 +185,11 @@ contract FIFOMarket is
...
       address supplierAddress = removalToAdd.supplierAddress();
       uint256 removalAmount = batchedAmounts[i];
-      require(
-        _activeSupply[supplierAddress].insertRemovalByVintage(removalToAdd),
-        "Market: Unable to add removal by vintage" // TODO (Gas Optimization): Use custom error
-      );
+      if (
+        !_activeSupply[supplierAddress].insertRemovalByVintage(removalToAdd)
+      ) {
+        revert FailedToAddRemovalToActiveSupply({removalId: removalToAdd});
```
jaycenhorton, line 191: rather than returning `bool` from `insertRemovalByVintage` should we just revert if it's not possible?

contracts/FIFOMarket.sol
```diff
@@ -347,16 +356,15 @@ contract FIFOMarket is
...
         ids[numberOfRemovals] = removalId;
         amounts[numberOfRemovals] = removalAmount; // this removal is getting used up
         suppliers[numberOfRemovals] = _currentSupplierAddress;
         remainingAmountToFill -= removalAmount;
-        require(
-          _activeSupply[_currentSupplierAddress].removeRemoval(removalId),
-          "Market: Failed to remove removal from supply"
-        ); // TODO (Gas Optimization): Use custom error
+        if (!_activeSupply[_currentSupplierAddress].removeRemoval(removalId)) {
+          revert FailedToRemoveRemovalFromSupply({removalId: removalId});
```
jaycenhorton, line 366: https://github.com/nori-dot-eco/contracts/pull/280/files#r913286493

contracts/FIFOMarket.sol
```diff
@@ -404,7 +412,9 @@ contract FIFOMarket is
         .queueByVintage[vintage]
         .length();
     }
-    require(totalNumberOfRemovalsForSupplier > 0, "Market: Not enough supply");
+    if (!(totalNumberOfRemovalsForSupplier > 0)) {
```
jaycenhorton, line 415:
```suggestion
    if (totalNumberOfRemovalsForSupplier == 0) {
```

contracts/FIFOMarket.sol
```diff
@@ -424,15 +434,14 @@ contract FIFOMarket is
...
         }
         ids[numberOfRemovals] = removalId;
         amounts[numberOfRemovals] = removalAmount; // this removal is getting used up
         remainingAmountToFill -= removalAmount;
-        require(
-          supplierRemovalQueue.removeRemoval(removalId),
-          "Market: Failed to remove removal from supply"
-        ); // TODO (Gas Optimization): Use custom error
+        if (!supplierRemovalQueue.removeRemoval(removalId)) {
+          revert FailedToRemoveRemovalFromSupply({removalId: removalId});
```
jaycenhorton, line 443: https://github.com/nori-dot-eco/contracts/pull/280/files#r913286493

contracts/FIFOMarket.sol
```diff
@@ -519,10 +528,9 @@ contract FIFOMarket is
    */
   function reserveRemoval(uint256 removalId) external returns (bool) {
     address supplierAddress = removalId.supplierAddress();
-    require(
-      _activeSupply[supplierAddress].removeRemoval(removalId),
-      "Market: removal not in active supply"
-    ); // TODO (Gas Optimization): Use custom error
+    if (!_activeSupply[supplierAddress].removeRemoval(removalId)) {
+      revert RemovalNotInActiveSupply({removalId: removalId});
```
jaycenhorton, line 532: https://github.com/nori-dot-eco/contracts/pull/280/files#r913286493

contracts/FIFOMarket.sol
```diff
@@ -532,7 +540,9 @@ contract FIFOMarket is
       _removeActiveSupplier(supplierAddress);
     }
     // todo any checks on whether this id was already in there?
-    require(_reservedSupply.add(removalId), "Market: Removal already reserved"); // TODO (Gas Optimization): Use custom error
+    if (!_reservedSupply.add(removalId)) {
+      revert RemovalAlreadyReserved({removalId: removalId});
```
jaycenhorton, line 544: https://github.com/nori-dot-eco/contracts/pull/280/files#r913286493

contracts/FIFOMarket.sol
```diff
@@ -547,10 +557,9 @@ contract FIFOMarket is
   function unreserveRemoval(uint256 removalId) external returns (bool) {
     address supplierAddress = removalId.supplierAddress();
 
-    require(
-      _reservedSupply.remove(removalId),
-      "Market: removal not in reserved supply"
-    ); // TODO (Gas Optimization): Use custom error
+    if (!_reservedSupply.remove(removalId)) {
+      revert RemovalNotInReservedSupply({removalId: removalId});
```
jaycenhorton, line 561: https://github.com/nori-dot-eco/contracts/pull/280/files#r913286493

contracts/FIFOMarket.sol
```diff
@@ -559,10 +568,10 @@ contract FIFOMarket is
     if (_activeSupply[supplierAddress].isRemovalQueueEmpty()) {
       _addActiveSupplier(supplierAddress);
     }
-    require(
-      _activeSupply[supplierAddress].insertRemovalByVintage(removalId),
-      "Market: Unable to unreserve removal" // TODO (Gas Optimization): Use custom error
-    ); // returns true if the value was added to the set, that is, if it was not already present
+    // returns true if the value was added to the set, that is, if it was not already present
+    if (!_activeSupply[supplierAddress].insertRemovalByVintage(removalId)) {
+      revert FailedToAddRemovalToActiveSupply({removalId: removalId});
```
jaycenhorton, line 573: https://github.com/nori-dot-eco/contracts/pull/280/files#r913286493

contracts/Removal.sol
```diff
@@ -260,7 +259,7 @@ contract Removal is
     uint256,
     bytes memory
   ) public pure override {
-    revert("Removal: ERC 1155 mint disabled");
+    revert MintDisabled();
```
jaycenhorton, line 262: I would rather just change how this works. We should:
- stop inheriting `ERC1155PresetMinterPauserUpgradeable`
- change the `mint` to allow minting `n` removals for one supplier
- change `mintBatch` to allow minting `n` removals for `m` suppliers
- perhaps this needs to be further extended one more dimension for projects?

contracts/FIFOMarket.sol
jaycenhorton: isn't this actually redundant given [this block of code](https://github.com/nori-dot-eco/contracts/pull/280/files#diff-15c315f8423ca6a9233aa5946874fc34024b7a4c80f9a6cd5c102bfc3cfa35f8R58-R65)

contracts/FIFOMarket.sol
jaycenhorton: isn't this actually redundant given [this block of code](https://github.com/nori-dot-eco/contracts/pull/280/files#diff-15c315f8423ca6a9233aa5946874fc34024b7a4c80f9a6cd5c102bfc3cfa35f8R58-R65)

contracts/FIFOMarket.sol
jaycenhorton: isn't this actually redundant given [this block of code](https://github.com/nori-dot-eco/contracts/pull/280/files#diff-15c315f8423ca6a9233aa5946874fc34024b7a4c80f9a6cd5c102bfc3cfa35f8R58-R65)

contracts/FIFOMarket.sol
jaycenhorton: Why does this use a different error than [this line](https://github.com/nori-dot-eco/contracts/pull/280/files#diff-ddd1c3489433b771aafb62a8b29f1334d571292323137759c94ad4e1545dad2bR531)?

contracts/FIFOMarket.sol
jaycenhorton: Why does this use a different error than [this line](https://github.com/nori-dot-eco/contracts/pull/280/files#diff-ddd1c3489433b771aafb62a8b29f1334d571292323137759c94ad4e1545dad2bR531)?

contracts/Certificate.sol
amiecorso: great, I think we have a ticket now to help prevent the possibility of this further upstream

contracts/FIFOMarket.sol
amiecorso: so i'm digging into the implementation of this a bit and `insertRemovalByVintage` returns the return value of `EnumerableSetUpgradable.add()`, which returns 1 if the value was added to the set (that is, if it was not already present) and 0 if it was already present, in which case great, the id is already there (though not great because this should never happen and would indicate a logical problem).

In terms of this error message, I feel like it's misleading because if it's triggered, it's not because we failed to add the removal to active supply, it's because it was already there to begin with.  I think what I'll do is as you suggested, put the revert into `insertRemovalByVintage` and make the error more obvious.

contracts/FIFOMarket.sol
amiecorso: cleaned all of this up by letting things revert inside `removeRemoval` and in `insertRemovalByVintage` if there's an exceptional case and not up here

contracts/FIFOMarket.sol
amiecorso: I'm going to leave these errors here since `_reservedSupply` is just a straight up enumerable set and we do care about the return value in this case

contracts/FIFOMarket.sol
amiecorso: same here

contracts/Removal.sol
amiecorso: So I stopped inheriting `ERC1155PresetMinterPauserUpgradeable` and added in the infrastructure needed otherwise (like inheriting `ERC1155Burnable`, initializing some roles, calling initializers, etc., got rid of the `mint` overrride.  I haven't made any changes to `mintBatch` yet though.  It does mint `n` removals for one supplier (without validation that they all actually belong to that supplier with encoded address). I'm not sure how we would set things up to mint `n` removals for `m` suppliers so I'm going to leave that to a future PR

contracts/Certificate.sol
amiecorso: do you think we should stop inheriting burning in `ERC1155PresetPausableNonTransferrable` too? (`ERC1155PresetMinterPauserUpgradeable` inherits burning, so when we stop inheriting `ERC1155PresetMinterPauserUpgradeable` we can decide at a granular level what we want to pull in.  maybe just pausable and no burning or minting since we have no need for burning on certificates?)

jaycenhorton: Honestly, might even be worth not inheriting any of the presets since none of them use custom errors. Up to you!

contracts/FIFOMarket.sol
amiecorso: (and `add` is a method in the open zeppelin enumerable set upgradeable utility, which is why we can't push this even deeper)

test/FIFOMarket.test.ts
```diff
@@ -960,6 +906,62 @@ describe('FIFOMarket', () => {
         expect(noriFinalNoriBalance).to.equal(noriInitialNoriBalance);
         expect(await certificate.balanceOf(buyer.address, 0)).to.equal(0);
       });
+      // TODO disallow listing removals with 0 balance in the first place.
```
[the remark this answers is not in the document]
amiecorso: I did yeah... we have a ticket in place to address the fact that removal amounts of 0 should be prevented upstream before a certificate is being minted. such that this test shouldn't even be viable because the place where it lists removals it lists them with an amount of 0, something that we will prevent in the future.  the test broken because I removed the revert statement when minting certificates and instead of fixing it I think we should just fix it when we make the final fix

contracts/FIFOMarket.sol
jaycenhorton: oh, that's the bit i missed!

scottynomad (approved):

PR #286, opened 2022-07-07
[NO-1801] Update `Certificate.sol` sources bookkeeping

amiecorso:
NO-1801


contracts/Certificate.sol
```diff
@@ -25,6 +25,12 @@ contract Certificate is ERC1155PresetPausableNonTransferrable {
     uint256 amount;
   }
 
+  struct CertificateSources {
```
jaycenhorton, line 28: what was the reason for not going with one of these approaches? 

A.
```solidity
 struct Source {
    uint256 sourceId;
    uint256 removalId;
    uint256 amount;
    uint256 replacesSourceId;
  }

  mapping(uint256 => Source[]) private _sources; 
 ```
 ---
B.
```solidity
mapping(uint256 => EnumerableMapUpgradeable.UintToUintMap) private _sources; // certificateId => removalId => amount (maybe could be expanded to assign an id for a source as well? not sure if there's value add

// then rely on burn events and metadata to determine the replacements (alternatively store replaced sources in a similar data structure and then have one parent struct
```

scottynomad: I don't think you can assume replacements will always be 1:1.  You might need multiple new removals to replace a single old one.

scottynomad (approved):

contracts/Certificate.sol
jaycenhorton: Good point. I do think A could be tweaked to address that, and I think it's mostly irrelevant for V assuming we just rely on events

amiecorso: I'm sure there's a way to rely on events for determining replacement removals (not as simple as just things burned or things purchased, since burning and purchases made by Nori's treasury will occur in other contexts besides replacing removals...).  I can't help but feel like for the sake of transparency, the information about what comprises a certificate should be accessible without majorly esoteric knowledge about exactly how to parse and interpret event scenarios

PR #288, opened 2022-07-07
NO-1859 Adds batchCreateGrants with permit signing for the funding transfer of BridgedPolygonNORI

scottynomad:


contracts/test/PermitSigner.sol
```diff
@@ -0,0 +1,26 @@
+// SPDX-License-Identifier: MIT
+pragma solidity =0.8.15;
+
+import "./SigUtils.sol";
+import "@openzeppelin/contracts-upgradeable/token/ERC20/extensions/draft-IERC20PermitUpgradeable.sol";
+
+contract PermitSigner {
+  function digestPermitCall(
```
scottynomad, line 8: I was originally trying to have this call `vm.sign` in this function but the import of foundry std lib confuses the hardhat compilation phase...

test/LockedNORI.t.sol
```diff
@@ -195,22 +215,75 @@ contract LockedNORITest is Test, ERC777ERC1820 {
...
+
+    helper.assertSimplePastGrant(address(lNori), lNori.getGrant(recipient));
+    assertEq(erc20.balanceOf(grantAdmin), 0);
+    assertEq(erc20.balanceOf(address(lNori)), GRANT_AMOUNT);
+
+    assertEq(lNori.balanceOf(recipient), GRANT_AMOUNT);
+    assertEq(lNori.vestedBalanceOf(recipient), 0);
+    assertEq(lNori.unlockedBalanceOf(recipient), 0);
+
+    vm.warp(3656 days);
```
jaycenhorton, line 271: is this a typo or is this just accounting for leap years?

contracts/LockedNORIV2.sol
```diff
@@ -260,6 +260,43 @@ contract LockedNORIV2 is ERC777PresetPausablePermissioned {
     revert("lNori: Transfer to underlying asset failed");
   }
 
+  event DumbLog(string message);
```
jaycenhorton, line 263:
```suggestion
```

jaycenhorton (approved):

test/LockedNORI.t.sol
scottynomad: typo, thanks!

test/LockedNORI.t.sol
```diff
@@ -195,35 +215,90 @@ contract LockedNORITest is Test, ERC777ERC1820 {
       address(this)
     );
 
-    bpNori = new TestToken777();
-    assertEq(bpNori.balanceOf(address(this)), SEED_AMOUNT);
+    erc777 = deployMockERC777();
+    assertEq(erc777.balanceOf(address(this)), SEED_AMOUNT);
```
jaycenhorton, line 219: FYI, something I learned is that assertions don't work like they do in tests ([assertions in setup() and constructors don't fail tests](https://github.com/foundry-rs/foundry/issues/1291)). Workaround is to do `if(erc777.balanceOf(address(this)) !== SEED_AMOUNT){ revert("Balance !== SEED_AMOUNT");}`

scottynomad: Good tip. These are probably unnecessary assertions in the setup anyway.

PR #289, opened 2022-07-08
[NO-1883] Change `RestrictedNORI.batchRevokeUnreleasedTokenAmounts` to `RestrictedNORI.revokeUnreleasedTokens`

amiecorso:
When revoking, we actually don't care about which specific removals we are revoking for, but rather which project they belong to, since all of those insurance tokens for a single project are indistinguishable once in the `RestrictedNORI` contract.  This refactors the revocation endpoint to be singular and at the project level, with the idea being that any revocation transactions being performed will only ever be for one project at a time.  This simplifies the complexity of the function and allows us to avoid making external calls inside a loop.


jaycenhorton (approved):
:100: 

We should honestly consider removing batch functionality from a lot of our contracts because it seems gratuitous to include it when you can usually achieve the same effect by inheriting multicall, OR by using an external contract or api. There are probably some exceptions. (especially true with view functions-- and even more so with functions only meant for truly **external** (e.g. off chain) view functions)

PR #290, opened 2022-07-08
Make certificates ERC-998-like to enable Certificate token IDs to own removal token IDs

jaycenhorton:
Not sure why, but to get this branch to work correctly you will need to set `VIA_IR=false`


scottynomad:
I like it!  Nice work!  I think this is a promising direction we should continue on with.  I was surprised how minimal the changes to the market contract were.  I'd like to find a way to avoid the market owning the certificate token id generation.  Maybe this could be a two call exchange to first mint the certificate and then send removals to it or wrap it up in the payload of the safeTransferTo call to indicate that the intention is to create a new certificate (vs a replacement)

contracts/Market.sol
```diff
@@ -470,6 +468,10 @@ contract FIFOMarket is
       this.totalReservedSupply(); // todo store reserved amount in removal data instead
```
amiecorso, line 468: I think there's a problem with this because `totalReservedSupply()` is giving you the total number of NRTs that are reserved, not the number of individual distinct removal token ids, which is what `numberOfTokensOwnedByAddress` is giving you, so these apples and oranges shouldn't be being subtracted from each other

amiecorso: fixed this in our pair branch from today

PR #293, opened 2022-07-11
Some 1155 cleanup tasks

amiecorso:
[NO-1863] Make sure all external (non-view) functions are pausable
[NO-1870] Validate the removal balances cannot be zero or ever be listed in the market as zero


contracts/Removal.sol
```diff
@@ -270,6 +271,11 @@ contract Removal is
   ) public override {
     // todo perhaps call this listRemovals instead
     // todo do we add any validation to enforce that all removals in batch belong to the same project id?
+    for (uint256 i = 0; i < amounts.length; i++) {
```
jaycenhorton, line 274: Kinda wonder if the better place for this is inside `_beforeTokenTransfer` where we check if amount is 0 as that would also prevent other cases where transfers of zero could occur

PR #295, opened 2022-07-18
[NO-1879] Refactor schedule functions into `RestrictedNORIUtils` library

amiecorso:


contracts/RestrictedNORIUtils.sol
```diff
@@ -0,0 +1,100 @@
...
+    }
+    /* solhint-enable not-rely-on-time */
+    return linearAmountAvailable;
+  }
+
+  /**
+   * @notice Reconstructs a schedule's true total based on claimed and unclaimed tokens.
+   *
+   * @dev claiming burns the 1155, so the true total of a schedule has to be reconstructed
+   * from the totalSupply of the token and any claimed amount.
```
jaycenhorton, line 91:
```suggestion
   * @dev claiming burns the ERC1155 token, so the true total of a schedule has to be reconstructed
   * from the `totalSupply` and any claimed amount.
```

contracts/RestrictedNORIUtils.sol
```diff
@@ -0,0 +1,100 @@
...
+   * or the released amount floor, which is set at the current released amount whenever the balance of a
+   * schedule is decreased through revocation or withdrawal.
+   */
+  function _releasedBalanceOfSingleSchedule(
+    Schedule storage schedule,
+    uint256 totalSupply
+  ) internal view returns (uint256) {
+    return
+      MathUpgradeable.max(
+        _linearReleaseAmountAvailable(schedule, totalSupply),
```
jaycenhorton, line 61: see note about using library within itself [here](https://github.com/nori-dot-eco/contracts/pull/295#r924492073)

contracts/RestrictedNORIUtils.sol
```diff
@@ -0,0 +1,100 @@
...
+ * ```
+ *  using ScheduleUtils for Schedule;
+ *
+ *  mapping(address => Schedule) schedules = Schedules;
+ *  Schedule s = schedules[account];
+ *  s.startTime = 1644436100;
+ *  s.endTime = 1645436100;
+ *  s.totalAmount = 1000000;
+ *  s.addCliff(1644436200, 50000);
+ *  s.amountAvailable(1644436200);
```
jaycenhorton, line 44:
```suggestion
```

test/RestrictedNORI.t.sol
```diff
@@ -12,3 +14,27 @@ contract RestrictedNORI_initialize is UpgradableRestrictedNORIMock {
     // todo rest
   }
 }
+
+contract RestrictedNORI__linearReleaseAmountAvailable is
+  NonUpgradableRestrictedNORIMock
+{
+  using RestrictedNORIUtils for Schedule;
```
jaycenhorton, line 21: I'd be surprised if this is required given that NonupgradeableRestrictedNORIMock inherits from RestrictedNORI

contracts/RestrictedNORI.sol
```diff
@@ -880,9 +836,8 @@ contract RestrictedNORI is
     uint256 balanceOfAccount
   ) private view returns (uint256) {
     Schedule storage schedule = _scheduleIdToScheduleStruct[scheduleId];
-    uint256 scheduleTrueTotal = _scheduleTrueTotal(
-      schedule.totalClaimedAmount,
-      scheduleId
+    uint256 scheduleTrueTotal = schedule._scheduleTrueTotal(
```
jaycenhorton, line 839: I think we should just be passing around `Schedule` as a parameter now so we don't keep re-loading it from the schedule id in the previous line

contracts/RestrictedNORI.sol
```diff
@@ -6,6 +6,7 @@ import "@openzeppelin/contracts-upgradeable/token/ERC1155/extensions/ERC1155Paus
 import "@openzeppelin/contracts-upgradeable/token/ERC1155/extensions/ERC1155SupplyUpgradeable.sol";
 import "./BridgedPolygonNORI.sol";
 import "./Removal.sol";
+import {RestrictedNORIUtils, Schedule} from "./RestrictedNORIUtils.sol";
```
jaycenhorton, line 9: we can remove the MathUpgradeable import on [line 4](https://github.com/nori-dot-eco/contracts/pull/295#diff-de34eb981fc0fa6e6b114737c58140e5e1b03e6f25139d619de635e4e0a9d7e0R4)

contracts/RestrictedNORI.sol
```diff
@@ -400,12 +389,10 @@ contract RestrictedNORI is
...
     }
+    uint256 totalSupply = totalSupply(scheduleId);
     return
-      _scheduleTrueTotal(schedule.totalClaimedAmount, scheduleId) -
-      _releasedBalanceOfSingleSchedule(
-        scheduleId,
-        schedule.releasedAmountFloor
-      );
+      schedule._scheduleTrueTotal(totalSupply) -
+      schedule._releasedBalanceOfSingleSchedule(totalSupply);
```
jaycenhorton, line 395: should probably just add another library function for this

contracts/RestrictedNORI.sol
```diff
@@ -421,10 +408,8 @@ contract RestrictedNORI is
       revert NonexistentSchedule({scheduleId: scheduleId});
     }
     return
-      _releasedBalanceOfSingleSchedule(
-        scheduleId,
-        schedule.releasedAmountFloor
-      ) - schedule.totalClaimedAmount;
+      schedule._releasedBalanceOfSingleSchedule(totalSupply(scheduleId)) -
+      schedule.totalClaimedAmount;
```
jaycenhorton, line 412: should probably just add another library function for this

jaycenhorton (commented):
:100: 

A few comments added inline

contracts/RestrictedNORIUtils.sol
```diff
@@ -0,0 +1,100 @@
...
+    uint256 totalSupply
+  ) internal view returns (uint256) {
+    uint256 linearAmountAvailable = totalSupply;
+    /* solhint-disable not-rely-on-time, this is time-dependent */
+    if (block.timestamp < schedule.endTime) {
+      uint256 rampTotalTime = schedule.endTime - schedule.startTime;
+      linearAmountAvailable = block.timestamp < schedule.startTime
+        ? 0
+        : (_scheduleTrueTotal(schedule, totalSupply) *
+          (block.timestamp - schedule.startTime)) / rampTotalTime;
```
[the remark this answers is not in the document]
amiecorso: oh nice! the first comment here isn't true because the first if is checking for whether the block time is before the schedule end time, but the ternary is asking about whether the block time is before the schedule start time (to avoid any negative math, it should just be 0)

good catch on the library, seems to be working

test/RestrictedNORI.t.sol
```diff
@@ -2,6 +2,8 @@
 // SPDX-License-Identifier: MIT
 pragma solidity =0.8.15;
 import "@/test/helpers/restricted-nori.sol";
+import "@/test/checkout.int.t.sol";
+import {RestrictedNORIUtils, Schedule} from "../contracts/RestrictedNORIUtils.sol";
```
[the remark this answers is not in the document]
amiecorso: yeah looks like I can get rid of it entirely

contracts/RestrictedNORI.sol
amiecorso: good catch

amiecorso: looked through the rest of this contract for any other places where something like this may be happening and I think we're in good shape.  found one small optimization to get rid of one more load

contracts/RestrictedNORI.sol
amiecorso: already part of the lib

contracts/RestrictedNORI.sol
amiecorso: this is just part of the struct though

contracts/RestrictedNORI.sol
amiecorso: I see what your'e saying now haha.  added library functions `_claimableBalanceForSchedule`, `_claimableBalanceForScheduleForAccount`, and `_revocableQuantityForSchedule`

contracts/RestrictedNORIUtils.sol
```diff
@@ -0,0 +1,100 @@
...
+ *  Schedule s = schedules[account];
+ *  s.startTime = 1644436100;
+ *  s.endTime = 1645436100;
+ *  s.totalAmount = 1000000;
+ *  s.addCliff(1644436200, 50000);
+ *  s.amountAvailable(1644436200);
+ * ```
+ *
+ */
+library RestrictedNORIUtils {
```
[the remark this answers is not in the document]
amiecorso: went ahead and renamed `ScheduleUtils` and `RemovalUtils` and all corresponding imports/test harness files etc. to use a Lib suffix.  let me know what you think of the names, definitely open to better ones

test/Removal.t.sol
```diff
@@ -271,6 +271,20 @@ contract Removal_release_listed is UpgradeableMarket {
 
 contract Removal_cummulativeBalanceOfBatch is UpgradeableRemoval {
   function test() external {
+<<<<<<< HEAD
```
jaycenhorton, line 274: merge conflict

jaycenhorton (commented):
Mind running `forge snapshot` and commiting that? We should make it a habit until we can automate it.

.gas-snapshot
```diff
@@ -19,7 +20,7 @@ Removal_release_retired:test() (gas: 92565)
 Removal_release_retired_burned:test() (gas: 92565)
 Removal_release_retired_oneHundredCertificates:test() (gas: 3200343)
 Removal_release_unlisted:test() (gas: 222850)
+RestrictedNORI__linearReleaseAmountAvailable:test() (gas: 11875)
 RestrictedNORI_initialize:test() (gas: 21855)
-ScheduleUtils_availableAmount:test() (gas: 12354)
-Checkout_buyingFromOneRemoval:test() (gas: 826726)
```
jaycenhorton, line 24: why didn't this get run?

jaycenhorton (approved):

PR #300, opened 2022-07-22
[NO-1963] add ability to withdraw instead of reserving removals

jaycenhorton:


utils/deploy.ts
```diff
@@ -125,7 +125,7 @@ export const deployRemovalContract = async ({
   return hre.deployOrUpgradeProxy<Removal, Removal__factory>({
     contractName: 'Removal',
     args: [],
-    options: { initializer: 'initialize()' },
+    options: { initializer: 'initialize()', unsafeAllow: ['delegatecall'] },
```
rlfarman, line 128: Obligatory why do we need to allow something unsafe?

jaycenhorton: https://app.slack.com/client/T7SRSSUN7/C023A5VN86R/1656529393.031139

contracts/Market.sol
```diff
@@ -651,28 +611,30 @@ contract Market is PausableAccessPreset {
...
-    address supplierAddress = RemovalIdLib.supplierAddress(removalId);
-    _removeActiveRemoval(supplierAddress, removalId);
-    if (!_reservedSupply.add(removalId)) {
-      revert RemovalAlreadyReserved({removalId: removalId});
+  function withdraw(uint256 removalId) external whenNotPaused {
+    if (
+      _msgSender() == RemovalIdLib.supplierAddress(removalId) ||
+      hasRole(DEFAULT_ADMIN_ROLE, _msgSender())
+    ) {
+      address supplierAddress = RemovalIdLib.supplierAddress(removalId);
```
amiecorso, line 621: since the supplier address is also used on line 618, maybe do the lookup once at the top of the function?

contracts/Market.sol
```diff
@@ -175,17 +168,6 @@ contract Market is PausableAccessPreset {
...
-   * The amount of supply available for anyone to buy.
-   */
-  function totalUnrestrictedSupply() external view returns (uint256) {
-    uint256 activeSupply = _removal.cumulativeBalanceOf(address(this));
-    return
-      activeSupply < _priorityRestrictedThreshold
-        ? 0
-        : activeSupply - _priorityRestrictedThreshold; // todo compare this against trySub?
-  }
-
```
amiecorso, line 188: i'm pretty sure we still want this function because restricted supply is not the same thing as reserved supply.  this has to do with priority restricted threshold that we hold for allowlisted buyers when supply gets low, totally different concept from the reserved removal pool

contracts/RemovalIdLib.sol
```diff
@@ -39,8 +39,8 @@ uint256 constant _ASCII_CAP_LETTER_MAX_VAL = 90;
 /**
  * @dev Library encapsulating the logic around encoding and decoding removal token ids.
  *
- * The token ids used for a given ERC1155 token in Removal.sol encode information about the carbon removal
- * in the following format(s), where the first byte encodes the format version:
+ * The token IDs used for a given ERC1155 token in Removal encode information about the carbon removal the following
```
amiecorso, line 42:
```suggestion
 * The token IDs used for a given ERC1155 token in Removal encode information about the carbon removal in the following
```

contracts/RemovalIdLib.sol
```diff
@@ -90,6 +90,7 @@ library RemovalIdLib {
   /**
    * @notice Packs data about a removal into a 256-bit token id for the removal.
    * @dev Performs some possible validations on the data before attempting to create the id.
+   * sdf
```
amiecorso, line 93: typo?

scottynomad (approved):

contracts/ArrayLib.sol
```diff
@@ -14,6 +15,33 @@ library ArrayLib {
     }
   }
 
+  function fill(uint256[] memory arr, uint256 val)
```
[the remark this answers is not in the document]
jaycenhorton: how would you accomplish `new uint256[](3).fill(1) // [1,1,1]` without passing it in?

contracts/Market.sol
```diff
@@ -60,20 +58,16 @@ contract Market is PausableAccessPreset {
   uint256 private _noriFeePercentage;
   uint256 private _priorityRestrictedThreshold;
   address private _currentSupplierAddress;
-  mapping(address => RoundRobinOrder) private _suppliersInRoundRobinOrder;
-  mapping(address => RemovalQueueByVintage) private _activeSupply;
-  EnumerableSetUpgradeable.UintSet private _reservedSupply;
+  // todo if we order removals in round robin order when listing, why care about tracking the order of addresses?
+  // todo why can't the encoded vintages be used to order removals instead of storing their set here
+  mapping(address => RoundRobinOrder) private _suppliersInRoundRobinOrder; // todo can this be optimized? redundant?
```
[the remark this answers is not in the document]
jaycenhorton: tracked in #301

contracts/Market.sol
```diff
@@ -38,11 +37,10 @@ contract Market is PausableAccessPreset {
   using RemovalQueue for RemovalQueueByVintage;
 
   error InsufficientSupply();
+  error OnlyAdminOrSupplierCanWithdraw(); // todo consider allowing operators
```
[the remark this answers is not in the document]
jaycenhorton: I like that-- perhaps the reverse though? `UnauthorizedWithdrawal`

contracts/Market.sol
jaycenhorton: it can be retrieved without a helper function

amiecorso: oh good point, by getting the restriction threshold and the active supply we can calculate that oflfine

PR #305, opened 2022-07-27
release across addresses

jaycenhorton:


contracts/Removal.sol
```diff
@@ -383,42 +380,67 @@ contract Removal is ERC1155SupplyUpgradeable, PausableAccessPreset {
...
+    address market = this.marketAddress();
+    address certificate = this.certificateAddress();
+    uint256 unlistedBalance = balanceOf(RemovalIdLib.supplierAddress(id), id);
+    uint256 listedBalance = balanceOf(market, id);
+    uint256 soldBalance = balanceOf(certificate, id);
+    if (unlistedBalance > 0) {
+      amountBurned += MathUpgradeable.min(amount, unlistedBalance);
+      super._burn(RemovalIdLib.supplierAddress(id), id, amountBurned); // todo single call to _burnBatch or emit event
+    }
+    if (amount - amountBurned <= amount) {
```
amiecorso, line 405: this if statement will always be true

amiecorso: I think we want to check if amountBurned < amount?

contracts/Removal.sol
```diff
@@ -383,42 +380,67 @@ contract Removal is ERC1155SupplyUpgradeable, PausableAccessPreset {
...
+      if (listedBalance > 0) {
```
amiecorso, line 406:
```suggestion
      if (amountBurned < amount && listedBalance > 0) {
```

contracts/Removal.sol
```diff
@@ -383,42 +380,67 @@ contract Removal is ERC1155SupplyUpgradeable, PausableAccessPreset {
...
+        uint256 amountToReleaseFromMarket = 0;
+        amountToReleaseFromMarket = MathUpgradeable.min(
+          amount - amountBurned,
+          listedBalance
+        );
+        amountBurned += amountToReleaseFromMarket;
+        super._burn(market, id, amountToReleaseFromMarket);
+        _market.release(id, amountToReleaseFromMarket);
+      }
+      if (amount - amountBurned <= amount && soldBalance > 0) {
```
amiecorso, line 416:
```suggestion
      if (amountBurned < amount && soldBalance > 0) {
```

test/Removal.t.sol
```diff
@@ -259,7 +256,7 @@ contract Removal_release_listed is UpgradeableMarket {
       0
     );
     assertEq(_removal.balanceOf(address(_market), REMOVAL_ID_FIXTURE), 1);
-    _removal.release(address(_market), REMOVAL_ID_FIXTURE, 1);
+    _removal.release(REMOVAL_ID_FIXTURE, 1);
     assertEq(
       _removal.balanceOf(_namedAccounts.supplier, REMOVAL_ID_FIXTURE),
       0
```
amiecorso, line 262: should we add a test case that covers the situation where a removal's balances are spread across the original supplier, the market contract, and the certificate contract, and see that `release` correctly burns from all three owners?

jaycenhorton:
replaced by #306

contracts/Removal.sol
jaycenhorton: this should be addressed [here](https://github.com/nori-dot-eco/contracts/pull/306/files#diff-1f1e10ef0e300f31ed72a37f4be0f5ec9ed7512c3ab3a09020b43f623913e6bfR402) now. Had wrong checks almost entirely in this PR

test/Removal.t.sol
jaycenhorton: def. tracking on #306

PR #306, opened 2022-07-27
release across addresses 

jaycenhorton:
recreating #305 due to weird merge conflicts

In addition to changing the logic of `Removal.release`, this PR bumps the submodules to take advantage of the latest from foundry.

 - [x] address comments in #305 
 - [x] add a test case that covers the situation where a removal's balances are spread across the original supplier, the market contract, and the certificate contract, and see that release correctly burns from all three owners


scottynomad (approved):

contracts/Removal.sol
```diff
@@ -165,28 +165,62 @@ contract Removal is
...
    *
+   * - Releasing burns first from unlisted balances, second from listed balances and third from certificates.
+   * - If the removal is unlisted (e.g., owned by any account other than the market or certificate), the removal is
+   * simply burned.
+   * - If the removal is listed, it is delisted from the market and burned.
+   * - If the removal is owned by one or more certificates, the removal is burned iteratively across each certificate
+   * until the amount is exhausted (e.g., if a removal of amount 3 releases an amount of 2.5 and that removal is owned
+   * by 3 certificates containing an amount of 1 from the released removal, the resulting certificate's removal balances
+   * for this removal are: 0, 0, and 0.5).
+   * - If the removal is included as part of any certificates, the certificate balances are decremented by `amount`.
```
amiecorso, line 181:
```suggestion
```

contracts/Removal.sol
```diff
@@ -165,28 +165,62 @@ contract Removal is
...
-   * @param amount The amount of the removal to release.
-   *
-   * @dev
+   * @notice Releases an amount of a removal.
+   * @dev Releases `amount` of removal by `removalId` by burning them.
    *
    * ##### Requirements:
    *
+   * - Releasing burns first from unlisted balances, second from listed balances and third from certificates.
+   * - If the removal is unlisted (e.g., owned by any account other than the market or certificate), the removal is
```
amiecorso, line 174:
```suggestion
   * - If the removal is unlisted (e.g., owned by the supplier account encoded in the removal id and not the market or certificate contracts), the removal is
```

test/Removal.t.sol
```diff
@@ -272,6 +267,143 @@ contract Removal_release_listed is UpgradeableMarket {
...
+        _expectedOwners,
+        new uint256[](3).fill(_removalIds[0])
+      ),
+      _expectedBalances
+    );
+  }
+}
+
+/** @dev A test that asserts the ability to release a removal retired across 2 certificates */
+contract Removal_release_unlisted_listed_and_retired_2x is UpgradeableMarket {
```
amiecorso, line 339:
```suggestion
contract Removal_release_retired_2x is UpgradeableMarket {
```

amiecorso (approved):
just a couple naming/doc suggestions otherwise this looks great and nice test cases

contracts/Removal.sol
jaycenhorton: Not sure why this is wrong

contracts/Removal.sol
jaycenhorton: I think I prefer the original as it's still correct and future proof

test/Removal.t.sol
jaycenhorton: I prefer the original as it's more descriptive and follows the name of the non-2x version

PR #307, opened 2022-07-27
Fix checkSupply functionality for a specific supplier

rlfarman:
Creates a new function, `_checkSupplyOfSupplier` which will check the total available supply of a supplier (by using `getTotalBalanceFromRemovalQueue` to read from the balance of their associated `RemovalQueue` in `_activeSupply`), and reverting if any of the following conditions are true:

1. The supplier has no active supply.
2. The supplier currently does not have enough supply in `_activeSupply` to cover `purchaseAmount`
3. The market currently does not have supply above the `priorityRestrictedThreshold`, and the buyer is not allowlisted.

Reuses the errors declared for `_checkSupply`, but I can declare a new errors for these scenarios if that makes more sense..


contracts/Market.sol
```diff
@@ -339,6 +338,32 @@ contract Market is PausableAccessPreset {
     }
   }
 
+  /**
+   * @dev Reverts if supplier is out of stock or if total available supply in the market is being reserved for priority buyers
+   * and buyer is not priority.
+   *
+   * @param purchaseAmount The number of carbon removals being purchased.
+   */
+  function _checkSupplyOfSupplier(
```
jaycenhorton, line 347: this isn't ever called

rlfarman: Fixed

jaycenhorton (commented):
can you run `forge snapshot` and commit the gas report file? otherwise lgtm

rlfarman:
No changes from running `forge snapshot`

jaycenhorton (approved):

PR #309, opened 2022-07-27
track certificate deficits and expose view functions to return current/original balances

jaycenhorton:
4 different approaches exist in my mind (these are mostly ordered in terms of my personal preference):

1. **today/previously**-- Rely on events + underlying removals to determine current balance and deficit balances

 _benefit of not tracking any additional state; downside of requiring certificate amounts to be summed using state + events_
 
2. **this PR uses this approach**-- track removals released as a deficit and add a view function that returns the underlying removals + the deficit (e.g., getOriginalAmount)

 _benefit of only tracking additional state in some cases vs all cases_
 
3. pass through the amount generated in the market contract on the certificate contract (e.g., the value returned from `Market.certificateAmountFromPurchaseTotal(amount)` )

 _tracks yet another additional state tracking some kind of balance_
 
4. sum the removals passed through to the certificate and stamp the amount on the certificate contract when it received the batch of them.  

_tracks yet another additional state tracking some kind of balance_

### DIFF
--- a/contracts/Certificate.sol
+++ b/contracts/Certificate.sol
@@ -44,6 +44,8 @@ contract Certificate is
   mapping(uint256 => mapping(uint256 => uint256))
     private _removalBalancesOfCertificate;
 
+  mapping(uint256 => uint256) private _deficits; // todo naming consistency for mappings (e.g, plural/non-plural)
+
   /*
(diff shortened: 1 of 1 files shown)


scottynomad (changes requested):
I don't think this approach works unless we also add logic do knock the deficit out again when new removals are sent to the certificate to replace the bad ones.

scottynomad (approved):

PR #311, opened 2022-07-28
move rNori schedule creation responsibility to the removal contract

jaycenhorton:


scottynomad (approved):

contracts/Removal.sol
```diff
@@ -153,6 +153,7 @@ contract Removal is
       methodologyVersion: RemovalIdLib.methodologyVersion(firstRemoval)
     });
     _mintBatch(to, ids, amounts, "");
+    RestrictedNORI(_market.restrictedNoriAddress()).createSchedule(projectId);
```
amiecorso, line 156: do we want to do this when supply is minted? or when it's listed?  schedule creation was originally here in the removal contract and was happening in `safeBatchTransferFrom` (as opposed to `mintBatch`) before we moved it to `onERC1155BatchReceived`

jaycenhorton: Unless we decide to decouple it entirely from both minting and lsiting, I think schedules should always be create when minting, not just when listing. This addresses a todo item in the todo PR that mentioned we needed to validate that all removals being listed had the same project ID (previously we were using `removals[0]` otherwise you end up in a place where you could list after minting using removals from multiple projects which would be an issue AFAICT. 

I think I might have suggested this get decoupled from the minting of removals initially solely to save gas. But I don't think I intended to offload it to listing, I think I had originally meant to decouple it from both (e.g., create schedule then mint).That's probably over-engineering though and I think this might be the best approach considering larger projects with hundreds of removals might have to mint calling this func multiple times anyways (I think that's probably unavoidable for now). 

Do you think there's a strong reason to keep it in the market contract?

amiecorso: nope, I thought it made more sense to have in the removal contract originally.  was just thinking about whether it made more sense to have here while minting v.s. while listing.  I think I understand what your'e talking about regarding the project id.... since the project id literally gets set as the same value here for all removals being minted, creating the schedules at this time guarantees that everything lines up correctly between the regsitered project id for a removal, and the corresponding (identical) schedule id in restricted nori.  i'm convinced!

amiecorso (approved):

PR #313, opened 2022-07-28
use interfaces for core contracts

jaycenhorton:
Temporarily setting this aside as it's turning out to be high-effort low-impact. 

- [all about interfaces](https://medium.com/coinmonks/solidity-tutorial-all-about-interfaces-f547d2869499)
- the interface was initially generated using `cast interface artifacts/contracts/Removal.sol/Removal.json -o contracts/interfaces/IRemoval.sol -n IRemoval` Which defines the interface really well, but duplicates inherited interfaces. The commented out sections in IRemoval are all inherited and thus are duplicates. The only thing I manually added was the errors and the interface inheritance and the names of the struct keys which weren't auto detected.

Opening this PR now though as the interfaces generated provide a really good view into what functions are public even if they are inherited which allows us to more easily guarantee we've appropriately constrained access to these functions as well as added pausability.


contracts/interfaces/ILockedNORIV2
```diff
@@ -0,0 +1,73 @@
+pragma solidity ^0.8.10;
+
+interface ILockedNORIV2 {
```
scottynomad, line 3: Looks like a bug that this in missing an file extension.

contracts/Removal.sol
```diff
@@ -361,7 +340,7 @@ contract Removal is
     uint256[] memory ids,
     uint256[] memory amounts,
     bytes memory data
-  ) public override {
+  ) public override(IERC1155Upgradeable, ERC1155Upgradeable) {
```
scottynomad, line 343: why are both the interface and the implementation contract in here?

contracts/Removal.sol
```diff
@@ -385,7 +364,11 @@ contract Removal is
   function supportsInterface(bytes4 interfaceId)
     public
     view
-    override(ERC1155Upgradeable, AccessControlEnumerableUpgradeable)
+    override(
+      IERC165Upgradeable,
```
scottynomad, line 368: Is change related?

scottynomad (commented):
What problem are we solving here?  These are interesting for their documentation value but don't seem to change the code significantly.

What would be the ongoing dev workflow here?  Are these treated like pure codegen that gets rewritten as part of the build process or do people need to update them manually?

contracts/Removal.sol
jaycenhorton: Was required

contracts/interfaces/ILockedNORIV2
jaycenhorton: Yeah quickly generated these. So probably just a bug on my cli use

jaycenhorton:
> What problem are we solving here? These are interesting for their documentation value but don't seem to change the code significantly.
> 
> What would be the ongoing dev workflow here? Are these treated like pure codegen that gets rewritten as part of the build process or do people need to update them manually?

As I mentioned I identified this as low impact, high effort, so this is likely to not be completed.


I haven't seen a **great** reason to prefer interfaces over importing the contracts themselves yet, but there are lots of points linked in the article in the original comment that are noteworthy 

**The primary reason I generated these was for reference, anything beyond that is less interesting**


However, I was also interested in using interfaces as a means to find:
- some kind of consistency for error, event and struct declarations that didn't require importing the contract code into files that shared these definitions. I think the net result is a decreased contract size which our solc runs constantly warn that we are over limit for. 
- the only way to use the inheritdoc natspec tag is via interfaces
- defining events in interfaces seems like the only clean way to re-use event definitions in forge without having to redefine them

Afaik there's no codegen process for declaring interfaces. They are usual

PR #315, opened 2022-07-29
Maintain running total of `Market`'s balance of removal balances

amiecorso:
Creates a state variable to keep a running total of the `Market` contract's balance across all removal tokens it owns.  This way, we can determine whether we have enough supply to fulfill an order (and easily calculate how much non-priority supply we have) in a gas efficient way.


test/Removal.t.sol
```diff
@@ -456,12 +504,15 @@ contract Removal__beforeTokenTransfer is NonUpgradableRemoval {
     );
   }
 
-  function test_zeroValueTransfer_reverts_RemovalAmountZero() external {
+  function test_zeroValueTransferToMarket_reverts_RemovalAmountZero() external {
+    address mockMarketAddress = vm.addr(1);
+    // todo figure out stdstorage
+    // stdstore.target(address(super)).sig(super._market.selector).checked_write(mockMarketAddress);
```
amiecorso, line 510: this is currently broken because i changed the implementation such that transfers with 0 amounts are only reverted if they are to or from the market contract (a decision we could revisit, it just seemed more efficient bc it shouldn't otherwise matter).  so I'm trying to get this test to work by giving the test contract a notion of what the market contract's address is, which is not straightforward but *should* be possible without over-complicating this test with a full upgradeablemarket ecosystem

amiecorso:
we might consider getting rid of `removal.cumulativeBalanceOf` entirely, but it's currently being used in some test expectations

scottynomad:
Looks clean and reasonable to me!

test/Removal.t.sol
amiecorso: resolved

jaycenhorton (commented):
did you mean to update the forge-std submodule?

test/Removal.t.sol
```diff
@@ -456,12 +504,18 @@ contract Removal__beforeTokenTransfer is NonUpgradableRemoval {
...
   }
 
-  function test_zeroValueTransfer_reverts_RemovalAmountZero() external {
+  function test_zeroValueTransferToMarket_reverts_RemovalAmountZero() external {
+    address mockMarketAddress = vm.addr(1);
+    vm.store(
+      address(this),
+      bytes32(uint256(401)),
+      bytes32(uint256(uint160(address(mockMarketAddress))))
+    );
```
jaycenhorton, line 513: can you move everything except the `super.beforeTokenTransfer` to a setup function? gives more realistic gas estimates when running `forge snapshot` as it'll exclude these things from the gas calculation (also make sure to commit the resulting `.gas-snapshot`

amiecorso: moot since this function no longer depends on having a specific market address to revert on 0 amount (per our discussion to restore that behavior)

test/Removal.t.sol
```diff
@@ -429,6 +429,54 @@ contract Removal_multicall is UpgradeableMarket {
   }
 }
 
+contract Removal_market_balance is UpgradeableMarket {
```
jaycenhorton, line 432:
```suggestion
contract Removal_getMarketBalance is UpgradeableMarket {
```

test/Removal.t.sol
```diff
@@ -429,6 +429,54 @@ contract Removal_multicall is UpgradeableMarket {
...
+  uint256[] private _removalIds;
+
+  function test_market_balance() external {
```
jaycenhorton, line 435:
```suggestion
  function test() external {
```

jaycenhorton (approved):
two minor name change requests, but otherwise lgtm

test/Removal.t.sol
jaycenhorton: naming convention thus far is 
`ContractName_MethodName(_variant)`

test names are 
`test // base case`
`test(_variant)`

.gas-snapshot
```diff
@@ -12,34 +12,35 @@ LockedNORILib_availableAmount:test() (gas: 12340)
...
+Market_withdraw_as_supplier:test() (gas: 264038)
 Market_withdraw_reverts:test() (gas: 135794)
-Removal__beforeTokenTransfer:test() (gas: 5311)
+Removal__beforeTokenTransfer:test() (gas: 7593)
 Removal__beforeTokenTransfer:test_paused_reverts_Paused() (gas: 29426)
-Removal__beforeTokenTransfer:test_zeroValueTransfer_reverts_RemovalAmountZero() (gas: 8938)
-Removal_mintBatch:test() (gas: 429101)
-Removal_multicall:test_balanceOfBatch() (gas: 598147)
+Removal__beforeTokenTransfer:test_zeroValueTransferToMarket_reverts_RemovalAmountZero() (gas: 8960)
+Removal_market_balance:test_market_balance() (gas: 1281480)
```
jaycenhorton, line 26: also FYI-- this test provides a high value because it does all the test setup within the `test_market_balance` fn instead of `setUp()`

jaycenhorton (approved):

jaycenhorton (approved):

.gas-snapshot
amiecorso: yeah, I could maybe move one bit of setup outside of this test but it basically relies on assertions happening in between making changes like listing, selling, withdrawing, so I can't really extract everything cleanly

PR #321, opened 2022-08-01
refactor withdraw function to enable more granular tests + readability

jaycenhorton:


contracts/Market.sol
```diff
@@ -663,6 +656,12 @@ contract Market is PausableAccessPreset {
     }
   }
 
+  function _isAuthorizedWithdrawl(address owner) internal view returns (bool) {
```
amiecorso, line 659: "Withdrawal" is actually the correct spelling -- replace?

amiecorso (approved):

PR #322, opened 2022-08-01
rm unused state from removal contract

jaycenhorton:


contracts/Removal.sol
```diff
@@ -138,23 +127,17 @@ contract Removal is
...
-      uint256 id = ids[i];
-      if (exists(id) || _removalIdToRemovalData[id].projectId != 0) {
-        revert TokenIdExists({tokenId: id});
-      }
-      _removalIdToRemovalData[id].projectId = projectId; // todo access _removalIdToRemovalData[removalId] once per loop
-      _removalIdToRemovalData[id].holdbackPercentage = holdbackPercentage;
-    }
+    _createRemovalDataBatch({ids: ids, projectId: projectId});
+    _removalIdToProjectId[projectId] = holdbackPercentage;
+    _removalIdToProjectId[projectId] = data.holdbackPercentage;
```
amiecorso, line 132: something's wrong with these assignments I believe, because they're both assigning to `_removalIdToProjectId` AND they're both assigning not the project id but the holdback percentage

amiecorso: also this mapping gets assigned to in `_createRemovalData`

jaycenhorton: projectId assignment is handled in `_createRemovalData`

This gets further cleaned up in #323 which removes the data param altogether. Definitely didn't mean to assign this value twice though. fixing.

jaycenhorton: fixed. Gas further reduced as a result as well.

jaycenhorton: This now shaves ~75k gas per removal minted. I believe even more  is shaved in #323 but will see once I sync everything

amiecorso (approved):

PR #320, opened 2022-08-01
enable non-batch receipts of removals

jaycenhorton:
This is a pre-requisite for a cleanup of the Removal 1155 inheritance branch I am working on

PR #323, opened 2022-08-01
clean up removal contract

jaycenhorton:
**This PR includes changes from the following PRs, so see those first so that I can merge them into this PR**
- [x] #316
- [x] #317
- [x] #318
- [x] #319
- [x] #320
- [x] #321
- [x] #322


contracts/Market.sol
```diff
@@ -367,7 +367,8 @@ contract Market is PausableAccessPreset {
   }
 
   /**
-   * @dev Reverts if available stock is being reserved for priority buyers and buyer is not priority.
+   * @dev Reverts if supplier is out of stock or if total available supply in the market is being reserved for priority
+   * buyers and buyer is not priority.
```
amiecorso, line 371: I think we actually want the original description here because this function only validates priority threshold

contracts/Removal.sol
```diff
@@ -351,6 +288,29 @@ contract Removal is
...
   }
 
+  function isApprovedForAll(address account, address operator)
+    public
+    view
+    override
+    returns (bool)
+  {
+    return
+      account == address(_market) ||
```
amiecorso, line 298: confused about this because this will always return true if asking whether someone is an approved operator for the market, which doesn't seem right at all, especially because in `setApprovalForAll` the market is explicitly not allowed to become an operator for anything.  maybe I'm missing something

contracts/Removal.sol
```diff
@@ -393,15 +353,21 @@ contract Removal is
   }
 
   function _createRemovalDataBatch(
-    uint256[] calldata removalIds,
+    UnpackedRemovalIdV0[] calldata removals,
```
amiecorso, line 356: potentially concerned about this approach because unless we're passing the ids to `mintBatch`, we don't necessarily know what those ids are going to be, and we can't reliably stamp them onto the removal entities that we need to be keeping track of off-chain.  in writing the first version of the admin minting page, there wasn't a reliable way to depend on events to unequivocally link a minted ID to the removal for which it was being minted.  of course we can generate a removal's id using the `RemovalIdLib`, but unless we then also explicitly pass that id to the mint function, there's still a disconnect that could lead to something being corrupted

amiecorso (commented):
I'm gonna approve this and you can consider my comment about passing the removal id v.s. removal data structs to `mintBatch`.... really worried that this might make our lives harder in the administrative part of all this but I will leave the rest of that decision to you as I see there's a lot of work in here already converting all the tests and stuff

amiecorso (approved):

contracts/Removal.sol
jaycenhorton: the intent of this is to hijack the internals of `safe(Batch)TransferFrom` such that we don't ever need to set an approval for the market and that the market is an approved sender for all accounts always (such that it enables calling the release and withdraw functions). Approvals don't currently have a use case otherwise, but they aren't precluded in case we want to leverage it later. The only reason setting the market as approved/unapproved reverts is because setting the approval there wouldn't do anything given the way this check works.

Previously we were setting approvals on every mint, but this was adding to the overhead of minting as you either had to set state or load state on every mint. This circumvents this and puts it further down the stack as it is always only checked if the sender isn't the owner  e.g., `from == _msgSender() || isApprovedForAll(from, _msgSender()),`

Does that make sense? 

I suppose another way to do this would be to approve the market in the initializer and then continuing to disallow changing that approval in the approval function... Maybe that would be preferable?

jaycenhorton: I think I'd include that in a follow up PR though unless this is creating unexpected behavior that I'm otherwise missing

jaycenhorton: here's how this is used internally

```sol
 function safeTransferFrom(
        address from,
        address to,
        uint256 id,
        uint256 amount,
        bytes memory data
    ) public virtual override {
        require(
            from == _msgSender() || isApprovedForAll(from, _msgSender()),
            "ERC1155: caller is not token owner nor approved"
        );
        _safeTransferFrom(from, to, id, amount, data);
    }
```

AFAICT unless you're the market (_msgSender + from) this should be safe-- but maybe I'm overlooking something obvious

jaycenhorton:
```
ontract Market_withdraw_reverts is MarketBalanceTestHelper {
  function setUp() external {
    _removalIds = _seedRemovals({
      to: _namedAccounts.supplier,
      count: 1,
      list: true
    });
    _suppliers = new address[](1).fill(_namedAccounts.supplier);
    _expectedRemovalBalances = [0];
    _expectedMarketSupply = _amountPerRemoval * _removalIds.length;
    _expectedTokenCount.set(_namedAccounts.supplier, 0);
    _expectedTokenCount.set(address(_market), _removalIds.length);
    _assertCorrectStates();
  }

  function test_correctlyReverts() external {
    vm.prank(_namedAccounts.supplier2);
    vm.expectRevert(UnauthorizedWithdrawal.selector);
    _market.withdraw(_removalIds[0]);
    _assertCorrectStates();
  }

  function test_incorrectlyAllow() external {
    vm.prank(_namedAccounts.supplier2);
    vm.expectRevert(UnauthorizedWithdrawal.selector);
    _removal.safeTransferFrom({
      from: address(_market),
      to: _namedAccounts.supplier2,
      id: _removalIds[0],
      amount: 1 ether,
      data: ""
    });
    _assertCorrectStates();
  }
}
```
**`[FAIL. Reason: Call did not revert as expected] test_incorrectlyAllow`**
you're right, this allows be to withdraw by `safeTransferFrom` on the removal. good catch. I guess the initializer is the best approach here

amiecorso: innnnnnteresting... I actually didn't catch this issue, was mostly just confused, but I see now what's happening.  I wonder if that's because we really just wanted to be asking (inside `isApprovedForAll`) to `return operator == address(_market) || ....` instead of `return account == address(_market) || ....` as it's written.  we want to know if it's the market _calling_ `safe(Batch)TransferFrom`, not whether it's the market from which the transfer is being pulled

amiecorso: I think a piece that's still not making sense is actually why the market needs to be an approved operator at all... I know that we were setting nori admin address (whoever deploys i think) to be an operator so that we can list supplier removals for sale in the market, but once that happens the market literally owns the removals anyway, and so is free to transfer them to the certificate.  and `release` is initiated by the removal contract (who also performs the burning of the market's balance because the message sender for that call is a nori admin address), and `withdraw` has its own set of permissions around it and again just involves the market transferring its own balance of a removal to someone else.  so I still can't understand why the market contract needs permanent operator approval for all removal tokens

contracts/Removal.sol
jaycenhorton: there emitted in the `TransferBatch` event, no?

jaycenhorton: I honestly hated encoding them off chain as it makes any validation on-chain basically unreliable. Strongly in favor of keeping it this way, but totally onboard to add whatever needed to do what you're asking for assuming the event that's emitted isn't enough

jaycenhorton: (you get zero-cost intrinsic validation by using a struct as the param as you can't pass invalid types, btw)

contracts/Removal.sol
jaycenhorton: hmm, maybe it doesn't. will check

jaycenhorton: we were approving it as an operator previously in the mint function which is where I think this requirement came from

jaycenhorton: all tests pass without it :shrug: any idea why we had it to begin with? This isn't something new I added, just something I shifted the responsibility for

contracts/Removal.sol
amiecorso: if I recall, the issue is that yes they are emitted in the TransferBatch event but if you have multiple successive mintings happening to the same supplier (and I think even just in a single mint with multiple removals) you just don't have a reliable way to create a perfect 1:1 linkage between the given removal in our database and the exact on chain id it got.  i think in the case that there's only one minting event for the supplier you can rely on the order of the argument in the array to map 1:1 your batch and their ids (though even this feels a little brittle) but the bigger deal is whether you have multiple mints for a supplier

amiecorso: it's hard to think about what additional event information could be emitted that would solve for this problem unless some other uniquely identifying off-chain id was connected to the entity and then also passed in as part of `mintBatch` and used as a sort of identifying tag to match up with the official removal id.  like the issue here isn't one of implementation but rather of information actually being unique or not.  I have a feeling that if we go with this approach we are going to end up generating the ids off-chain too with a call to `createRemovalId` and then just hoping that everything was done correctly.

jaycenhorton: We could just return the ids from the mint function. Either way we have to track some data here if we cant just get it on chain. We could also just expose a function that let us paginate through existing IDs. because they definitely do exist on-chain

jaycenhorton: if we need to store a mapping in a db somewhere, seems reasonable to just use the return value as the thing we store

amiecorso: transactions don't have return values though... I think that's the issue and why we have to rely on events

amiecorso: exposing a function that allows us to paginate through existing ids is something I've thought we might want to have anyway.... but I don't think it helps us with this problem. the ID is the unique identifier of a removal.  unless we say off-chain that THIS removal has THIS ID, and then use that exact piece of data to mint it, there's just no other uniquely identifying way that we can keep the two things completely in sync.  I'm having a lot of opinions about this because this is literally a problem that we have grappled with before and landed in a place where passing the ids made the most sense.  I'm worried about the headache it's going to cause us when we try to finish the app portion of this

amiecorso: so i'm up for doing it differently but I just feel like I want to have a clear plan that we know will work before making this commitment and I can't put it all together yet

PR #324, opened 2022-08-01
Fix priority threshold check bug

amiecorso:
Fixes a bug where we were just checking whether there was ANY non-priority active supply available for a non-priority purchase.  But what we really need to check is whether the active supply minus the purchase amount STILL keeps us above that priority threshold.  Otherwise we risk going ahead and fulfilling a large purchase when there may have only been a tiny bit of active non-priority supply actually available.


jaycenhorton (changes requested):
can you checkin the updated forge snapshot?

jaycenhorton (approved):

PR #325, opened 2022-08-01
Add events for changing Removal state

rlfarman:


jaycenhorton (changes requested):
can you check in the updated forge snapshot?

scottynomad (changes requested):
@rlfarman I feel like this is suffering some of the same issues I pointed out with the market events yesterday -- while we want to emit events on state changes, the granularity at which we do so should map more onto the external interface of the contract rather that onto the individual implementation details of the internal data structures.

i.e. I would expect only one event per removal minted in the `mintBatch` function.  One to say that a schedule was created.

My instinct would also be to omit the balance updated one since that's state we're rolling up internally for efficiency sake only -- however I know Jaycen is concerned about that getting out of sync and this is the only way we'd know if it did so probably smart to keep it in.

.gas-snapshot
```diff
@@ -15,7 +15,7 @@ UInt256ArrayLib_sum:test_library() (gas: 16743)
 UInt256ArrayLib_sum:test_reference() (gas: 33815)
 BridgedPolygonNORI_name:test() (gas: 17211)
 BridgedPolygonNORI_permit:test() (gas: 91892)
-Certificate__msgSenderERC721A:test() (gas: 354)
+Certificate**msgSenderERC721A:test() (gas: 354)
```
jaycenhorton, line 18: fyi i think something got messed up in this file-- can you regenerate it before merging?

contracts/Removal.sol
```diff
@@ -46,6 +46,18 @@ contract Removal is
...
 
+  /**
+   * @notice Emitted on updating `_removalIdToRemovalData` with a new removal.
+   * @param removalId The id of the removal that was released.
+   * @param releasedFrom The address the removal was released from.
+   * @param amount The amount that was released.
+   */
+  event RemovalReleased(
+    uint256 indexed removalId,
+    address indexed releasedFrom,
```
jaycenhorton, line 57: naming feels redundant given the name of the event, perhaps change to `from`, `account` `owner` `supplier` or something like that

contracts/Removal.sol
```diff
@@ -46,6 +46,18 @@ contract Removal is
...
   uint256 private _currentMarketBalance;
 
+  /**
+   * @notice Emitted on updating `_removalIdToRemovalData` with a new removal.
+   * @param removalId The id of the removal that was released.
+   * @param releasedFrom The address the removal was released from.
+   * @param amount The amount that was released.
+   */
+  event RemovalReleased(
+    uint256 indexed removalId,
```
jaycenhorton, line 56: naming feels redundant given the name of the event, perhaps change to `id` `removal` or something similar

contracts/Removal.sol
```diff
@@ -365,6 +380,11 @@ contract Removal is
         removalId,
         amountToReleaseFromCertificate
       );
+      emit RemovalReleased(
+        removalId,
+        this.certificateAddress(),
```
jaycenhorton, line 385: using `address(_certificate)` _may_ make this cheaper

contracts/Removal.sol
```diff
@@ -339,6 +353,7 @@ contract Removal is
   function _releaseFromMarket(uint256 removalId, uint256 amount) internal {
     super._burn(this.marketAddress(), removalId, amount);
     _market.release(removalId, amount);
+    emit RemovalReleased(removalId, this.marketAddress(), amount);
```
jaycenhorton, line 356: using `address(_market)` _may_ make this cheaper

contracts/Removal.sol
```diff
@@ -46,6 +46,18 @@ contract Removal is
     private _addressToOwnedTokenIds;
   uint256 private _currentMarketBalance;
 
+  /**
+   * @notice Emitted on updating `_removalIdToRemovalData` with a new removal.
```
amiecorso, line 50: fix comment ^

amiecorso (approved):
just a copy paste typo in the event docs, otherwise looks good. realizing we may rewrite `mintBatch` to customize our minting event behavior and that that would come in a separate PR

PR #326, opened 2022-08-01
Add events for changing Certificate state

rlfarman:


contracts/Certificate.sol
```diff
@@ -58,6 +58,51 @@ contract Certificate is
...
+   * @param certificatedId The certificate to connected to the removal.
+   * @param removalId The removal to update the balance for.
+   * @param isAdded Whether the amount was added or removed.
+   * @param amount The amount to added or removed for the certificate.
+   */
+  event RemovalBalancesOfCertificateUpdated(
+    uint256 indexed certificatedId,
+    uint256 indexed removalId,
+    bool indexed isAdded,
+    uint256 amount
```
rlfarman, line 72: For something like this, would it be prefereable to use two different events (IE - RemovalBalancesOfCertificateAdded and RemovalBalancesOfCertificateSubtracted), and then we remove the `isAdded` param and index `amount`? Or is it better to have a single event that can return all updates to this state (but without the ability to index on a particular amount)?

jaycenhorton: A single event is preferable where possible as events are expensive

Would be best to stay consistent with event tense (e.g., OZ uses present tense `Transfer`). I think this event should be called `Release`

The flipside of this case is already covered (in `_receiveRemovalBatch` is already covered by ERC721.Mint)

jaycenhorton: not sure I see the value in making these general purpose with the `isAdded` arg

scottynomad (changes requested):
As on Removals the goal here should be to emit events to the outside world that are concerned with the overall state of the contract and not about the internals of the contract.  Along the lines of:
* Certificate created
* Removal release from certificate

contracts/ICertificate.sol
```diff
@@ -9,4 +9,16 @@ interface ICertificate {
     uint256[] removalIds,
     uint256[] removalAmounts
   );
+
+  /**
+   * @notice Emitted when a removal is released from a Certificate.
+   * @param certificatedId The certificate to connected to the removal.
+   * @param removalId The removal to update the balance for.
+   * @param amount The amount to added or removed for the certificate.
```
amiecorso, line 17:
```suggestion
   * @param amount The amount removed from the certificate.
```

amiecorso: releasing is only ever a decrementing of balance.  in the future we have to implement the replacement mechanism which will add new removal tokens to the certificate

amiecorso (approved):
nice, just my note on the docs

PR #327, opened 2022-08-01
Add events for changing Market state

rlfarman:
TODO on this one: All of the state changes that happen inside RemovalQueue. Probably will open a PR for those separately.


contracts/Market.sol
```diff
@@ -57,9 +57,72 @@ contract Market is PausableAccessPreset {
...
+   * @notice Emitted on setting of `_noriFeePercentage`.
+   * @param updatedFeePercentage The updated fee percentage for Nori.
+   */
+  event NoriFeePercentageUpdated(uint256 updatedFeePercentage);
+
+  /**
+   * @notice Emitted on setting of `_currentSupplierAddress`.
+   * @param updatedSupplierAddress The updated supplier address.
+   */
+  event CurrentSupplierAddressChanged(address updatedSupplierAddress);
```
amiecorso, line 80: I feel like this piece of state change actually should not involve emitting an event because this is likely to change many times in the course of fulfilling a single purchase and will add to the gas overhead for the buyer while not being particularly useful to us

contracts/Market.sol
```diff
@@ -57,9 +57,72 @@ contract Market is PausableAccessPreset {
...
+
+  /**
+   * @notice Emitted when a removal is added to `_activeSupply`.
+   * @param removalId The removal that was added.
+   * @param supplierAddress The address of the supplier for the removal.
+   */
+  event RemovalAdded(
+    uint256 indexed removalId,
+    address indexed supplierAddress
+  );
```
amiecorso, line 114: the changes I'm making in [this PR ](https://github.com/nori-dot-eco/contracts/pull/337) change it so that removals can be added to a `RemovalQueue` without reverting even if they already existed in the queue.  I wonder if we should either push this event emission into the `RemovalQueue` library and predicate it on the return value from `enumerableset.add(removalId)`, or if we should have `insertRemovalByVintage` return true or false based on whether the set had already contained the removal, and then predicate the emission of this event on that return value.  Just so that we're only emitting this event when a removal was truly added to the queue

contracts/Market.sol
```diff
@@ -691,14 +751,31 @@ contract Market is PausableAccessPreset {
...
+  {
     _activeSupply[supplierAddress].removeRemoval(removalId);
     if (_activeSupply[supplierAddress].isRemovalQueueEmpty()) {
       _removeActiveSupplier(supplierAddress); // todo can this be combined inside .removeRemoval?
     }
+    emit RemovalRemoved(removalId, supplierAddress);
+  }
+
+  /**
+   * @notice Adds the specified removal id from the active supply data structure.
```
amiecorso, line 765:
```suggestion
   * @notice Adds the specified removal id to the active supply data structure.
```

contracts/Market.sol
```diff
@@ -746,6 +822,9 @@ contract Market is PausableAccessPreset {
     _currentSupplierAddress = _suppliersInRoundRobinOrder[
       _currentSupplierAddress
     ].nextSupplierAddress;
+    emit CurrentSupplierAddressChanged(
```
amiecorso, line 825: rm

contracts/Market.sol
```diff
@@ -756,37 +835,47 @@ contract Market is PausableAccessPreset {
...
-  function _addActiveSupplier(address supplierAddress) private {
+  function _addActiveSupplier(address newSupplierAddress) private {
     // If this is the first supplier to be added, update the intialized addresses.
     if (_currentSupplierAddress == address(0)) {
-      _currentSupplierAddress = supplierAddress;
-      _suppliersInRoundRobinOrder[supplierAddress] = RoundRobinOrder({
-        previousSupplierAddress: supplierAddress,
-        nextSupplierAddress: supplierAddress
+      _currentSupplierAddress = newSupplierAddress;
+      emit CurrentSupplierAddressChanged(newSupplierAddress);
```
amiecorso, line 842: rm

contracts/Market.sol
```diff
@@ -800,29 +889,32 @@ contract Market is PausableAccessPreset {
...
     /**
      * If this is the last supplier, clear all current tracked addresses.
      */
-    if (
-      addressToRemove ==
-      _suppliersInRoundRobinOrder[addressToRemove].nextSupplierAddress
-    ) {
+    if (addressToRemove == nextOfRemovedSupplierAddress) {
       _currentSupplierAddress = address(0);
+      emit CurrentSupplierAddressChanged(address(0));
```
amiecorso, line 904: rm

.gas-snapshot
```diff
@@ -98,5 +98,5 @@ RestrictedNORI__validateSchedule_reverts:test_startTimeZero() (gas: 3350)
 RestrictedNORI_initialize:test() (gas: 21825)
 RestrictedNORI_scheduleExists:test() (gas: 15027)
 RestrictedNORI_scheduleExists:test_doesntExist() (gas: 15006)
-Checkout_buyingFromOneRemoval:test() (gas: 810319)
-Checkout_buyingFromTenRemovals:test() (gas: 3082458)
+Checkout_buyingFromOneRemoval:test() (gas: 817084)
+Checkout_buyingFromTenRemovals:test() (gas: 3102580)
```
jaycenhorton, line 102: 20,000 gas more expensive :skull:

contracts/Market.sol
```diff
@@ -57,9 +57,66 @@ contract Market is PausableAccessPreset {
...
+    address indexed previousSupplierAddress
+  );
+
+  /**
+   * @notice Emitted when a removal is added to `_activeSupply`.
+   * @param removalId The removal that was added.
+   * @param supplierAddress The address of the supplier for the removal.
+   */
+  event RemovalAdded(
+    uint256 indexed removalId,
```
jaycenhorton, line 106: seems redundant to use `removalId` in an event called `RemovalAdded` , perhaps `id` would be better?

contracts/Market.sol
```diff
@@ -676,9 +730,9 @@ contract Market is PausableAccessPreset {
       _removal.isApprovedForAll({account: owner, operator: _msgSender()}));
   }
 
-  function _listForSale(uint256 id) internal {
-    address supplierAddress = RemovalIdLib.supplierAddress(id);
-    _activeSupply[supplierAddress].insertRemovalByVintage(id);
+  function _addActiveRemoval(uint256 removalId) internal {
```
jaycenhorton, line 733: `removalId` seems redundant given the function name :shrug:

jaycenhorton: applies to other fns as well

.gas-snapshot
amiecorso: I think a standard ETH transfer is 21,000 gas, so this is more expensive by only about 1 standard transaction

contracts/Market.sol
```diff
@@ -57,9 +57,66 @@ contract Market is PausableAccessPreset {
...
+
+  /**
+   * @notice Emitted when a Removal is removed from `_activeSupply`.
+   * @param removalId The removal that was added.
+   * @param supplierAddress The address of the supplier for the removal.
+   */
+  event RemovalRemoved(
+    uint256 indexed removalId,
+    address indexed supplierAddress
+  );
```
amiecorso, line 118: I'm actually thinking that this may not be an event we want... mostly because this is going to be emitted as purchases are made and so is probably what's responsible for the significant gas increase of buying from multiple removals while not adding a lot of value. I think if we're ever in a position of having to debug our market, we won't be relying on these events to do that.  And if we ever want to see which removals specifically have sold through we can use the data about the certificates being created for that

amiecorso (approved):

PR #331, opened 2022-08-03
cleanup and improve the array library

jaycenhorton:
image


contracts/RemovalQueue.sol
```diff
@@ -188,4 +190,23 @@ library RemovalQueue {
...
+    internal
+    pure
+    returns (address[] memory)
+  {
+    // uint256 n = from.length;
+    // address[] memory to = new address[](from.length);
+    assembly {
+      // Create an dynamic sized array manually.
+      let memOffset := mload(0x40) // 0x40 is the address where next free memory slot is stored in Solidity.
+      mstore(memOffset, 0x20) // single dimensional array, data offset is 0x20
```
amiecorso, line 204: wondering if there's any reason to have consistency between hex and decimal for this 32 value

amiecorso (approved):
this is so awesome

contracts/RemovalQueue.sol
jaycenhorton: oh, thought I caught that already and changed everything to decimal-- good catch

PR #332, opened 2022-08-03
rm balanceOfIds function and add tests for RemovalQueue lib that depended on it

jaycenhorton:


amiecorso (approved):
very nice! I don't think the `RemovalQueueLib` actually depended on `balanceOfIds` (but it seems like nothing depended on it that I can see in this PR at least so good cut), and great new tests

jaycenhorton:
> very nice! I don't think the `RemovalQueueLib` actually depended on `balanceOfIds` (but it seems like nothing depended on it that I can see in this PR at least so good cut), and great new tests
It does on [master](https://github.com/nori-dot-eco/contracts/blob/master/contracts/RemovalQueue.sol#L181), just a weird diff since this is branched off the `jaycen-arraylib-cleanup`

PR #335, opened 2022-08-03
short circuit certificate child token queries when empty

jaycenhorton:


amiecorso (approved):
interesting.... does this actually provide gas savings?  I would imagine that if the number of removals is 0 the loop simply wouldn't run at all the result would be the same?

PR #336, opened 2022-08-03
Use multicall to get rid of `batchGetHoldbackPercentages` and use single variant

amiecorso:


contracts/Market.sol
```diff
@@ -417,7 +417,6 @@ contract Market is PausableAccessPreset {
     for (uint256 i = 0; i < numberOfActiveRemovalsInMarket; i++) {
       uint256 removalId = _activeSupply[_currentSupplierAddress]
         .getNextRemovalForSale();
-      // todo retrieve balances in a single batch call
```
amiecorso, line 420: I don't believe we can do this in a batch call because that batch call would have to get the balances for the entire supply of the market, which is super wasteful.  seems like we should keep doing it this way and only get the balances for removals as-needed because we might only need to move through 1 or a few of them to fulfill the order

rlfarman (approved):

contracts/Market.sol
jaycenhorton: think you're right-- unfortunate

contracts/Market.sol
```diff
@@ -738,6 +735,34 @@ contract Market is PausableAccessPreset {
     return super.supportsInterface(interfaceId);
   }
 
+  /**
```
jaycenhorton, line 738: IMO this should be a fn of the removal contract

contracts/Market.sol
```diff
@@ -738,6 +735,34 @@ contract Market is PausableAccessPreset {
...
+   * @dev Gets the holdback percentages for a batch of removal ids using multicall.
+   * @param ids The removal token IDs for which to retrieve the holdback percentages
+   */
+  function _getHoldbackPercentages(uint256[] memory ids)
+    internal
+    returns (uint8[] memory)
+  {
+    uint256 numberOfRemovals = ids.length;
+    bytes[] memory getHoldbackPercentageCalls = new bytes[](numberOfRemovals);
+    for (uint256 i = 0; i < numberOfRemovals; i++) {
```
jaycenhorton, line 748: use ++i and wrap the for loop in unchecked
```
 // Skip overflow check as for loop is indexed starting at zero.
    unchecked {
    ...
    }
```

contracts/Market.sol
```diff
@@ -738,6 +735,34 @@ contract Market is PausableAccessPreset {
...
+      getHoldbackPercentageCalls[i] = abi.encodeWithSelector(
+        _removal.getHoldbackPercentage.selector,
+        ids[i]
+      );
+    }
+    bytes[] memory holdbackPercentageResults = _removal.multicall(
+      getHoldbackPercentageCalls
+    );
+    uint8[] memory holdbackPercentages = new uint8[](numberOfRemovals);
+    for (uint256 i = 0; i < numberOfRemovals; i++) {
```
jaycenhorton, line 758: use ++i and wrap the for loop in unchecked
```
 // Skip overflow check as for loop is indexed starting at zero.
    unchecked {
    ```
    ...
    }

contracts/Market.sol
jaycenhorton: nvm I see what you're doing and approve of the pattern

jaycenhorton (approved):
approved; but def. add those unchecked and ++i changes

PR #337, opened 2022-08-04
Allow listing removal in portions

amiecorso:
Addressing the second-to-last TODO in the "Possible Features" section of https://github.com/nori-dot-eco/contracts/pull/301

 "Consider enabling listing more of a removal that was partially listed in the queue and then subsequently wants to list the rest after the queue state contains other removals from other suppliers/vintages/etc? Document and write tests for current behavior if we decide not to do this. Discussion here https://norinauts.slack.com/archives/C03QHCALNPQ/p1659019570004879"

There is no reason to revert when attempting to add a removal to a `RemovalQueueByVintage`.  These are sets anyway, and so cannot end up with duplicate token id data.  By making this change, we are able to `consign` a portion of a removal, and then call `consign` for any still unlisted balance of that removal and have things work as expected using the existing infrastructure.


contracts/RemovalQueue.sol
```diff
@@ -40,12 +40,7 @@ library RemovalQueue {
     } else if (vintageOfRemoval > removalQueue.latestYear) {
       removalQueue.latestYear = vintageOfRemoval;
     }
-    if (!removalQueue.queueByVintage[vintageOfRemoval].add(removalToInsert)) {
-      revert RemovalAlreadyInQueue({
```
rlfarman, line 44: We can remove this custom error as well, I suppose>

rlfarman (approved):

jaycenhorton (approved):

PR #340, opened 2022-08-04
Disallow receiving ERC1155 from any but Removal contract in Market

amiecorso:
Long story short: ended up writing elaborate tests that spoof a different 1155 contract that sends to the market in order to test the revert.  Turns out I could have just used the existing simple tests and omitted the (now required) `vm.prank(address(_removal))`.

Maybe we simplify the tests, maybe we leave in the infrastructure for this fake extra 1155....


contracts/Market.sol
```diff
@@ -182,7 +185,7 @@ contract Market is PausableAccessPreset {
     uint256[] memory,
     bytes memory
   ) external returns (bytes4) {
-    // todo revert if Market.onERC1155BatchReceived sender is not the removal contract
+    require(_msgSender() == address(_removal), "Sender not Removal contract");
```
amiecorso, line 188: we use revert strings here instead of custom errors because the inherited code that catches errors in this `onERC1155(Batch)Received` hook can't match custom errors, and we end up getting the misleading error "Transfer to Non-ERC1155 Receiver".

rlfarman (approved):

jaycenhorton (approved):

PR #341, opened 2022-08-04
Prevents any address from owning removal tokens except supplier, Market, Certificate or zero address

amiecorso:
One of our security TODOs.  It's essential that no address besides the encoded supplier, market, certificate, or 0 address owns a removal token (otherwise `release` won't work as expected because it may not accurately know how much unlisted balance is held because it could be help by an arbitrary address).  By putting this in `_beforeTokenTransfer`, we also get validation upon minting that we don't mint the wrong token ids to the wrong address.


jaycenhorton (approved):
significant gas increase :disappointed:

PR #344, opened 2022-08-05
Functions permissioned/pausable/disabled as necessary

amiecorso:
Helpful syntax for generating interfaces to examine:
`cast interface artifacts/contracts/RestrictedNORI.sol/RestrictedNORI.json -o contracts/interfaces/IRestrictedNORI.sol -n IRestrictedNORI`

- [x] Removal.sol
- [x] Certificate.sol (burn, safeTransferFrom, safeTransferFrom, transferFrom)
- [x] Market.sol
- [x] ~~RestrictedNORI.sol~~ _no changes required_
- [x] ~~BridgedPolygonNORI.sol~~ _no changes required_
- [x] ~~NORI.sol~~ _no changes required_


contracts/Removal.sol
```diff
@@ -331,6 +332,38 @@ contract Removal is
...
+   * @dev See {IERC1155-safeTransferFrom}.
+   */
+  function safeTransferFrom(
+    address from,
+    address to,
+    uint256 id,
+    uint256 amount,
+    bytes memory data
+  ) public override whenNotPaused {
+    if (_msgSender() != address(_market)) {
```
amiecorso, line 345: the only reason we have to keep these transfer functions around (`safeTransferFrom` and `safeBatchTransferFrom`) is because the Market contract uses them while completing purchases (to send to Certificate) and in the `withdraw` implementation.  We already don't allow anyone to own a removal besides the original supplier, the market, the certificate, or the 0 address. By adding this check that only the market can use this function, we prevent a supplier from using this function directly to transfer their tokens to either the market or certificate contracts (or even 0 address)

contracts/Removal.sol
```diff
@@ -403,32 +436,6 @@ contract Removal is
...
   }
 
-  function isApprovedForAll(address account, address operator)
-    public
-    view
-    override
-    returns (bool)
-  {
-    return
-      account == address(_market) ||
```
amiecorso, line 413: I am pretty certain that this stuff was accidentally included in my documentation PR when it had actually been removed from Jaycen's Removal contract cleanup PR.  we explicitly talked about not needing the market to have operator approval for any reason, and actually including this always-true check for the market being approved introduced a security vulnerability where someone random could call `withdraw` on the market contract and have the call succeed.

test/Removal.t.sol
```diff
@@ -837,3 +861,42 @@ contract Removal_safeBatchTransferFrom_reverts_ForbiddenTransfer is
     });
   }
 }
+
```
amiecorso, line 864: for these just wanted to experiment to see that these functions are pausable by way of the pausable overrides of their internal implementations inside `PausableAccessPreset`.  As reading in the solidity docs, I believe that (for example and WLOG) since `grantRole` calls `_grantRole` (and not `super._grantRole`), it will resolve `_grantRole` starting with the most derived contract possible and walking up the inheritance tree from there, so indeed it calls the version defined in `PausableAccessPreset` and not the version defined in its own class

contracts/Market.sol
```diff
@@ -598,21 +606,19 @@ contract Market is PausableAccessPreset {
...
-    uint256[] calldata ids,
-    uint256[] calldata amounts,
+    uint256[] memory ids,
+    uint256[] memory amounts,
     address[] memory suppliers
-  ) external {
-    // todo verify changes to `fulfillOrder` (memory->calldata arr args) that enabled [:index] arr slicing syntax is ok
-    uint256[] memory batchedIds = ids[:numberOfRemovals];
-    uint256[] memory batchedAmounts = amounts[:numberOfRemovals];
+  ) internal {
```
amiecorso, line 619: based on the answer to [this post](https://ethereum.stackexchange.com/questions/123169/can-calldata-be-used-in-every-function-visibility), I believe that `calldata` arguments really do have to come from the data that was submitted with the initial transaction externally, because it lives in a different place in memory.  I've had unexpected behavior when marking params with `calldata` that are actually called internally with non-calldata arguments.  Jaycen was trying to obtain gas savings on this function using calldata but it's not appropriate here

scottynomad (approved):
LGTM!

PR #345, opened 2022-08-08
Upgrade Solidity to `0.8.16`

amiecorso:
not yet working locally.... get "Unknown version specified" when trying to run `forge test` locally.  `hardhat compile` seems to work


jaycenhorton:
latest version of hardhat supports 0.8.16-- maybe that'll help?

PR #347, opened 2022-08-08
Unchained initializers all called, all contracts inherit `MulticallUpgradeable`

amiecorso:
Went through all contracts to ensure all unchained initializers in their inheritance tree are called, and took this opportunity to make sure all contracts inherit from `MulticallUpgradeable` because that also affects initializers.

Edit: found that I cannot make `LockedNORIV2.sol` `MulticallUpgradeable` because it changes the storage layout and is therefore not upgrade safe


contracts/Market.sol
```diff
@@ -2,6 +2,8 @@
 pragma solidity =0.8.15;
 import "./Certificate.sol";
 import "./RestrictedNORI.sol";
+import "./PausableAccessPreset.sol";
```
amiecorso, line 5: adding imports where I find them missing because we discovered that's why the IDE has trouble with cmd-click: it only works if the relevant file is explicitly imported

contracts/Removal.sol
```diff
@@ -143,6 +144,7 @@ contract Removal is
 
   function initialize() external initializer {
     __Context_init_unchained();
+    // __ERC165_init_unchained();
```
scottynomad, line 147: delete?

scottynomad (approved):
Are you concerned about the slither warning about unprotected initializer?  (See bottom of Market.sol and Removal.sol)

contracts/Removal.sol
amiecorso: uncomment

amiecorso:
> Are you concerned about the slither warning about unprotected initializer? (See bottom of Market.sol and Removal.sol)

yes, but addressing in the slither PR / have a note out to leonard / would like to potentially write a test for this to see if we really can destroy the contract with that call

PR #348, opened 2022-08-08
Slither fixes and triage

amiecorso:
Running Slither on each file of concern with:
`slither contracts/FileName.sol --filter-paths "node_modules/" --triage-mode` (filtering out node_modules prevents analyzing our dependencies)
Triage mode docs: https://github.com/crytic/slither/wiki/Usage#triage-mode

Things that I am ignoring (because it's not in a readable format in `slither.db.json`)
- unchecked return from `transfer` and `transferFrom`. the vulnerability here is that some implementations of `transfer(From)` may not revert on failure, meaning our contract might perform accounting and allow the transaction to complete successfully when it shouldn't.  However, we are only interacting with our own implementation of `transfer(From)` in `BPNori`, and this function *does* revert on failure so we can save gas by not checking this return value
- unchecked return values from set operations: the return values from set operations (`add/remove`) give us information about whether items were already in the set, not whether the operation succeeded or failed.  In all of these cases, we don't care whether the item was already in the set. 
- unchecked return values from `multicall` on `Certificate.releaseRemoval` - this function doesn't return anything. verified with a test that if the function reverts during execution, the whole transaction reverts (which is good)
- variable name casing not being mixed case (we are ok with using leading underscores to indicate privacy)
- using timestamp for comparisons (necessary in `RestrictedNORI` which is time-dependent)
- using assembly
- costly operations inside a loop for adding/removing suppliers (this should be a rare operation that is actually performed very infrequently because it only happens when a new supplier is added to the market or when a supplier completely sells out)
- Yelling at us to not use solidity 0.8.15 (we don't need to downgrade as we are even considering upgrading to 0.8.16)
- variable names being too similar to each other (ok if variable and its setter are similar)
- unused library functions: Slither can't detect the fact that libraries are supposed to be used by other contracts and so if an internal function declared in a library isn't actually used internally in the library itself, slither flags it as unused but that's actually ok and expected
- reentrancy in `Removal.release(...)` - the external calls are only to contracts that we have control over and do not make additional external calls

Note: The only issue I'm seeing that is concerning and not yet fixed in this PR is: `
```0: NORI (contracts/NORI.sol#7-29) is an upgradeable contract that does not protect its initiliaze functions: NORI.initialize() (contracts/NORI.sol#15-28). Anyone can delete the contract with: MulticallUpgradeable.multicall(bytes[]) (node_modules/@openzeppelin/contracts-upgradeable/utils/MulticallUpgradeable.sol#23-29)Reference: https://github.com/crytic/slither/wiki/Detector-Documentation#unprotected-upgradeable-contract```

Which I have asked Len to add to our list of questions for the auditors, noting " I don’t understand the issue / there’s not much I can find online about it besides the fact that false positives for vulnerable upgradeable contracts are a known slither issue in several cases…  Can we add this to the list of questions for the auditors?  It’s kinda related to something already in that list about the safety of Multicall with upgradeable contracts, but I want to make sure this specific situation is accounted for"

Update:  added tests to try to produce a vulnerability along these lines and things revert as expected.  this is almost certainly a false positive from slither but we will still ask the auditors


scottynomad (approved):
Nice cleanups!  Thanks for going through and running triage mode, that's going to make this process much less painful in the future!

PR #349, opened 2022-08-09
Add getters and setters for registering contract events

rlfarman:
Adds getters and setters and update some docs for updating contract addresses across `Removal`, `Certificate`, and `Market`.

I haven't found any other places where getters or setters seem necessary.


amiecorso (approved):

amiecorso:
there's going to be a challenge here in overriding the `_mintBatch` function.... which is that is references the private variable `_balances`, which is not accessible outside of the inherited contract :/

PR #352, opened 2022-08-10
Use safemath for mul and div

rlfarman:
Increases gas tho???


amiecorso:
according to this https://forum.openzeppelin.com/t/oz-contracts-v4-safemath/9759/4, SafeMath isn't explicitly needed anymore... I recall Jaycen telling me this too, so I'm not totally sure why it was a todo

PR #355, opened 2022-08-10
Declare variables outside of loop

rlfarman:
This may not save gas, will be easier to tell once #353 is merged


contracts/RemovalQueue.sol
```diff
@@ -164,14 +164,15 @@ library RemovalQueue {
...
+    uint256 i = 0;
     for (uint256 year = removalQueue.earliestYear; year <= latestYear; ++year) {
       EnumerableSetUpgradeable.UintSet storage vintagesForYear = removalQueue
         .queueByVintage[year];
       uint256[] memory removalIds = new uint256[](vintagesForYear.length());
       uint256 numberOfRemovals = removalIds.length;
       // Skip overflow check as for loop is indexed starting at zero.
       unchecked {
-        for (uint256 i = 0; i < numberOfRemovals; ++i) {
+        for (i = 0; i < numberOfRemovals; ++i) {
```
amiecorso, line 175: interesting, yeah I wonder if this will save any gas... because I think what we're really trying to get at here is making sure that we're not declaring a variable inside the loop body such that is unnecessarily re-declared on every iteration when we could declare it once outside

PR #360, opened 2022-08-11
Array improvement check

amiecorso:
Checked all of our market contract and library files for places where the efficient `ArrayLib` functions could be used to make an improvement, only found one place.  Added some tests to illuminate the small gas savings there and because we didn't have such tests.


contracts/RemovalQueue.sol
```diff
@@ -176,7 +176,6 @@ library RemovalQueue {
         }
       }
       totalBalance += removal
-      // todo extracting this outside of the loop into a single call would drastically improve performance
```
amiecorso, line 179: it wouldn't

rlfarman (approved):

PR #363, opened 2022-08-11
Conventions!

amiecorso:
The result of our triplet programming time to create more consistency and better naming conventions


amiecorso:
i think the renames and stuff have removed some of the slither triaging that was quieting these warnings so might have to do another pass on that

scottynomad (approved):
So conventional!  🙌

PR #367, opened 2022-08-12
More doc updates for `Removal.sol`

amiecorso:
and some function reordering


contracts/Removal.sol
```diff
@@ -499,6 +446,58 @@ contract Removal is
...
+    }
+    super.safeTransferFrom(from, to, id, amount, data);
+  }
+
+  /**
+   * @dev Batched version of {safeTransferFrom}.
+   *
+   * Emits a {TransferBatch} event.
+   *
+   * Requirements:
```
rlfarman, line 481: You'd wanna throw `#####` before this to keep the formatting the same as elsewhere

contracts/Removal.sol
```diff
@@ -655,6 +643,72 @@ contract Removal is
...
+      if (amounts[i] == 0) {
+        revert InvalidTokenTransfer({tokenId: id});
+      }
+      if (to == market) {
+        _currentMarketBalance += amounts[i];
+      }
+      if (from == market) {
+        _currentMarketBalance -= amounts[i];
+      }
+      if (
```
rlfarman, line 680: This undoes our optimization from earlier

contracts/Removal.sol
```diff
@@ -499,6 +446,58 @@ contract Removal is
     return id.decodeRemovalIdV0();
   }
 
+  /**
+   * @dev Transfers `amount` tokens of token type `id` from `from` to `to`.
+   *
+   * Emits a {TransferSingle} event.
```
rlfarman, line 452: From the todo list:

> Remove natspec comments that use syntax including {...} (e.g., See {[https://docs.openzeppelin.com/contracts/4.x/api/token/erc20#ERC20Permit}](https://docs.openzeppelin.com/contracts/4.x/api/token/erc20#ERC20Permit%7D)) as this syntax is only valid for OpenZeppelins custom internal docgen

contracts/Removal.sol
```diff
@@ -499,6 +446,58 @@ contract Removal is
...
+   *
+   * Requirements:
+   *
+   * - Can only be called by the `Market` contract.
+   * - `to` cannot be the zero address.
+   * - If the caller is not `from`, it must have been approved to spend ``from``'s tokens via {setApprovalForAll}.
```
rlfarman, line 458: Same comment about `{...]`

contracts/Removal.sol
```diff
@@ -499,6 +446,58 @@ contract Removal is
...
+   * - `from` must have a balance of tokens of type `id` of at least `amount`.
+   * - If `to` refers to a smart contract, it must implement {IERC1155Receiver-onERC1155Received} and return the
```
rlfarman, line 460: `{...}`

contracts/Removal.sol
```diff
@@ -499,6 +446,58 @@ contract Removal is
     return id.decodeRemovalIdV0();
   }
 
+  /**
+   * @dev Transfers `amount` tokens of token type `id` from `from` to `to`.
```
rlfarman, line 450: Generally, if we're keeping documentation to a minimum, I'd prefer we rely on `@notice` over `@dev`. [From the natspec docs](https://docs.soliditylang.org/en/latest/natspec-format.html):

> @notice - Explain to an end user what this does
> @dev - Explain to a developer any extra details

I think this is our main goal for documentation at the moment is to help the auditors understand what all these functions do at a higher level, rather than explain more subtle implementation details. So @notice is going to be more appropriate, since its purpose it to help a reader understand what a function does as a baseline, and then @dev is where you put any details to help the reader understand how it works.

contracts/Removal.sol
```diff
@@ -688,6 +750,12 @@ contract Removal is
     }
   }
 
+  /**
+   * @dev Validates that the provided `id` should be minted.
```
rlfarman, line 754: I feel like we might be overly relying on the `@dev` tags here.
Generally, if we're keeping documentation to a minimum, I'd prefer we rely on `@notice` over `@dev`. [From the natspec docs](https://docs.soliditylang.org/en/latest/natspec-format.html):

> @notice - Explain to an end user what this does
> @dev - Explain to a developer any extra details

So `@notice` is meant to help a reader understand what a function does as a baseline, and then @dev is where you put any extra implementation details to help a developer understand how it works. I think generally it's preferable to have some kind of `@notice` to explain what something does at a high level to a general audience, and then optionally have `@dev` flags to explain any subtle implementation details that a developer should know.

I don't know what Jaycen meant by the todo " @dev vs @notice consistency", but I think the idea behind being consistent with `@dev` and `@notice`.

rlfarman (changes requested):
We should at least fix the merge conflict with the optimization - otherwise just some documentation/formatting notes.

contracts/Removal.sol
scottynomad: that's my bad, I started doing that in LockedNORI not realizing that the curly braces were OZ private sauce.

contracts/Removal.sol
scottynomad: I think it would be worth calling out that this is implementation of one `safeTransferFrom` in the ERC1155 interface.

contracts/Removal.sol
```diff
@@ -613,17 +562,56 @@ contract Removal is
...
 
+  /**
+   * @dev Burns `amount` of token ID `id` from the supplier address encoded in the ID.
+   *
+   * Emits a {RemovalReleased} event.
+   *
+   * @param id The token ID to burn.
+   * @param amount The amount to burn.
+   */
+  function _releaseFromSupplier(uint256 id, uint256 amount) internal {
```
scottynomad, line 585: Unrelated to this PR but I wonder if calling these internal helpers `_burnFrom...` rather than `_releaseFrom...` would be clearer.

scottynomad (commented):
Looks good so far!  Could I request brief dev docs on the three main mappings?

contracts/Removal.sol
rlfarman: Sure, thanks, in that case a `@notice` definitely isn't super necessary, I just chose this function at random.

contracts/Removal.sol
amiecorso: oh jeez good catch, wonder how that happened

contracts/Removal.sol
amiecorso: thought about that too, but for consistency with the `Certificate` situation (which involves more than just burning) I'm going to leave all of these consistent.  They also emit events called `RemovalReleased`

scottynomad (approved):

PR #371, opened 2022-08-15
fix deployments and deploy to mumbai + defender as v2

jaycenhorton:


contracts.json
```diff
@@ -93,19 +93,28 @@
...
-      "proxyAddress": "0x61819E8238A3493547Ae1131dc657D87F62833FD"
+      "proxyAddress": "0x53D76F0d411F8Fc5278999Ae7929C163Dbeac7d7"
     },
     "LockedNORI": {
       "proxyAddress": "0xCdcB43CB7B668F0C1cA04FE4b60DA7F8C62Be393"
+    },
+    "RestrictedNORI": {
+      "proxyAddress": "0x0924196d683daBe242216Eb3e27bcf8FccCeCaD1"
+    },
+    "LockedNORIV2": {
```
scottynomad, line 113: Hmm this is interesting. At this level I think it would have made more sense for this to replace `LockedNORI`.  Also it looks like this deploy replaced the `LockedNORI` proxy rather than upgrading it which means we've lost all the state.

plugins/index.ts
```diff
@@ -78,6 +79,7 @@ const deployOrUpgradeProxy = async <
     }
   }
   const [signer] = await hre.getSigners();
+  console.log("-----",await hre.getSigners())
```
scottynomad, line 82: leftovers

contracts.json
amiecorso: if it replaced the proxy, doesn't that mean the old proxy/implementation is still out there and we could still upgrade it?

amiecorso (approved):
nice!

scottynomad (approved):

PR #372, opened 2022-08-16
Scripting and autotask stuff 

amiecorso:
This PR includes:

- Autotask code just because it's too hard to edit directly in defender (will remove from here eventually or move it into its own place such that it can be managed by Richie's defender API work to automate this part)

- Scripts for automatically minting BPNori and minting/listing removals for sale.

Note that for these scripts to work, the environment variables will have to be set like:
```
MUMBAI_RPC_URL='https://polygon-mumbai.infura.io/v3/YOUR_INFURA_KEY'
PRIVATE_KEY=0XDON'T COMMIT ME
```

where the `PRIVATE_KEY` is the private key generated from the staging mnemonic (or any valid, funded address you can sign with - just needs to be a simple private key signer because i'm not sure how to get fireblocks working with foundry), which has been granted any necessary admin/depositor/consignor roles in the relevant contracts via hardhat command line task transactions that used the fireblocks (deployer) credentials as signer. (one time thing, per deployment at least)

Here are the hardhat commands for granting roles to whatever address you're going to admin with (in this case the 0x465... staging mnemonic address):
`hardhat --network mumbai Removal  --func grantRole 0xa269776b75ac4c5fa422bb11bec3ed3cee626848d07687372583174b209261fb 0x465d5a3fFeA4CD109043499Fa576c3E16f918463`

(or like this for making a BPNORI deposit from command line:)
`hardhat --network mumbai BridgedPolygonNORI --func deposit 0x6dc772f80495f47d8000530a59ee975b67b7c646 "0x0000000000000000000000000000000000000000000000056bc75e2d63100000"` <-- encoded version of the equivalent of 100 NORI

And here's how I generated the private key to add to my `.env` from the mnemonic:
```
let mnemonic = "dont commit your mnemonic";
let mnemonicWallet = hre.ethers.Wallet.fromMnemonic(mnemonic);
console.log(mnemonicWallet.privateKey);
```

then, the script can be run like:
`forge script script/MintAndListRemovals.s.sol:MintAndListRemovals --rpc-url $MUMBAI_RPC_URL --private-key $PRIVATE_KEY --broadcast -vvvv`

or (for the minting)
`forge script script/MintBPNori.s.sol:MintBPNori.sol --rpc-url $MUMBAI_RPC_URL --private-key $PRIVATE_KEY --broadcast -vvvv`
 
NOTE: txns for minting removals seem really slow on mumbai... like taking almost 15 minutes to get included. These script commands can take a `--gas-price` flag and it might be worth bumping up to whatever "fast" is on polygonscan? (haven't tried this yet but need to remember for next time)

see `forge script --help` for many more command line options and cool stuff.
also `cast` provides lots of cool command line utilities for learning things about txns, accounts, etc, basically just needs an RPC_URL


.gitignore
```diff
@@ -105,6 +105,7 @@ scripts/tenderly.ts
 # foundry
 out
 cache_hardhat
+broadcast
```
amiecorso, line 108: this is where forge script command puts info about the txns it's creating and broadcasting, so want to gitignore

contracts.json
```diff
@@ -93,16 +93,16 @@
...
     "Removal": {
-      "proxyAddress": "0xa484c59eB391afFd88C7de0dBF4B6E7e2A9E84fE"
+      "proxyAddress": "0xa051E9EeaC803d2fCA1DbF415b78AD9BfEB723b0"
     },
     "Certificate": {
       "proxyAddress": "0xF5bafc8B26Fb38ea788Bc6eAf3d84ba3bc4Ca91C"
     },
     "Market": {
-      "proxyAddress": "0x61819E8238A3493547Ae1131dc657D87F62833FD"
+      "proxyAddress": "0x53D76F0d411F8Fc5278999Ae7929C163Dbeac7d7"
```
amiecorso, line 105: for some reason the addresses here do not match the addresses of these contracts as registered in Defender... so something may still be not right about our deployment process because we want to keep these in sync.

The changes in this PR are manually changing these addresses to point to the defender addresses so that our hardhat task commands would work on the correct contracts.... but this is still a mystery we need to solve

scottynomad (commented):
Nice!  Love the solidity scripts!  On the `MUMBAI_RPC_URL` front I'd like to encourage people to use Alchemy rather than Infura so we get better logging / debugging.  You should have access to log in and get the RPC URL for mumbai.

Is it worth trying to get the autotask script to load the ABI from elsewhere in the repo or is this going to go away too soon to make that worth it?

scripts/autotask.js
```diff
@@ -0,0 +1,1901 @@
+const {
+  DefenderRelaySigner,
+  DefenderRelayProvider,
+} = require('defender-relay-client/lib/ethers');
+const ethers = require('ethers');
+const { hexZeroPad } = require('@ethersproject/bytes');
+const { parseUnits, formatUnits } = require('@ethersproject/units');
+const { KeyValueStoreClient } = require('defender-kvstore-client');
+
+const MARKET_ABI = [
```
jaycenhorton, line 10: fwiw you only need the part of the abi which you use, so this could be reduced to just the following (assuming you can't just import the abi instead):

```ts
const MARKET_ABI = [
{
    inputs: [
      {
        internalType: 'address',
        name: 'recipient',
        type: 'address',
      },
      {
        internalType: 'uint256',
        name: 'amount',
        type: 'uint256',
      },
      {
        internalType: 'uint256',
        name: 'deadline',
        type: 'uint256',
      },
      {
        internalType: 'uint8',
        name: 'v',
        type: 'uint8',
      },
      {
        internalType: 'bytes32',
        name: 'r',
        type: 'bytes32',
      },
      {
        internalType: 'bytes32',
        name: 's',
        type: 'bytes32',
      },
    ],
    name: 'swap',
    outputs: [],
    stateMutability: 'nonpayable',
    type: 'function',
  }
]
```

scripts/autotask.js
```diff
@@ -0,0 +1,1901 @@
...
+  const { v, r, s } = ethers.utils.splitSignature(signature);
+  console.log({ owner, name, nonce, chainId, signature, v, r, s });
+  const { hash: transactionHash } = await marketContract.swap(
+    buyerWalletAddress || RELAYER_WALLET,
+    parseUnits(amount.toString(), 18).toString(),
+    deadline,
+    v,
+    r,
+    s,
+    { gasLimit: '2000000' }
```
jaycenhorton, line 1897: was this needed?

scripts/autotask.js
```diff
@@ -0,0 +1,1901 @@
...
+    MARKET_ABI,
+    signer
+  );
+  const owner = await signer.getAddress();
+  const chainId = 80001;
+  const name = await bpNoriContract.name();
+  const nonce = await bpNoriContract.nonces(owner);
+  const latestBlockTime = (await provider.getBlock('latest')).timestamp;
+  // const deadline = latestBlockTime + 3600; // 1 hour into future
+  const deadline = ethers.constants.MaxUint256;
```
jaycenhorton, line 1876: why was this needed?

script/MintAndListRemovals.s.sol
```diff
@@ -0,0 +1,47 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity =0.8.15;
+
+import "forge-std/Script.sol";
+import "../contracts/Removal.sol";
+import {UInt256ArrayLib} from "../contracts/ArrayLib.sol";
+
+contract MintAndListRemovals is Script {
```
jaycenhorton, line 8: why can't we just use [`cast send ...`](https://book.getfoundry.sh/reference/cast/cast-send)  for stuff like this?

script/MintBPNori.s.sol
```diff
@@ -0,0 +1,25 @@
...
+  function run() external {
+    vm.startBroadcast();
+    _bridgedPolygonNORI = BridgedPolygonNORI(bpNoriAddress);
+    _bridgedPolygonNORI.deposit(
+      defenderRelayer,
+      abi.encode(200000000000000000000)
+    ); // 200 NORI
+    vm.stopBroadcast();
+  }
+}
```
jaycenhorton, line 25: this is the same as the following, right?

```bash
cast send \
  -c mumbai \
  --mnemonic-path $MNEMONIC_PATH \
  0xBD16B6880F76CA8A3543dd7A2a60d375d0485dBB \
  "deposit(address,bytes)" \
  0x6Dc772F80495F47d8000530A59eE975B67B7c646 \
  `cast --to-uint256 200000000000000000000` \
  /
```

scripts/autotask.js
amiecorso: at some point our hope is to manage this autotask code with the defender client and thus incorporate it as part of our repo.... not sure if we'll ever be able to import the ABIs though.  the code in this repo is actually just for editing it in a decent text editor, I probably shouldn't actually commit this at all since we can always just go grab whatever the actual current state of the autotask code is from defender

scripts/autotask.js
amiecorso: originally i'd set the permit deadline for one hour after the current block time. switched it to intmax as part of debugging just in case that was the problem and i was doing my math wrong or something.  I think it's TBD what we want the official deadline to be, and actually, something that Richie and i identified as an issue is the fact that if this relayer is handling a lot of cash transactions in parallel or near parallel, there's a race condition where the relayer will create more than one permit that use the same nonce.  so our actual solution here will potentially involve specifying the market as an approved operator for all of the relayer's NORI to handle cash transactions, or creating a large, long-lived permit from the relayer for the market to spend its funds

script/MintBPNori.s.sol
amiecorso: pretty sure it is... and I think using cast could make sense for something as simple as a deposit, but using a solidity script is going to make it so much simpler to handle something more complex and with much more complex arguments such as minting removals, especially if we want to make it a little bit easier to quickly parameterize, such as with the right number of different suppliers or removals or amounts.  running the solidity scripts is even easier than that cast command

jaycenhorton: For sure, there's plentiful use cases for scripting, just want to make sure we're not rewriting their API just to constrain it. They do have a pretty robust cli and w/ a pretty nice DX (you can encode the structs via the cli as well for minting). 

Not sold on needing a script to batch mint to multiple suppliers for local use beyond seeding, but this is a useful example nonetheless! :100:

jaycenhorton (approved):

PR #376, opened 2022-08-19
Fixed a couple of bugs with the deployments and swapped the naming of old and new LockedNORI 

scottynomad:
The new LockedNORI now owns the original name and the old one is called LockedNORIV1.

I *think* this is upgrading cleanly but I want to do some more testing.  One of the obstacles here is getting the right original storage layout into openzeppelin's artifacts as a basis to upgrade from.

### DIFF
--- a/config/ethernal.ts
+++ b/config/ethernal.ts
@@ -7,8 +7,7 @@ export const getEthernalConfig = (
   const ethernal: HardhatUserConfig['ethernal'] =
     ETHERNAL &&
     typeof ETHERNAL_EMAIL === 'string' &&
-    typeof ETHERNAL_PASSWORD === 'string' &&
-    ['localhost', 'hardhat'].includes(hre.network.name)
+    typeof ETHERNAL_PASSWORD === 'string'
       ? {
           email: ETHERNAL_EMAIL,
           password: ETHERNAL_PASSWORD,
--- a/script/DeployLockedNORIV1.s.sol
+++ b/script/DeployLockedNORIV1.s.sol
@@ -52,7 +52,7 @@ contract DeployLockedNORIV1 is Script {
 
   function _deployMockERC777() internal returns (MockERC777) {
     MockERC777 impl = new MockERC777();
-    bytes memory initializer = abi.encodeWithSignature("initialize()");
+abi.encodeWithSelector(impl.initialize.selector);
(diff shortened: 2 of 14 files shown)


config/ethernal.ts
```diff
@@ -7,8 +7,7 @@ export const getEthernalConfig = (
   const ethernal: HardhatUserConfig['ethernal'] =
     ETHERNAL &&
     typeof ETHERNAL_EMAIL === 'string' &&
-    typeof ETHERNAL_PASSWORD === 'string' &&
-    ['localhost', 'hardhat'].includes(hre.network.name)
```
jaycenhorton, line 11: does ethernal work on prod networks?

scottynomad: No but this blocks it working for a localhost standalone node because hre is not available at this stage of the config assembly.  I think relying on `ETHERNAL` being true is sufficient.

script/DeployLockedNORIV1.s.sol
```diff
@@ -0,0 +1,58 @@
...
+    bytes memory initializer = abi.encodeWithSelector(
+      impl.initialize.selector,
+      erc777_
+    );
+    return LockedNORIV1(_deployProxy(address(impl), initializer));
+  }
+
+  function _deployMockERC777() internal returns (MockERC777) {
+    MockERC777 impl = new MockERC777();
+    bytes memory initializer = abi.encodeWithSignature("initialize()");
```
jaycenhorton, line 55: I think this is a little cleaner + better type safety/intellisense support
```suggestion
abi.encodeWithSelector(impl.initialize.selector);
```

scottynomad:
github is not understanding the renaming fyi.

It actually went:

`contracts/LockedNORIV2.sol` -> `contracts/LockedNORI.sol`
`contreacts/deprecated/LockedNORI.sol` -> `contracts/deprecated/LockedNORIV1.sol`

jaycenhorton (approved):
some slither things might worth addressing-- otherwise lgtm

script/DeployLockedNORIV1.s.sol
scottynomad: Had to revert this.  `impl.initialize` is ambiguous so we need the version that includes the argument signature.

PR #378, opened 2022-08-26
[NO-2051] Audit responses

amiecorso:
Left some notes in the subtasks for each issue about this one... Most of these things I agreed with and were simple fixes.  The first one PVE001 I believe is a misunderstanding on behalf of the auditors and should not be fixed.  PVE004 may also not need to be fixed, as I believe we get the type of field-length validation logic by way of passing arguments to the removal ID creation as a tightly typed struct, such that each field is already restricted to its correct size and cannot overwrite other fields as the ID is assembled.  In fact, we may be able to get rid of the validation for methodology and methodology version that's already in there.  Need to confirm this one.  Finally, the admin keys is an issue that I believe we are simply ok with.  All of the other fixes were simple and seemed important.  Especially good catch on the `getAllRemovalIds` bug.


scottynomad (approved):
Thanks for addressing these @amiecorso !

contracts/Market.sol
```diff
@@ -335,6 +335,9 @@ contract Market is
     onlyRole(MARKET_ADMIN_ROLE)
     whenNotPaused
   {
+    if (noriFeePercentage_ > 100) {
```
jaycenhorton, line 338: should we also validate for > 0?

PR #381, opened 2022-08-31
fix bpNori deployment address in the depliyment artifact

jaycenhorton:
Noticed + needed in #380

### DIFF
--- a/deployments/mumbai/BridgedPolygonNORI.json
+++ b/deployments/mumbai/BridgedPolygonNORI.json
@@ -1,5 +1,5 @@
 {
-  "address": "0x8D7713C76A2A6Ec517FB3C032524dd86425E24f4",
+  "address": "0xBD16B6880F76CA8A3543dd7A2a60d375d0485dBB",
   "abi": [
     {
       "inputs": [],


amiecorso (approved):

scottynomad (changes requested):
Can you give more context please?  Is the new address the new contract on mumbai?  If so the rest of the information in this file is probably wrong too isn't it?  How did this get out of sync? How could we have avoided that?

jaycenhorton:
@scottynomad Pretty sure it's related to whatever issue @amiecorso  fixed when to get the market contract to have the correctly registered contract address. 

I just needed this change in #380 but since that isn't expected to get merged any time soon, I'd be ok fixing this with a re-deploy instead since we need to do that anyways (to use more appropriate compiler settings + either fix this issue altogether or suss out the underlying issue)

scottynomad (approved):
I still want us to do a clean redeploy of all the contracts on mumbai but if this unblocks things in the mean time I'm fine with that.

jaycenhorton:
To be clear this only unblocks local development as relates to using tenderly functionality (or anything else that locally uses this deployments file in any way).

PR #384, opened 2022-09-08
update hardhat dependencies

aqualen:
Please make sure the ones I changed are the right ones, and nothing important was missed.  I did try all my local builds.


jaycenhorton (changes requested):
I think you need to update the coverage plugin too:
https://github.com/NomicFoundation/hardhat/issues/3130

PR #389, opened 2022-09-13
[NO-2101]  add migrate function to the removal contract

jaycenhorton:
Other misc changes:
- adds `profile.test` to foundry.toml which enables a faster mode for testing. Use it like this: ` FOUNDRY_PROFILE=test forge test`
- uses foundry to test events using `vm.recordLogs();` and `vm.getRecordedLogs();`


contracts/Removal.sol
```diff
@@ -316,6 +318,81 @@ contract Removal is
...
+        _safeBatchTransferFrom({
+          from: owners[i],
+          to: _msgSender(),
+          ids: ids[i],
+          amounts: amounts[i],
+          data: ""
+        });
+        numberOfRemovals += ids[i].length;
+      }
+    }
```
jaycenhorton, line 367: :(

contracts/Removal.sol
```diff
@@ -316,6 +318,81 @@ contract Removal is
...
 
+  /**
+   * @notice Transfers the provided `amounts` of the specified removal `ids` directly to the Certificate contract to
+   * mint a legacy certificate.
+   *
+   * @dev The Certificate contract implements `onERC1155BatchReceived`, which is invoked upon receipt of any removals.
+   * This function circumvents the market contract's lifecycle. This function first transfers the removals from the
+   * supplier accounts to a single account that has the consignor role-- this is necessary because
+   * `_safeBatchTransferFrom` only accepts one `from` address and `Certificate.onERC1155BatchReceived` will mint a *new*
+   * certificate every time an additional batch is received.
```
jaycenhorton, line 329: This was a small oversight in our planning

contracts/Removal.sol
```diff
@@ -684,7 +763,8 @@ contract Removal is
   ) internal virtual override whenNotPaused {
     address market = address(_market);
     bool isToAllowed = to == market ||
-      (to == address(_certificate) || to == address(0));
+      (to == address(_certificate) || to == address(0)) ||
+      hasRole({role: CONSIGNOR_ROLE, account: _msgSender()});
```
jaycenhorton, line 767: Another small oversight

contracts/Removal.sol
```diff
@@ -483,7 +580,10 @@ contract Removal is
     uint256 amount,
     bytes memory data
   ) public override whenNotPaused {
-    if (_msgSender() != address(_market)) {
+    if (
+      _msgSender() != address(_market)
+      //&& hasRole({account: from, role: CONSIGNOR_ROLE})
```
rlfarman, line 585: Fix

contracts/Removal.sol
```diff
@@ -171,6 +173,23 @@ contract Removal is
     uint256 amount
   );
 
+  /**
+   * @notice Emitted when legacy removals are minted and then immediately used to migrate a legacy certificate.
+   *
+   * @param certificateRecipient The recipient of the certificate to mint via migration.
+   * @param certificateAmount The total amount of a the certificate to mint via migration.
```
amiecorso, line 180:
```suggestion
   * @param certificateAmount The total amount of the certificate to mint via migration.
```

contracts/Removal.sol
```diff
@@ -316,6 +335,85 @@ contract Removal is
...
+        });
+        numberOfRemovals += ids[i].length;
+      }
+    }
+    uint256[] memory flattenedAmounts = new uint256[](numberOfRemovals);
+    uint256[] memory flattenedIds = new uint256[](numberOfRemovals);
+    uint256 removalsTotal = 0;
+    // Skip overflow check as for loop is indexed starting at zero.
+    unchecked {
+      for (uint256 i = 0; i < flattenedAmounts.length; ++i) {
```
amiecorso, line 390: I think `i` will get too big here to be using as an index to `amounts` right?  `i` spans the full number of removals (as it is iterating to `flattenedAmounts.length`, but `amounts` only has a few subarrays

contracts/Removal.sol
```diff
@@ -316,6 +335,85 @@ contract Removal is
...
+        for (uint256 j = 0; j < amounts[i].length; ++j) {
+          removalsTotal += amounts[i][j];
+          flattenedAmounts[i] = amounts[i][j];
+          flattenedIds[i] = ids[i][j];
```
amiecorso, line 394: same for indexing into `ids`

contracts/Removal.sol
```diff
@@ -483,7 +580,10 @@ contract Removal is
     uint256 amount,
     bytes memory data
   ) public override whenNotPaused {
-    if (_msgSender() != address(_market)) {
+    if (
+      _msgSender() != address(_market) &&
+      hasRole({account: from, role: CONSIGNOR_ROLE})
```
amiecorso, line 585: is this logic right? this `&&` will be false if the sender (`from`) does NOT have the `CONSIGNOR_ROLE`, which would bypass the revert and succeed if some random sender tried to use this function

contracts/Removal.sol
```diff
@@ -683,8 +783,12 @@ contract Removal is
     bytes memory data
   ) internal virtual override whenNotPaused {
     address market = address(_market);
+    address certificate = address(_certificate);
     bool isToAllowed = to == market ||
-      (to == address(_certificate) || to == address(0));
+      to == certificate ||
+      to == address(0) ||
+      (hasRole({role: CONSIGNOR_ROLE, account: _msgSender()}) &&
+        (to == certificate || hasRole({role: CONSIGNOR_ROLE, account: to})));
```
amiecorso, line 791: maybe we should change the name of this variable to be `validTransfer` or something because the logic is no longer just about who `to` is?

amiecorso (commented):
still have to take a look at the test! but just wanted to submit some first review comments

contracts/Removal.sol
jaycenhorton: Oh, I don't think I even meant to change this function. Will revert the changes.

contracts/Removal.sol
jaycenhorton: Good catch. Tests were only accounting for a [3][1] array so this wasn't an issue there. Will adjust the tests to use a better test data set that will catch that kind of thing (e.g. [3][2])

contracts/Removal.sol
scottynomad, line 180: Can we make clear that this is denominated in NRTs?

contracts/Removal.sol
```diff
@@ -316,6 +335,85 @@ contract Removal is
     });
   }
 
+  /**
+   * @notice Transfers the provided `amounts` of the specified removal `ids` directly to the Certificate contract to
```
scottynomad, line 339: Can we expand on this to mention that this to cover:
* this is a one-off migration
* what a legacy certificate is
* that `amount` is denominated in NRTs.

contracts/Removal.sol
```diff
@@ -316,6 +335,85 @@ contract Removal is
...
+   * mint a legacy certificate.
+   *
+   * @dev The Certificate contract implements `onERC1155BatchReceived`, which is invoked upon receipt of any removals.
+   * This function circumvents the market contract's lifecycle. This function first transfers the removals from the
```
scottynomad, line 343:
```suggestion
   * This function circumvents the market contract's lifecycle by transferring the removals from the
```

contracts/Removal.sol
```diff
@@ -316,6 +335,85 @@ contract Removal is
...
+   * @param owners The owners of the specified removal IDs to transfer from.
+   * @param ids 2D array of the removal IDs to add to transfer to the Certificate contract. Each row in the array
+   * represents a list of one owners removal IDs.
+   * @param amounts 2D array of the balances of each token ID to transfer to the Certificate contract. Each row in the
+   * array represents a list of one owner's removal amounts, each of which corresponds to the same row and column
+   * defined in `ids`.
+   * @param certificateRecipient The recipient of the certificate to be minted.
+   * @param certificateAmount TThe total amount of the certificate.
+   */
+  function migrate(
```
scottynomad, line 363: Given that the owner address is encoded in the recipient id aren't `owners` and the the first dimension of `ids` redundant ?

scottynomad: It certainly makes the code short given this structure but we'd need to validate the arguments.

contracts/Removal.sol
scottynomad: Agreed.  I would be in favor of splitting up this logic so it's easier to test and reason about.

scottynomad:
Can you update `package.json` to pass the test profile when on `yarn test`?

contracts/Removal.sol
```diff
@@ -171,6 +173,23 @@ contract Removal is
...
+  /**
+   * @notice Emitted when legacy removals are minted and then immediately used to migrate a legacy certificate.
+   *
+   * @param certificateRecipient The recipient of the certificate to mint via migration.
+   * @param certificateAmount The total amount of a the certificate to mint via migration.
+   * @param certificateId The ID of the certificate to mint via migration.
+   * @param removalIds The removal IDs to use to mint the certificate via migration.
+   * @param removalAmounts The amounts for each corresponding removal ID to use to mint the certificate via migration.
+   */
+  event Migration(
```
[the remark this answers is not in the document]
jaycenhorton: That was a mistake IMO-- all OZ events are in an active voice

contracts/Removal.sol
jaycenhorton: it would require yet another loop where we have to loop through all the ids and sort them on-chain to get rebuild this argument. Is it really worth it given this function is already extemely expensive (worried about gas limits, not cost) as is?

jaycenhorton: as it stands, it's about 2.5m gas for a call to `migrate` that spans 5 removals across 2 suppliers.

scottynomad: Well, let's test with the worst case (well known) certificate and make decisions on the need for optimization based on that.

It's not pretty as it but if it does that trick and only gets used once nbd.

test/Removal.t.sol
```diff
@@ -3,12 +3,128 @@
...
+      owners: suppliers,
+      ids: idsForAllSuppliers,
+      amounts: amountsForAllSuppliers,
+      certificateRecipient: _namedAccounts.buyer,
+      certificateAmount: certificateAmount
+    });
+    Vm.Log[] memory entries = vm.getRecordedLogs();
+    assertEq(entries.length, 6);
+    assertEq(
+      entries[2].topics[0],
```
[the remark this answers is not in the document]
jaycenhorton: I think you can use `expectEmit` for that but without refactoring our repo to use interfaces there doesnt seem to be a great way to do that without re-defining the event 

https://book.getfoundry.sh/cheatcodes/expect-emit?highlight=event#examples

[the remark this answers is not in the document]
jaycenhorton: in either case I think I'd wait for a better test API to come around until preferring one or the other. Keeping this as is so that we have a living example floating around for this approach (we already have examples in other places for expectEmit)

contracts/Removal.sol
jaycenhorton: Just ran a test based on this data you sent over: 
```json
{
  "certificateId": "5962695850328064",
  "certificateUri": "https://nori.com/certificate/1394",
  "removalCount": "190",
  "NRTs": "15000.0",
  "projecCount": "7"
}
```
and we're sitting at 44m gas (~double the block gas limit). Even as is we need to optimize for this to work.

jaycenhorton:
Replaced by #394

PR #390, opened 2022-09-14
[NO-2095] Allow 0-amount removal minting to suppliers

amiecorso:
Tweaks the logic in `Removal._beforeTokenTransfer` so that only transfers to the Market contract cannot have an amount of 0. This allows us to mint 0-amount removals for projects so that those removals will appear in event logs as part of the project for transparent recordkeeping.

### DIFF
--- a/test/Removal.t.sol
+++ b/test/Removal.t.sol
@@ -94,6 +94,60 @@ contract Removal_mintBatch_reverts_mint_to_wrong_address is UpgradeableMarket {
   }
 }
 
+contract Removal_mintBatch_zero_amount_removal is UpgradeableMarket {
+  function test() external {
+    DecodedRemovalIdV0[] memory ids = new DecodedRemovalIdV0[](1);
+    ids[0] = DecodedRemovalIdV0({
+      idVersion: 0,
+      methodology: 1,
+      methodologyVersion: 0,
+      vintage: 2018,
+      country: "US",
+      subdivision: "IA",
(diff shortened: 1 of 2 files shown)


test/Removal.t.sol
```diff
@@ -94,6 +94,60 @@ contract Removal_mintBatch_reverts_mint_to_wrong_address is UpgradeableMarket {
...
+    DecodedRemovalIdV0[] memory ids = new DecodedRemovalIdV0[](1);
+    ids[0] = DecodedRemovalIdV0({
+      idVersion: 0,
+      methodology: 1,
+      methodologyVersion: 0,
+      vintage: 2018,
+      country: "US",
+      subdivision: "IA",
+      supplierAddress: _namedAccounts.supplier,
+      subIdentifier: _REMOVAL_FIXTURES[0].subIdentifier + 1
```
jaycenhorton, line 108: The `+ 1` isnt needed here

test/Removal.t.sol
```diff
@@ -94,6 +94,60 @@ contract Removal_mintBatch_reverts_mint_to_wrong_address is UpgradeableMarket {
...
```
jaycenhorton, line 134: The `+ 1` isnt needed here

jaycenhorton (approved):

test/Removal.t.sol
amiecorso: ah right that was from pulling from a different test that was trying to mint a second removal and keep it unique

PR #393, opened 2022-09-20
Reduce scope of migrate fn; Add slight gas optimizations; Add largest known migration test

jaycenhorton:
- The test is currently set to fail with the gas used for `migrate` is >= 30m. This can be used as a harness to make sure we fix gas consumption pre migration.


jaycenhorton:
replaced by #394

PR #394, opened 2022-09-21
[NO-2101] remove state from certificate and add migration functionality

jaycenhorton:
- Removes state from certificate as outlined in [option 1](https://www.notion.so/teamnori/1-Stop-using-state-to-track-which-removals-belong-to-which-certificates-and-vice-versa-8dfa94d1adc74a7487f9118e77029054)
- bumps erc721a version

Replaces/combines #393 and #389


scottynomad:
Looks reasonable to me @jaycenhorton.  Let's get the state removal merged and tests updated first and then reintroduce the migration.

test/Removal.t.sol
```diff
@@ -649,11 +844,59 @@ contract Removal_release_unlisted_listed_and_retired is UpgradeableMarket {
...
+        );
+        assertEq(entries[i].topics[3], bytes32(uint256(uint160(address(0)))));
+        (uint256 id, uint256 value) = abi.decode(
+          entries[i].data,
+          (uint256, uint256)
+        );
+        assertEq(id, _removalIds[0]);
+        assertEq(value, _expectedReleasedBalances[i / 2]);
+      }
+    }
```
jaycenhorton, line 894: Fairly robust reference on testing event emissions using forge for the future

test/helpers/removal.sol
```diff
@@ -147,6 +147,38 @@ abstract contract UpgradeableRemoval is Upgradeable {
     return _removalIds;
   }
 
+  function _seedRemovals(
```
rlfarman, line 150: 👍

rlfarman (approved):

PR #397, opened 2022-09-22
Trying out a different way to handle rcp urls in networks.ts

scottynomad:
Thoughts on this kind of approach to handling rcp urls?  Alchemy ties an api key to a network afaik so this gives us some more flexibility.

Another option would be to use a single env variable like we do in nori-graphql (`WEB3_RPC_ENDPOINT`) but I think that would end up being a fairly frustrating dev ex in this repo for anyone interacting with more than one network.


jaycenhorton (commented):
seems reasonable

jaycenhorton (approved):

PR #398, opened 2022-09-23
cleanup readme; add rpc config for foundry; improve hardhat accounts task;

jaycenhorton:
### Readme overhaul

- took a stab at making the main readme easier to follow along with
- added more cast and hardhat examples

### Foundry config overhaul

If your .env files are loaded into your environment, you can use:

- `cast send --rpc-url mumbai` note that mumbai/goerli/mainnet/polygon expand to the correct alchemy endpoints. This works for other foundry commands beyond `cast send`
- some foundry commands will now autoload the `MNEMONIC` environment variable if it's set. When not, you can use either `--mnemonics=$MNEMONIC` or `--mnemonic=$MNEMONIC ` (forge script uses the former, cast send uses the latter, other commands might use even more still as config support seems to be incomplete in the foundry ecosystem) 

### `hardhat accounts`

image


amiecorso (approved):

amiecorso:
beauty

foundry.toml
```diff
@@ -24,4 +24,10 @@ optimizer_runs = 0
 optimizer = false
 sparse_mode = true
 
+[rpc_endpoints]
+mumbai = "${MUMBAI_RPC_URL}"
```
scottynomad, line 28: thanks.   sorry I forgot about foundry. :(

scottynomad (approved):

PR #400, opened 2022-09-26
[NO-2166] Update vesting cli to use permit signatures for grant creation

amiecorso:
** See my comment below to easily find the line numbers in `vesting.ts` that involve the important update -- there's a ton of noise in this PR because most of the affected lines were simply uncommenting.

Updates the vesting task to correctly create grants for the current version of `LockedNORI.sol`.  This is now done by calling `LockedNORI.batchCreateGrants` and supplying a valid permit from the transaction signer for `LockedNORI` to transfer some of the signer's `BridgedPolygonNORI` to itself. 

Note -- because our local hardhat node and tests don't currently work with a Fireblocks signer (get named account errors, @Scott is aware of this) I haven't actually tested whether this works with a real fireblocks signer.  I was able to test successfully against a local hardhat node using a mnemonic and see that grants were correctly created.

```
Using fireblocks signer
Fireblocks signer created for address: 0x0F032F48fD4b38eA605F438922CA19FA79d0e6A7 (mumbai)
LockedNORI: 0xCdcB43CB7B668F0C1cA04FE4b60DA7F8C62Be393
Reading grants from file
Grants obtained
Total supply: 70868901389056055203209747
WARNING: total supply of LockedNORI does not equal the amounts of all grants.  Was a line removed from grants CSV?
Total supply: 70868901.389056055203209747
Total amount of grants provided: 0.0
Found 1 grants that need updating
Total bpNORI to lock: 46425.5886
🎉 Dry run was successful!
```


tasks/vesting.ts
```diff
@@ -1,1028 +1,1069 @@
...
+        );
+      });
+      const userData = grantDiffs.map((grant) => buildUserData({ grant }));
+      const operatorData = grantDiffs.map((_) => '0x');
+      const requireReceptionAck = grantDiffs.map((_) => true);
+      hre.log(
+        `Total bpNORI to lock: ${ethers.utils.formatEther(
+          amounts.reduce((accumulator, v) => accumulator.add(v))
+        )}`
+      );
```
amiecorso, line 838: hard to see the actual diff because of all the uncommenting in this file but below this comment is where the actual changes that involve the permit are (through line 906 where `batchCreateGrants` is also called in the dryrun situation)

scottynomad: Nice!  Might be nice to pull the signing logic out into a separate function but other than that looks great!  I'll pull it this evening and try to sort out the named account thing.

jaycenhorton: [there's a helper util for this in utils/math.ts](https://github.com/nori-dot-eco/contracts/blob/master/utils/math.ts#L7)

jaycenhorton (commented):
looks good to me -- will defer to @scottynomad for the fireblocks stuff

scottynomad (approved):

PR #406, opened 2022-10-10
[NO-2117] `migrate-removals` hardhat task prototype

amiecorso:
This PR introduces a hardhat task that consumes JSON data that specifies removals to be minted for multiple projects.

The task can be run as follows:

```
hardhat --network [network] migrate-removals --file [JSON data filename] [--dry-ryn] [--output-file] [output data filename]
```
(where `--dry-run` is an optional flag that will simulate the transactions instead of actually run them and report their success and `--output-file` is an optional parameter that can be used to specify something other than a default output filename)

TODO: 
- write some documentation for this task
- specify the input and output file formats
- specify other contingencies of using this script

### DIFF
--- a/tasks/migrate-removals.ts
+++ b/tasks/migrate-removals.ts
@@ -1,0 +1,184 @@
+/* eslint-disable no-await-in-loop -- need to submit transactions synchronously to avoid nonce collisions */
+import { readFileSync, writeFileSync } from 'fs';
+
+import { divide } from 'mathjs';
+import { task, types } from 'hardhat/config';
+import type { BigNumber } from 'ethers';
+import { ethers } from 'ethers';
+import type { Signer } from '@ethersproject/abstract-signer';
+import chalk from 'chalk';
+
+import type { FireblocksSigner } from '../plugins/fireblocks/fireblocks-signer';
+
+interface MigrateRemovalsTaskOptions {
+  file: string;
+  outputFile?: string;
+  dryRun?: boolean;
+}
+
+type ParsedMigrateRemovalsTaskOptions = RequiredKeys<
+  MigrateRemovalsTaskOptions,
+  'file'
+>;
+
+const asciiStringToHexString = (ascii: string): string => {
+  return `0x${[...Array.from({ length: ascii.length }).keys()]
+    .map((index) => ascii.charCodeAt(index).toString(16))
+    .join('')}`;
+};
+
+const parseTransactionLogs = ({
+  contractInstance,
+  txReceipt,
+}: {
+  contractInstance: ethers.Contract;
+  txReceipt: ethers.providers.TransactionReceipt;
+}): ethers.utils.LogDescription[] => {
+  return txReceipt.logs
+    .filter((log) => log.address === contractInstance.address)
+    .map((log) => contractInstance.interface.parseLog(log));
+};
+
+export const GET_MIGRATE_REMOVALS_TASK = () =>
+  ({
+    name: 'migrate-removals',
+    description: 'Utility to mint legacy removals',
+    run: async (
+      options: MigrateRemovalsTaskOptions,
+      _: CustomHardHatRuntimeEnvironment
+    ): Promise<void> => {
+      const {
+        file,
+        outputFile = 'migrated-removals.json',
+        dryRun,
+      } = options as ParsedMigrateRemovalsTaskOptions;
+      const jsonData = JSON.parse(readFileSync(file, 'utf8'));
+      // console.log({ jsonData });
+
+      const [signer] = await hre.getSigners();
+      const { getRemoval } = await import('@/utils/contracts');
+      const removalContract = await getRemoval({
+        hre,
+        signer,
+      });
+      hre.log(`Removal contract address: ${removalContract.address}`);
+      // const fireblocksSigner = removalContract.signer as FireblocksSigner;
+
+      // submit all minting transactions synchronously to avoid nonce collision
+      const pendingTransactions = [];
+      for (const project of jsonData) {
+        const amounts = project.amounts.map((amount: string) =>
+          ethers.utils
+            .parseUnits(divide(Number(amount), 1_000_000).toString())
+            .toString()
+        );
+
+        const removals = project.removals.map((removal: any, index) => {
+          const removalData = {
+            idVersion: Number(removal.idVersion),
+            // methodology: Number(removal.methodology), // TODO
+            // methodologyVersion: Number(removal.methodologyVersion), // TODO
+            methodology: 1,
+            methodologyVersion: 0,
+            vintage: Number(removal.vintage),
+            country: asciiStringToHexString(removal.country),
+            subdivision: asciiStringToHexString(
+              removal.subdivision.slice(0, 2)
+            ), // TODO we only want the state code here
+            supplierAddress: '0x9A232b2f5FbBf857d153Af8B85c16CBDB4Ffb667', // TODO
+            subIdentifier: 2_004_318_071 + index, // TODO get from removal
+          };
+          // console.log({ index, removalData });
+          return removalData;
+        });
+
+        if (!dryRun) {
+          let pendingTx: ethers.ContractTransaction;
+          try {
+            pendingTx = await removalContract.mintBatch(
+              '0x9A232b2f5FbBf857d153Af8B85c16CBDB4Ffb667', // TODO need real addresses on the project
+              amounts,
+              removals,
+              project.projectId,
+              // firstProject.scheduleStartTime, // TODO this can't be 0
+              1_665_441_893,
+              project.holdbackPercentage
+            );
+            pendingTransactions.push(pendingTx);
+          } catch (error) {
+            console.error('Error submitting mintBatch');
+            console.error(error);
+          }
+        } else {
+          // dry run
+          try {
+            await removalContract.callStatic.mintBatch(
+              '0x9A232b2f5FbBf857d153Af8B85c16CBDB4Ffb667', // TODO need real addresses on the project
+              amounts,
+              removals,
+              project.projectId,
+              // firstProject.scheduleStartTime, // TODO this can't be 0
+              1_665_441_893, // default to now
+              project.holdbackPercentage
+            );
+            hre.log(
+              chalk.bold.bgWhiteBright.black(`🎉  Dry run was successful!`)
+            );
+          } catch (error) {
+            hre.log(
+              chalk.bold.bgRed.black(`💀 Dry run was unsuccessful!`, error)
+            );
+          }
+        }
+      }
+      if (!dryRun) {
+        // asynchronously await the completion of all transactions
+        const txResults = await Promise.all(
+          pendingTransactions.map(async (tx) => {
+            const result = await tx.wait(); // TODO specify more than one confirmation?
+            const txReceipt =
+              await removalContract.provider.getTransactionReceipt(
+                result.transactionHash
+              );
+            // TODO make sure the status on this receipt is 1 (success)?
+            // console.log({ txReceipt });
+            const tokenIds = parseTransactionLogs({
+              contractInstance: removalContract,
+              txReceipt,
+            })
+              .filter((log) => log.name === 'TransferBatch')
+              .flatMap((log) =>
+                log.args.ids.map((id: BigNumber) => id.toHexString())
+              );
+            // console.log({ tokenIds });
+            return {
+              txReceipt,
+              tokenIds,
+            };
+          })
+        );
+        const mintingResults = jsonData.map((project, index) => {
+          const txResult = txResults[index];
+          return {
+            ...project,
+            txReceipt: txResult.txReceipt,
+            tokenIds: txResult.tokenIds,
+          };
+        });
+        writeFileSync(outputFile, JSON.stringify(mintingResults));
+      }
+    },
+  } as const);
+
+(() => {
+  const { name, description, run } = GET_MIGRATE_REMOVALS_TASK();
+  task(name, description, run)
+    .addParam(
+      'file',
+      'JSON removal data to read',
+      undefined,
+      types.string,
+      false
+    )
+    .addFlag('dryRun', 'simulate the transaction without actually sending it');
+})();
--- a/config/networks.ts
+++ b/config/networks.ts
@@ -12,7 +12,7 @@ const {
 } = process.env;
 
 const hardhat: NetworksUserConfig['hardhat'] = {
-  blockGasLimit: 20_000_000,
+  blockGasLimit: 30_000_000,
   initialBaseFeePerGas: 1,
   gasPrice: 3,
   chainId: 9001,
@@ -24,7 +24,7 @@ const hardhat: NetworksUserConfig['hardhat'] = {
 };
 
 const localhost: NetworkUserConfig = {
-  blockGasLimit: 20_000_000,
+  blockGasLimit: 30_000_000,
   initialBaseFeePerGas: 1,
   gasPrice: 3,
   chainId: 9001,
--- a/tasks/index.ts
+++ b/tasks/index.ts
@@ -21,7 +21,8 @@ import { TASK as NORI_TASK } from './nori';
 import { TASK as LOCKED_NORI_TASK } from './locked-nori'; // todo make work with forked repo
 import { TASK as BRIDGED_POLYGON_NORI_TASK } from './bridged-polygon-nori';
 import { DEFENDER_ADD_TASK } from './defender';
-import { GET_VESTING_TASK as VESTING_TASK} from './vesting';  // todo make work with forked repo
+import { GET_VESTING_TASK as VESTING_TASK } from './vesting'; // todo make work with forked repo
+import { GET_MIGRATE_REMOVALS_TASK } from './migrate-removals';
 import { TASK as FORCE_UPGRADE_TASK } from './force-ugrade';
 import { TASK as SIGN_MESSAGE_TASK } from './sign-message';
 import { TASK as TEST_SIGN_TYPED_TASK } from './test-sign-typed';
@@ -69,10 +70,11 @@ export const TASKS = {
   [MARKET_TASK.name]: { ...MARKET_TASK },
   [REMOVAL_TASK.name]: { ...REMOVAL_TASK },
   [NORI_TASK.name]: { ...NORI_TASK },
-  [LOCKED_NORI_TASK.name]: { ...LOCKED_NORI_TASK },  // todo make work with forked repo
+  [LOCKED_NORI_TASK.name]: { ...LOCKED_NORI_TASK }, // todo make work with forked repo
   [BRIDGED_POLYGON_NORI_TASK.name]: { ...BRIDGED_POLYGON_NORI_TASK },
   [DEFENDER_ADD_TASK.name]: { ...DEFENDER_ADD_TASK },
-  [VESTING_TASK.name]: { ...VESTING_TASK },  // todo make work with forked repo
+  [VESTING_TASK.name]: { ...VESTING_TASK }, // todo make work with forked repo
+  [GET_MIGRATE_REMOVALS_TASK.name]: { ...GET_MIGRATE_REMOVALS_TASK },
   [FORCE_UPGRADE_TASK.name]: { ...FORCE_UPGRADE_TASK },
   [SIGN_MESSAGE_TASK.name]: { ...SIGN_MESSAGE_TASK },
   [TEST_SIGN_TYPED_TASK.name]: { ...TEST_SIGN_TYPED_TASK },


tasks/list-remaining-migrated-removals.ts
```diff
@@ -0,0 +1,207 @@
...
+      const [signer] = await hre.getSigners();
+      const signerAddress = await signer.getAddress();
+      const { getRemoval } = await import('@/utils/contracts');
+      const removalContract = await getRemoval({
+        hre,
+        signer,
+      });
+      hre.log(`Removal contract address: ${removalContract.address}`);
+      hre.log(`Signer address: ${signerAddress}`);
+      // const fireblocksSigner = removalContract.signer as FireblocksSigner;
```
rlfarman, line 46: rm ?

tasks/migrate-certificates.ts
```diff
@@ -0,0 +1,227 @@
...
+      options: MigrateCertificatesTaskOptions,
+      _: CustomHardHatRuntimeEnvironment
+    ): Promise<void> => {
+      const {
+        file,
+        outputFile = 'migrated-certificates.json',
+        dryRun,
+      } = options as ParsedMigrateCertificatesTaskOptions;
+      const jsonData = JSON.parse(readFileSync(file, 'utf8'));
+      // hre.log({ jsonData });
```
rlfarman, line 40: rm ?

tasks/migrate-certificates.ts
```diff
@@ -0,0 +1,227 @@
...
+        signer,
+      });
+      const certificateContract = await getCertificate({
+        hre,
+        signer,
+      });
+      hre.log(`Removal contract address: ${removalContract.address}`);
+      hre.log(`Certificate contract address: ${certificateContract.address}`);
+      hre.log(`Signer address: ${signerAddress}`);
+      // const fireblocksSigner = removalContract.signer as FireblocksSigner;
```
rlfarman, line 56: rm ?

tasks/migrate-removals.ts
```diff
@@ -0,0 +1,252 @@
...
+              // projectIndex === 1 ? 0 : 1_665_441_893, // <-- to induce a revert
+              1_665_441_893,
+              project.holdbackPercentage
+            );
+            hre.log(
+              `(${projectIndex}/${jsonData.length}) projectId: ${project.projectId} 🎉  Dry run was successful!`
+            );
+          } catch {
+            hre.log(
+              `(${projectIndex}/${jsonData.length}) projectId: ${project.projectId} 💀 Dry run was unsuccessful!`
```
rlfarman, line 147: 💀 I love it

tasks/list-remaining-migrated-removals.ts
amiecorso: I think we're going to actually want this when we do the migration on polygon / maybe even mumbai because the deployer is the admin is the fireblocks signer... we can use the vesting cli task as an example

tasks/list-remaining-migrated-removals.ts
```diff
@@ -0,0 +1,207 @@
...
+      options: ListMigratedRemovalsTaskOptions,
+      _: CustomHardHatRuntimeEnvironment
+    ): Promise<void> => {
+      const {
+        file,
+        outputFile = 'listed-migrated-removals.json',
+        dryRun,
+      } = options as ParsedListMigratedRemovalsTaskOptions;
+      const jsonData = JSON.parse(readFileSync(file, 'utf8'));
+      // hre.log({ jsonData });
```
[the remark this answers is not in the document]
amiecorso: sure I will pull out any commented logs.  was thinking we might end up toggling and untoggling these a few more times before we're really at the end of this but it's easy enough to add back in

tasks/migrate-removals.ts
amiecorso: haha stole this from jaycen's logging in the vesting cli

tasks/list-remaining-migrated-removals.ts
```diff
@@ -0,0 +1,207 @@
...
+    run: async (
+      options: ListMigratedRemovalsTaskOptions,
+      _: CustomHardHatRuntimeEnvironment
+    ): Promise<void> => {
+      const {
+        file,
+        outputFile = 'listed-migrated-removals.json',
+        dryRun,
+      } = options as ParsedListMigratedRemovalsTaskOptions;
+      const jsonData = JSON.parse(readFileSync(file, 'utf8'));
```
jaycenhorton, line 34:
```suggestion
      const jsonData = readJsonSync(file)
```

tasks/list-remaining-migrated-removals.ts
```diff
@@ -0,0 +1,207 @@
...
+      // hre.log({ jsonData });
+
+      const [signer] = await hre.getSigners();
+      const signerAddress = await signer.getAddress();
+      const { getRemoval } = await import('@/utils/contracts');
```
jaycenhorton, line 39: why isn't this imported at the top-level?

tasks/list-remaining-migrated-removals.ts
jaycenhorton: import { readJsonSync } from 'fs-extra';

tasks/list-remaining-migrated-removals.ts
```diff
@@ -0,0 +1,207 @@
...
+
+      const listableTokenIds = listableData.map((data) => data.tokenId);
+      const listableBalances = listableData.map((data) => data.balance);
+      // sum the listable balances and convert to ether
+      const totalListableBalance = listableBalances.reduce(
+        (accumulator, balance) => accumulator.add(balance),
+        ethers.BigNumber.from(0)
+      );
+      const totalListableBalanceInEther =
+        ethers.utils.formatEther(totalListableBalance);
```
jaycenhorton, line 98:
```suggestion
        formatTokenAmount(totalListableBalance);
```

tasks/list-remaining-migrated-removals.ts
```diff
@@ -0,0 +1,207 @@
...
+          chalk.white(
+            `👋 No listable token ids found (no non-zero balances), exiting without listing any removals...`
+          )
+        );
+        return;
+      }
+      if (!dryRun) {
+        hre.log(chalk.white(`🤞 Submitting multicall consign transaction...`));
+        let txResult;
+        let pendingTx: ethers.ContractTransaction;
```
jaycenhorton, line 117: looks like this can be const inside the try block

tasks/list-remaining-migrated-removals.ts
```diff
@@ -0,0 +1,207 @@
...
+          }
+          txResult = txReceipt;
+        } catch (error) {
+          hre.log(
+            chalk.red('❌ Error submitting multicall consign transaction: '),
+            error
+          );
+          txResult = error;
+        }
+        writeFileSync(
```
jaycenhorton, line 158: writeJsonSync

tasks/list-remaining-migrated-removals.ts
```diff
@@ -0,0 +1,207 @@
...
+        hre.log(
+          chalk.white(
+            `👋 No listable token ids found (no non-zero balances), exiting without listing any removals...`
+          )
+        );
+        return;
+      }
+      if (!dryRun) {
+        hre.log(chalk.white(`🤞 Submitting multicall consign transaction...`));
+        let txResult;
```
jaycenhorton, line 116: missing type

tasks/list-remaining-migrated-removals.ts
```diff
@@ -0,0 +1,207 @@
...
+      const {
+        file,
+        outputFile = 'listed-migrated-removals.json',
+        dryRun,
+      } = options as ParsedListMigratedRemovalsTaskOptions;
+      const jsonData = JSON.parse(readFileSync(file, 'utf8'));
+      // hre.log({ jsonData });
+
+      const [signer] = await hre.getSigners();
+      const signerAddress = await signer.getAddress();
```
jaycenhorton, line 38: wonder if we should do validation on network and address (e.g., to verify the address used is the consignor, on a valid network, and uses fireblocks depending on the network)

tasks/list-remaining-migrated-removals.ts
```diff
@@ -0,0 +1,207 @@
...
+              )
+            );
+          }
+          txResult = txReceipt;
+        } catch (error) {
+          hre.log(
+            chalk.red('❌ Error submitting multicall consign transaction: '),
+            error
+          );
+          txResult = error;
```
jaycenhorton, line 156: should this throw?

tasks/list-remaining-migrated-removals.ts
```diff
@@ -0,0 +1,207 @@
...
+          )
+        );
+        return;
+      }
+      if (!dryRun) {
+        hre.log(chalk.white(`🤞 Submitting multicall consign transaction...`));
+        let txResult;
+        let pendingTx: ethers.ContractTransaction;
+        try {
+          pendingTx = await removalContract.multicall(
```
jaycenhorton, line 119:
```suggestion
          const functionOrDryRun = dryRun ? removalContract.multicall : removalContract.callStatic.multicall;
          pendingTx = await functionOrDryRun(
```

Then delete the if/else blocks. Also wonder if we should always call callStatic before actually executing regardless if the flag is used.

tasks/migrate-certificates.ts
```diff
@@ -0,0 +1,227 @@
+/* eslint-disable no-await-in-loop -- need to submit transactions synchronously to avoid nonce collisions */
```
jaycenhorton, line 1: see notes about dryRun and validation in list-remaining-migrated-removals

tasks/migrate-certificates.ts
```diff
@@ -0,0 +1,227 @@
...
+                certificate.id
+              } txHash: ${chalk.green(pendingTx.hash)}`
+            );
+          } catch (error) {
+            hre.log(
+              `(${certificateIndex}/${jsonData.length}) datastore ID: ${
+                certificate.id
+              } ${chalk.red('error')}`
+            );
+            pendingTransactions.push(error);
```
jaycenhorton, line 106: should we throw here instead to guarantee on-chain ordering of certificates?

tasks/migrate-certificates.ts
```diff
@@ -0,0 +1,227 @@
...
+            } Transaction status: ${
+              resultForCertificate.txReceiptOrError.status ?? 'error'
+            } Token ID: ${resultForCertificate.tokenId}`
+          );
+          return {
+            ...certificate,
+            ...resultForCertificate,
+          };
+        });
+        writeFileSync(outputFile, JSON.stringify(migrateResults));
```
jaycenhorton, line 209: writeJsonSync

tasks/migrate-certificates.ts
```diff
@@ -0,0 +1,227 @@
...
+import { readFileSync, writeFileSync } from 'fs';
+
+import cliProgress from 'cli-progress';
+import { divide } from '@nori-dot-com/math';
+import { task, types } from 'hardhat/config';
+import { ethers } from 'ethers';
+import chalk from 'chalk';
+import { parseTransactionLogs } from '@nori-dot-com/contracts/utils/events';
+
+import type { FireblocksSigner } from '../plugins/fireblocks/fireblocks-signer';
```
jaycenhorton, line 11:
```suggestion
import type { FireblocksSigner } from '@/plugins/fireblocks/fireblocks-signer';
```

tasks/migrate-certificates.ts
```diff
@@ -0,0 +1,227 @@
...
+        file,
+        outputFile = 'migrated-certificates.json',
+        dryRun,
+      } = options as ParsedMigrateCertificatesTaskOptions;
+      const jsonData = JSON.parse(readFileSync(file, 'utf8'));
+      // hre.log({ jsonData });
+
+      const [signer] = await hre.getSigners();
+      const signerAddress = await signer.getAddress();
+      const { getRemoval } = await import('@/utils/contracts');
```
jaycenhorton, line 44: why isn't this imported top-level?

tasks/migrate-removals.ts
```diff
@@ -0,0 +1,252 @@
+/* eslint-disable no-await-in-loop -- need to submit transactions synchronously to avoid nonce collisions */
```
jaycenhorton, line 1: see notes about dryRun and validation in list-remaining-migrated-removals

tasks/migrate-removals.ts
```diff
@@ -0,0 +1,252 @@
...
+    run: async (
+      options: MigrateRemovalsTaskOptions,
+      _: CustomHardHatRuntimeEnvironment
+    ): Promise<void> => {
+      const {
+        file,
+        outputFile = 'migrated-removals.json',
+        dryRun,
+      } = options as ParsedMigrateRemovalsTaskOptions;
+      const jsonData = JSON.parse(readFileSync(file, 'utf8'));
```
jaycenhorton, line 45: readJsonSync

tasks/migrate-removals.ts
```diff
@@ -0,0 +1,252 @@
...
+
+import cliProgress from 'cli-progress';
+import { divide } from '@nori-dot-com/math';
+import { task, types } from 'hardhat/config';
+import type { BigNumber } from 'ethers';
+import { ethers } from 'ethers';
+import chalk from 'chalk';
+import { parseTransactionLogs } from '@nori-dot-com/contracts/utils/events';
+
+import type { FireblocksSigner } from '../plugins/fireblocks/fireblocks-signer';
```
jaycenhorton, line 12:
```suggestion
import type { FireblocksSigner } from '@/plugins/fireblocks/fireblocks-signer';
```

tasks/list-remaining-migrated-removals.ts
```diff
@@ -0,0 +1,207 @@
+import { readFileSync, writeFileSync } from 'fs';
+
+import { task, types } from 'hardhat/config';
+import chalk from 'chalk';
+import { ethers } from 'ethers';
+
+import type { FireblocksSigner } from '../plugins/fireblocks/fireblocks-signer';
```
jaycenhorton, line 7:
```suggestion
import type { FireblocksSigner } from '@/plugins/fireblocks/fireblocks-signer';
```

tasks/migrate-removals.ts
```diff
@@ -0,0 +1,252 @@
...
+import { getRemoval } from '../utils/contracts';
```
jaycenhorton, line 13:
```suggestion
import { getRemoval } from '@/utils/contracts';
```

tasks/migrate-removals.ts
```diff
@@ -0,0 +1,252 @@
+/* eslint-disable no-await-in-loop -- need to submit transactions synchronously to avoid nonce collisions */
+import { readFileSync, writeFileSync } from 'fs';
+
+import cliProgress from 'cli-progress';
+import { divide } from '@nori-dot-com/math';
+import { task, types } from 'hardhat/config';
+import type { BigNumber } from 'ethers';
+import { ethers } from 'ethers';
+import chalk from 'chalk';
+import { parseTransactionLogs } from '@nori-dot-com/contracts/utils/events';
```
jaycenhorton, line 10:
```suggestion
import { parseTransactionLogs } from '@/utils/events';
```

tasks/migrate-removals.ts
```diff
@@ -0,0 +1,252 @@
...
+          )
+        );
+      }
+      // submit all mintBatch transactions serially to avoid nonce collision!
+      const pendingTransactions = [];
+      let projectIndex = 1;
+      for (const project of jsonData) {
+        const amounts = project.amounts.map((amount: string) =>
+          ethers.utils
+            .parseUnits(divide(Number(amount), 1_000_000).toString())
```
jaycenhorton, line 78: use BigNumber.from(amount).div instead

tasks/migrate-removals.ts
```diff
@@ -0,0 +1,252 @@
...
+            // methodology: Number(removal.methodology), // TODO
+            // methodologyVersion: Number(removal.methodologyVersion), // TODO
+            methodology: 1,
+            methodologyVersion: 0,
+            vintage: Number(removal.vintage),
+            country: asciiStringToHexString(removal.country),
+            subdivision: asciiStringToHexString(
+              removal.subdivision.slice(0, 2)
+            ), // TODO we only want the state code here, original script was giving full substring
+            supplierAddress: '0x9A232b2f5FbBf857d153Af8B85c16CBDB4Ffb667', // TODO need real supplier address on the project
```
jaycenhorton, line 94: fix before merging

tasks/migrate-removals.ts
```diff
@@ -0,0 +1,252 @@
...
+            subIdentifier: (project.projectId % 1000) + index, // TODO get from removal
```
jaycenhorton, line 95: fix before merging

tasks/migrate-removals.ts
```diff
@@ -0,0 +1,252 @@
...
+                project.projectId
+              } txHash: ${chalk.green(pendingTx.hash)}`
+            );
+          } catch (error) {
+            hre.log(
+              `(${projectIndex}/${jsonData.length}) projectId: ${
+                project.projectId
+              } ${chalk.red('error')}`
+            );
+            pendingTransactions.push(error);
```
jaycenhorton, line 126: should we throw instead to enforce order of on-chain removals?

tasks/migrate-removals.ts
```diff
@@ -0,0 +1,252 @@
...
+      }
+      if (!dryRun) {
+        hre.log(chalk.white('\n👷 Waiting for transactions to finalize...'));
+        const PROGRESS_BAR = new cliProgress.SingleBar(
+          {},
+          cliProgress.Presets.shades_classic
+        );
+        PROGRESS_BAR.start(pendingTransactions.length, 0);
+
+        hre.log(`\n`);
```
jaycenhorton, line 161:
```suggestion
```

jaycenhorton: add to previous or next log statement instead if wanted

tasks/migrate-removals.ts
```diff
@@ -0,0 +1,252 @@
...
+            });
+            PROGRESS_BAR.increment();
+            return {
+              txReceiptOrError: txResult,
+              tokenIds,
+            };
+          })
+        );
+        PROGRESS_BAR.stop();
+        hre.log(`\n`);
```
jaycenhorton, line 207: https://github.com/nori-dot-eco/contracts/pull/406/files#r999819746

tasks/migrate-removals.ts
```diff
@@ -0,0 +1,252 @@
...
+            }`
+          );
+
+          return {
+            ...project,
+            txReceiptOrError: txResult.txReceiptOrError,
+            tokenIds: txResult.tokenIds,
+          };
+        });
+        writeFileSync(outputFile, JSON.stringify(mintingResults));
```
jaycenhorton, line 234: writeJsonSync

jaycenhorton (changes requested):
see comments. Nearly there!

tasks/list-remaining-migrated-removals.ts
amiecorso: haha no idea... I noticed this happen another time in a different file - it might even have been a careless copilot inclusion

tasks/list-remaining-migrated-removals.ts
amiecorso: this one goes from ether to wei, but we want wei to ether

tasks/list-remaining-migrated-removals.ts
amiecorso: yeah - forgot to bring this script in line with the others.  it'll now write some information to a file and then crash and throw

tasks/list-remaining-migrated-removals.ts
amiecorso: the issue with deleting the if else blocks is that the return value from the dry run is void, so we don't actually get a pending transaction and would have to write a bunch of more granular logic to avoid performing the rest of the transaction awaiting, file writing, etc. that is different between the two scenarios

tasks/migrate-certificates.ts
amiecorso: damn, I think I forgot to push my changes from yesterday which does this! I will push whatever I have now

amiecorso: (and yes, we should! which the scripts now do)

tasks/migrate-removals.ts
amiecorso: yeah I just hadn't pushed the latest... this will throw and stop when any error occurs (and also still write the file with any successful txns up to that point and the error too)

tasks/list-remaining-migrated-removals.ts
amiecorso: so I'm adding validation to make sure the signer has the `CONSIGNOR` role and that the network is one of `localhost`, `mumbai`, or `polygon` .... I'm not sure how to handle checking that the signer is a fireblocks signer.  maybe you just mean that based on the network we should cast the signer as a `FireblocksSigner` ? I guess we could then check that the signer has one of the fireblocks-only properties?

tasks/migrate-removals.ts
amiecorso: definitely will do - leaving this here just for now since otherwise we're going to have to toggle this back and forth for testing with data that doesn't have this value (none of our prod data actually has wallet addresses for suppliers... we should probably handle testing that by modifying the input file or something though)

rlfarman: I think it's best to merge code as close to the intended state as possible, and anything that's convenience for the sake of future testing should just be commented out for the time being. So I'd suggest commenting this and the sub-identifier out, and replace them with `removal. supplierAddress` and `removal.subIdentifier`.
Then when testing locally we can uncomment those out and leave it like that locally.

rlfarman (approved):

tasks/list-remaining-migrated-removals.ts
jaycenhorton: there should be fireblocks-exclusive properties on the signer object, e.g [this one](https://github.com/nori-dot-eco/contracts/blob/master/plugins/fireblocks/fireblocks-signer.ts#L35)

tasks/migrate-removals.ts
```diff
@@ -0,0 +1,220 @@
...
+            if (txReceipt.status !== 1) {
+              // log an error that this transaction hash failed and we are exiting early
+              hre.log(
+                chalk.red(
+                  `❌ Transaction ${pendingTx.hash} failed with failure status ${txReceipt.status} - exiting early`
+                )
+              );
+              return;
+            }
+            // TODO remove this random sleep
```
jaycenhorton, line 149: why was it needed?

jaycenhorton (commented):
Looking good-- still a couple of open threads to address

tasks/list-remaining-migrated-removals.ts
scottynomad: Have we thought about which Fireblocks wallet should function as the consignor?  Can it be the contract owner or should we set up a new wallet address to function as the consignor?

tasks/migrate-certificates.ts
```diff
@@ -0,0 +1,202 @@
...
+        signer,
+      });
+      const certificateContract = await getCertificate({
+        hre,
+        signer,
+      });
+      hre.log(`Removal contract address: ${removalContract.address}`);
+      hre.log(`Certificate contract address: ${certificateContract.address}`);
+      hre.log(`Signer address: ${signerAddress}`);
+      // const fireblocksSigner = removalContract.signer as FireblocksSigner;
```
scottynomad, line 58: I'd suggest using a similar technique to this one:

https://github.com/nori-dot-eco/contracts/blob/master/plugins/index.ts#L96

That will let you post memos on the transactions that the people signing the requests  will see.  Assume that's the only reason to care that it's a fireblocks signer?

i.e.

```
    const fireblocksSigner = signer as FireblocksSigner;
    if (typeof fireblocksSigner.setNextTransactionMemo === 'function') {
      fireblocksSigner.setNextTransactionMemo(
        `Migrate legacy certificates`
      );
    }
```

scottynomad:
@rlfarman @amiecorso Am I reading / recalling correctly that we're going to see one transaction per certificate?  If so then I think I'm going to need to tweak the fireblocks settings a bit so that we're using an API user that has permission to self-sign all these requests.  It will be intractable to get humans to click for each existing certificate.

.gitignore
```diff
@@ -114,3 +114,10 @@ broadcast
 # interfaces
 contracts/interfaces/
 
+# migration (temporary)
```
jaycenhorton, line 117: reminder to remove before merging

PR #416, opened 2022-11-07
[NO-2316] H2 Audit fix: `RestrictedNORI` claimable tokens can be over-withdrawn

amiecorso:
https://www.notion.so/0xmacro/Nori-High-Severity-Issue-2-dde2faa496344a4fbd8192128678a1e6

We decided to remediate this by disabling transfers for `RestrictedNORI` for now.  With transfers disabled, calculating the claimable amounts for a given token holder becomes straightforward and is robust to changes in schedule amounts as tokens are continually deposited over the course of the schedule (including to multiple token holders on the same schedule, for instance, if multiple supplier addresses became encoded in removal ids that all belong to the same project and thus schedule).

### DIFF
--- a/test/RestrictedNORI.t.sol
+++ b/test/RestrictedNORI.t.sol
@@ -4,7 +4,8 @@ pragma solidity =0.8.17;
 import "@/test/helpers/restricted-nori.sol";
 import "@/test/checkout.int.t.sol";
 
-contract Macro_Market_linearReleaseAmountAvailableSingleHolder is
+/** @dev Test correct claimable balance after claiming claimable then transferring remaining tokens, single holder */
+contract RestrictedNORI_withdrawFromSchedule_claimableBalancePostTransferSingleHolder is
   UpgradeableMarket
 {
   uint256[] removalIds;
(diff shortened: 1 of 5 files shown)


test/RestrictedNORI.t.sol
```diff
@@ -4,6 +4,469 @@ pragma solidity =0.8.17;
...
+      scheduleStartTime: block.timestamp,
+      holdbackPercentage: 50
+    });
+
+    _rNori.grantRole(_rNori.MINTER_ROLE(), address(this));
+    _bpNori.grantRole(_bpNori.DEPOSITOR_ROLE(), address(this));
+    _bpNori.deposit(address(_rNori), abi.encode(1 ether));
+  }
+
+  function testClaimableAmountWithMultipleSuppliers() external {
```
amiecorso, line 414: this is the test that illustrates the issue that can't be solved with our approach

test/RestrictedNORI.test.ts
```diff
@@ -3,6 +3,7 @@ import { defaultRemovalTokenIdFixture } from './fixtures/removal';
 
 import { formatTokenAmount, formatTokenString } from '@/utils/units';
 import { FINNEY, Zero } from '@/constants/units';
```
amiecorso, line 5: all the diffs in this file are just fixing tests that previously relied on the use of `safeTransferFrom` as part of state setup.... now we have to actually mint and restrict removals on behalf of the additional accounts that are holders for a multi-holder schedule which was a bit more boilerplate

amiecorso: that and getting rid of the transfer tests - I wrote forge tests for testing the disabling of transfers

scottynomad (approved):
This looks reasonable to me!  I'd like to set up some time tomorrow for us to take a crack at reimplementing the transfer logic together.   I think I can see a path to doing it and I think in the process we can clean up the interface between `RestrictedNORI` and `RestrictedNORILib` which feel a bit awkward at the moment.

PR #418, opened 2022-11-09
[NO-2376] M4 Prevent consigning from `Certificate` contract

amiecorso:
https://www.notion.so/0xmacro/Nori-1-Preliminary-Audit-Report-external-b70f6c3e0aa843489a1aeefc2b9f4b9b#3bf2d2d7ed8f4ce0a0974df6ef970925

Note that I would have liked to put this inside `_beforeTokenTransfer` so that ALL transfers from the `Certificate` contract (except for burns which would have to be explicitly checked) are disabled at a more fundamental layer, but all transfers are already effectively disabled because `safeTransferFrom` and `safeBatchTransferFrom` are the only functions that allow transferring besides consign and those can only be called by the market on behalf of its own tokens.  Putting these checks inside `_beforeTokenTransfer` means that they will have to be made before all transfers and therefore incur additional gas during purchases, so I opted not to (which is also in line with the proposed remediation from Macro)

### DIFF
--- a/contracts/Errors.sol
+++ b/contracts/Errors.sol
@@ -124,3 +124,7 @@ error IncorrectSupplyAllocation();
  * @notice Thrown when the caller specifies the zero address for the Nori fee wallet.
  */
 error NoriFeeWalletZeroAddress();
+/**
+ * @notice Thrown when attempting to list for sale a removal that already belongs to the Certificate contract.
+ */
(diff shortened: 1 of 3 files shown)


scottynomad (approved):
👍  Definitely makes sense not to incur that extra gas every time.

PR #419, opened 2022-11-09
[NO-2377] M5 Add validation for holdback percentages

amiecorso:
https://www.notion.so/0xmacro/Nori-1-Preliminary-Audit-Report-external-b70f6c3e0aa843489a1aeefc2b9f4b9b#2386da557cd44f9891b2db7caad82797

A holdback percentage above 100 would have caused a `Market.swap` call to revert with an arithmetic underflow so we add validation to avoid invalid percentages while setting this during removal minting.


scottynomad (approved):
💥

PR #420, opened 2022-11-09
[NO-2378] L3 Move check for balances sum equaling certificate amount into certificate contract

amiecorso:
https://www.notion.so/0xmacro/Nori-1-Preliminary-Audit-Report-external-b70f6c3e0aa843489a1aeefc2b9f4b9b#64b82218c335490ba9beb109fbf78f03

### DIFF
--- a/contracts/Certificate.sol
+++ b/contracts/Certificate.sol
@@ -56,6 +56,8 @@ contract Certificate is
   MulticallUpgradeable,
   AccessPresetPausable
 {
+  using UInt256ArrayLib for uint256[];
+
   /**
    * @notice Role conferring operator permissions.
    * @dev Assigned to operators which are the only addresses which can transfer certificates outside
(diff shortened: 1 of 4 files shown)


contracts/Certificate.sol
```diff
@@ -360,11 +363,16 @@ contract Certificate is
...
+   * @param certificateAmount The total number of tonnes of carbon removals represented by the new certificate.
    */
   function _validateReceivedRemovalBatch(
     uint256[] calldata removalIds,
-    uint256[] calldata removalAmounts
+    uint256[] calldata removalAmounts,
+    uint256 certificateAmount
   ) internal pure {
+    if (removalAmounts.sum() != certificateAmount) {
+      revert("Incorrect supply allocation");
```
amiecorso, line 374: note that I am intentionally choosing to use a traditional revert string because since this function is called from within `onERC1155BatchReceived` we have to rely on the try/catch error handling that's part of that implementation, and that error handling doesn't handle custom errors.  So if we use a custom error we fall through the try catch and revert with the far less correct and informative `'ERC1155: transfer to non ERC1155Receiver implementer'`

scottynomad (approved):
🚀

PR #421, opened 2022-11-09
[NO-2379] L4 Fix `Certificate.supportsInterface` 

amiecorso:
https://www.notion.so/0xmacro/Nori-1-Preliminary-Audit-Report-external-b70f6c3e0aa843489a1aeefc2b9f4b9b#6a9c67215f9f48faa1735d0d32fef489

I can't say I totally understand why `supportsInterface` wasn't picking up those interface IDs, since they are part of a superclass `supportsInterface` -- I think it just comes down to understanding inheritance better and realizing that that particular implementation of `supportsInterface` just doesn't get called.

### DIFF
--- a/contracts/Certificate.sol
+++ b/contracts/Certificate.sol
@@ -234,7 +234,10 @@ contract Certificate is
     )
     returns (bool)
   {
-    return super.supportsInterface({interfaceId: interfaceId});
+    return
+      super.supportsInterface({interfaceId: interfaceId}) ||
+      interfaceId == 0x80ac58cd || // interface ID for ERC721
+      interfaceId == 0x5b5e139f; // interface ID for ERC721Metadata
   }
 
   /**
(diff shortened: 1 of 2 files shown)


scottynomad (approved):
I believe this is the relevant spot in the docs that explains why `super` ends up calling the `AccessControlEnumerableUpgradeable` implementation of `supportsInterface` (Doing a depth first search starting from the right side of the `is` list of the contract `AccessPausablePreset` is first and it inherits `AccessControlEnumerableUpgradeable`). 

https://docs.soliditylang.org/en/v0.8.12/contracts.html#multiple-inheritance-and-linearization

This is also a straightforward example: https://solidity-by-example.org/inheritance/

Also worth noting that in a multiple inheritance situation the the super call always boils down to calling *one* implementation back up the inheritance tree but not all of them which would have needed to happen for the original implementation to work since those three different implementation are in different branches of the inheritance tree.

PR #423, opened 2022-11-10
Q2 NO-2381 Removing cyclic dependencies in contract code.

scottynomad:
Implemented minimal interface stubs to eliminate the cyclic dependencies between contract implementations.  Market continues to depend on most of the other contracts directly but all the other contracts around it now no longer depend on Market or each other.

Confirmed that I was able to run `hardhat flatten` now.


contracts/Market.sol
```diff
@@ -93,7 +100,7 @@ contract Market is
   /**
    * @notice The BridgedPolygonNORI contract.
    */
-  BridgedPolygonNORI private _bridgedPolygonNORI;
+  IERC20WithPermit private _bridgedPolygonNORI;
```
scottynomad, line 103: Looking ahead to the possible USDC work I'm tempted to rename this variable here and in `RestrictedNORI`.

amiecorso (approved):
nice! thanks for taking this ticket I actually wasn't sure how to approach it.  love the rename.

jaycenhorton:
Any reason we didn't just use the whole of the interface instead of the minimal? e.g., the output of 

`cast interface artifacts/contracts/Market.sol/Market.json`

scottynomad:
> Any reason we didn't just use the whole of the interface instead of the minimal? e.g., the output of
> 
> `cast interface artifacts/contracts/Market.sol/Market.json`

No strong preference from my side.  Keeping them minimal makes explicit where the cross-contract deps are but I'm open to just generating the interfaces as well.  I think we'd eventually want to automate generating them as the contracts change.  I'm open to trying this approach to see how it shakes out.

jaycenhorton:
> > Any reason we didn't just use the whole of the interface instead of the minimal? e.g., the output of
> > `cast interface artifacts/contracts/Market.sol/Market.json`
> 
> No strong preference from my side. Keeping them minimal makes explicit where the cross-contract deps are but I'm open to just generating the interfaces as well. I think we'd eventually want to automate generating them as the contracts change. I'm open to trying this approach to see how it shakes out.

Sounds good. It would mean that if one contract uses an additional function in the future in the interface (e.g., need to append a new entry to the interface), you'd probably want to upgrade any other contract that currently imports the interface or you end up with a bit of a strange DX

jaycenhorton (approved):

jaycenhorton (approved):

scottynomad:
> > > Any reason we didn't just use the whole of the interface instead of the minimal? e.g., the output of
> > > `cast interface artifacts/contracts/Market.sol/Market.json`
> > 
> > 
> > No strong preference from my side. Keeping them minimal makes explicit where the cross-contract deps are but I'm open to just generating the interfaces as well. I think we'd eventually want to automate generating them as the contracts change. I'm open to trying this approach to see how it shakes out.
> 
> Sounds good. It would mean that if one contract uses an additional function in the future in the interface (e.g., need to append a new entry to the interface), you'd probably want to upgrade any other contract that currently imports the interface or you end up with a bit of a strange DX (even though other contracts might not use the appended function that was defined)

I tried this out but because cast slurps in a complete interface including all the event definitions and inherited methods from the OZ base classes it causes multiple inheritance issues and would require us to make much more extensive changes to the contracts to implement all the ambiguous methods with disambiguating override directives.

We can revisit this but I'm going to merge what we have for now.  It mights make sense for us to manually maintain more comprehensive interfaces covering only the unique functions from our contracts and ignoring the inherited stuff.

PR #424, opened 2022-11-10
NO-2385 [Q6] add overflow protection to `UInt256ArrayLib.sum`

jaycenhorton:
- Addresses [Q-6](https://www.notion.so/0xmacro/Nori-1-Preliminary-Audit-Report-external-b70f6c3e0aa843489a1aeefc2b9f4b9b#275fbacb0b224fa29cc0201f55fcb839). Adds overflow protection to `UInt256ArrayLib.sum`. This adds a bit of additional gas overhead, but still benchmarks well below the standard implementation (~half the cost).  Might be a more efficient way to do this in yul, but should suffice for now.
- fixes a bug/typo in the natspec and corresponding docs
- fixes a typo in the foundry toml config file

### TODO
- [x] generate gas snapshot (requires the fix in #423 that fixes the `Removal_release_unlisted_listed_and_retired` test)


test/ArrayLib.t.sol
```diff
@@ -100,14 +100,22 @@ contract UInt256ArrayLib_sum is Global {
...
+    vals.sum();
   }
 
   function test_library() external view {
     this.libraryImplementation();
   }
 
+  function test_reference() external view {
+    this.standardImplementation();
+  }
```
jaycenhorton, line 117: unchanged, just moved it so I could colocate the two test_library fns without the solium linter noise

amiecorso (approved):

PR #425, opened 2022-11-11
Fixes logic in `Removal.release` to avoid reentrancy risk

amiecorso:


contracts/Removal.sol
```diff
@@ -704,8 +704,8 @@ contract Removal is
    * @param amount The amount to burn.
    */
   function _releaseFromMarket(uint256 id, uint256 amount) internal {
-    _market.release(id, amount);
     super._burn({from: this.marketAddress(), id: id, amount: amount});
+    _market.release(id, amount);
```
amiecorso, line 708: correct order to avoid reentrancy risk pattern

contracts/Market.sol
```diff
@@ -300,7 +300,7 @@ contract Market is
       account: address(this),
       id: removalId
     });
-    if (amount == removalBalance) {
+    if (removalBalance == 0) {
```
amiecorso, line 303: removal gets removed

jaycenhorton (approved):

PR #426, opened 2022-11-11
[NO-2374] [M2][L2] Make `LockedNORI` grants and wrapped `bpNORI` 1:1 to avoid overlocking or insolvency

amiecorso:
https://www.notion.so/0xmacro/Nori-1-Preliminary-Audit-Report-external-b70f6c3e0aa843489a1aeefc2b9f4b9b#86e5043d03eb4962b344daf29cc8dedb

Addressing issue M2 will also take care of L2, since making the `lNORI` and `bpNORI` relationship 1:1 in `LockedNORI` will avoid both insolvency in `LockedNORI` and also over-locking `bpNORI`.

Got rid of `createGrant` which allows for creation of a grant without wrapping underlying `bpNORI`, and got rid of `depositFor` which is no longer required because no grants can exist without wrapped tokens.  All grant creation now has to rely on `batchCreateGrants`.  Fixing tests was most of the work in this PR


contracts/test/LockedNORIV1Helper.sol
```diff
@@ -0,0 +1,148 @@
+// SPDX-License-Identifier: MIT
```
amiecorso, line 1: this file was renamed from LockedNORIHelper.sol to LockedNORIV1Helper.sol because now the two versions of the contracts differ in their interfaces to the point that V1 locked nori needs its own helper (this helper contract is used in the upgrade test).

`LockedNORIHelper.sol` is now up to date with the new version of `LockedNORI.sol` so it doesn't use`createGrant` and can instead create grants using `batchCreateGrants`

scottynomad (approved):

test/LockedNORI.test.ts
```diff
@@ -178,9 +177,20 @@ const setupWithGrant = async (
     },
   } as TokenGrantOptions;
 
-  await expect(
-    lNori.createGrant(
-      grantAmount,
+  const userData = ethers.utils.defaultAbiCoder.encode(
```
amiecorso, line 180: i'm seeing that I should probably create a utility for this grant creation using `batchCreateGrants` because it happens like 4 times in here

contracts/LockedNORI.sol
```diff
@@ -263,7 +243,20 @@ contract LockedNORI is ERC777PresetPausablePermissioned {
   }
 
   /**
-   * @notice Batch version of `createGrant` with permit support.
+   * @notice Sets up a batch of vesting + lockup schedules and wraps the corresponding BridgedPolygonNORI
+   * using EIP712 permit access.
```
jaycenhorton, line 247:
```suggestion
   * @notice Create a batch of vesting + lockup schedules.
   * @dev This function uses EIP712 to transfer and wrap bpNORI as lNORI with the specified grant parameters.
```

contracts/test/LockedNORIV1Helper.sol
jaycenhorton: do we still need tests for v1?

contracts/LockedNORI.sol
```diff
@@ -263,7 +243,20 @@ contract LockedNORI is ERC777PresetPausablePermissioned {
...
+   *
+   * ##### Requirements:
+   *
+   * - Can only be used when the contract is not paused.
+   * - Can only be used when the caller has the `TOKEN_GRANTER_ROLE` role.
+   *
+   * @param amounts The amount of each grant
```
jaycenhorton, line 254:
```suggestion
   * @param amounts The amount of bpNORI for each grant.
```

contracts/LockedNORI.sol
```diff
@@ -263,7 +243,20 @@ contract LockedNORI is ERC777PresetPausablePermissioned {
...
   /**
-   * @notice Batch version of `createGrant` with permit support.
+   * @notice Sets up a batch of vesting + lockup schedules and wraps the corresponding BridgedPolygonNORI
+   * using EIP712 permit access.
+   *
+   * ##### Requirements:
+   *
+   * - Can only be used when the contract is not paused.
+   * - Can only be used when the caller has the `TOKEN_GRANTER_ROLE` role.
+   *
```
jaycenhorton, line 253:
```suggestion
   *
```

contracts/LockedNORI.sol
```diff
@@ -263,7 +243,20 @@ contract LockedNORI is ERC777PresetPausablePermissioned {
...
+   * @param amounts The amount of each grant
+   * @param grantParams An array of `CreateTokenGrantParams` structs
+   * @param deadline The permit deadline
+   * @param v The permit v parameter
+   * @param r The permit r parameter
+   * @param s The permit s parameter
```
jaycenhorton, line 259:
```suggestion
   * @param grantParams An array of `CreateTokenGrantParams` structs
   * @param deadline The permit deadline.
   * @param v The permit v parameter.
   * @param r The permit r parameter.
   * @param s The permit s parameter.
```

These could honestly use better descriptions too, but considering the consumers of this contract, it's probably not worth the effort.

contracts/LockedNORI.sol
```diff
@@ -659,7 +613,7 @@ contract LockedNORI is ERC777PresetPausablePermissioned {
    * - The contract must not be paused.
    * - The recipient cannot be the zero address (e.g., no burning of tokens is allowed).
    * - One of the following must be true:
-   *    - The operation is minting (which should ONLY occur when BridgedPolygonNORI is being wrapped via `_depositFor`)
+   *    - The operation is minting (which should ONLY occur when BridgedPolygonNORI is being wrapped during `batchCreateGrants`)
```
jaycenhorton, line 616:
```suggestion
   *    - The operation is minting (which should ONLY occur when BridgedPolygonNORI is being wrapped during the execution of `batchCreateGrants`).
```

test/LockedNORIUpgrade.test.ts
```diff
@@ -5,20 +5,20 @@ import type {
...
+    const { namedSigners, namedAccounts } = hre;
+    const recipient = namedAccounts.investor1;
     const GRANT_AMOUNT = ethers.utils.parseEther('100');
 
     const helperFactory = await hre.ethers.getContractFactory(
-      'LockedNORIHelper'
+      'LockedNORIV1Helper'
     );
-    const helper = (await helperFactory.deploy()) as LockedNORIHelper;
+    const helper = (await helperFactory.deploy()) as LockedNORIV1Helper;
```
jaycenhorton, line 21:
```suggestion
    const helper = await helperFactory.deploy();
```

test/LockedNORIUpgrade.test.ts
```diff
@@ -5,20 +5,20 @@ import type {
...
     const { bpNori, hre } = await setupTest();
-
-    const recipient = hre.namedAccounts.investor1;
+    const { namedSigners, namedAccounts } = hre;
+    const recipient = namedAccounts.investor1;
     const GRANT_AMOUNT = ethers.utils.parseEther('100');
 
     const helperFactory = await hre.ethers.getContractFactory(
-      'LockedNORIHelper'
+      'LockedNORIV1Helper'
```
jaycenhorton, line 19: Use the generic instead to avoid casting in following lines (e.g., `const helper = `)
```ts
const helperFactory =
      await hre.ethers.getContractFactory<LockedNORIV1Helper__factory>(
        'LockedNORIHelper'
      );
```

jaycenhorton (approved):

jaycenhorton:
Can you update the gas snapshot as well?

contracts/test/LockedNORIV1Helper.sol
amiecorso: this is used in the upgrade test - otherwise the tests apply to V2

scottynomad: The upgrade test (and the V1 contract) could be removed now that the upgrade has happened but let's wait until the audit changes are done.

PR #428, opened 2022-11-15
add logger and migration types; make dryrun DRYer

jaycenhorton:
- adds types for the migration data to make it easier to reason about what's actually in the JSON
- adds a reusable logger type
- makes the certificate migration task DRYer
- fixes a couple of misc. bugs primarily dealing with loggers trying to log undefined values (certificate.id, and the total in the dry original run variant)


tasks/migrate-certificates.ts
```diff
@@ -106,104 +114,82 @@ export const GET_MIGRATE_CERTIFICATES_TASK = () =>
...
+            `❌ Error minting certificate ${
+              JSON.parse(certificate.key).id
+            } (number ${certificateIndex}/${jsonData.length}) - exiting early`
+          );
+          hre.log(error);
+          outputData.push({
+            ...certificate,
+            error,
+          });
+          writeJsonSync(outputFileName, outputData);
```
jaycenhorton, line 181: @amiecorso do you recall if/why this second write is needed since we're not throwing? Could the error just be initialized pre-loop and then conditionally added to the final write call?

amiecorso:
btw, the PR description on [this PR](https://github.com/nori-dot-eco/nori/pull/2924) contains the step-by-step instructions for running through the whole migration process end-to-end on a local chain with our migration dress rehearsal project

PR #432, opened 2022-11-16
[NO-2375] [M3] Specify removal ID in `RestrictedNORI.revokeUnreleasedTokens`

amiecorso:
https://www.notion.so/0xmacro/Nori-1-Preliminary-Audit-Report-external-b70f6c3e0aa843489a1aeefc2b9f4b9b#524f34f87a8e4537a6befdb33ee2007f

See [note in jira ticket](https://norinauts.atlassian.net/browse/NO-2375) for description of our decision to take this approach.

In short, adding the removal ID to the revocation event will allow for cross-referencing of events and will make it visible to the public whether any tokens that Nori is revoking from a supplier were indeed marked as released in the market contract system.


contracts/RestrictedNORI.sol
```diff
@@ -295,15 +298,16 @@ contract RestrictedNORI is
...
    * @param amount The amount to revoke.
    * @param toAccount The account to which the underlying ERC20 token should be sent.
    */
   function revokeUnreleasedTokens(
-    uint256 projectId,
+    uint256 removalId,
     uint256 amount,
     address toAccount
   ) external whenNotPaused onlyRole(TOKEN_REVOKER_ROLE) {
+    uint256 projectId = _removal.getProjectId(removalId);
```
jaycenhorton, line 310:
```suggestion
    uint256 projectId = _removal.getProjectId({id: removalId});
```

jaycenhorton (approved):
left one nit-picky comment, otherwise lgtm

contracts.json
```diff
@@ -4,7 +4,7 @@
       "proxyAddress": "0xEdf0bef28943aE9d2ee7d84cF1AEEC85588655b7"
     },
     "BridgedPolygonNORI": {
-      "proxyAddress": "0x5D4Ccd1c007cAe21347F3ceD346389b66f7a0ccB"
+      "proxyAddress": "0xb4c857d57b7130F033159958B5af4Fa3Bd0a04c6"
```
scottynomad, line 7: Can you leave this churn in the contracts.json out of the PR?  Does anyone know why this changes all the time?  Maybe the hardhat version should not be checked in at all?

contracts/RestrictedNORI.sol
```diff
@@ -270,12 +272,13 @@ contract RestrictedNORI is
...
   /**
-   * @notice Revokes amount of tokens from the specified project (schedule) ID and transfers to `toAccount`.
+   * @notice Revokes `amount` of tokens from the project (schedule) associated with the specificed
+   * `removalId` and transfers them to `toAccount`.
    * @dev The behavior of this function can be used in two specific ways:
    * 1. To revoke a specific number of tokens as specified by the `amount` parameter.
    * 2. To revoke all remaining revokable tokens in a schedule by specifying 0 as the `amount`.
    *
-   * Transfers any unreleased tokens in the specified schedule and reduces the total supply
+   * Transfers unreleased tokens in the removal's schedule and reduces the total supply
```
scottynomad, line 281: project's schedule?

scottynomad (approved):

contracts/RestrictedNORI.sol
amiecorso: going with "removal's project's"

contracts.json
amiecorso: i do not know why it changes all the time esp because I just confirmed that the contract address is independent of the bytecode (just based on the sending account and the current nonce for that account).  not sure that the hardhat version would affect this either.  seems to happen when I spin up a local hardhat node and i'm often excluding `contracts.json` from my commits

PR #433, opened 2022-11-16
[NO-2389] Gas optimizations from audit 

amiecorso:
https://www.notion.so/0xmacro/Nori-1-Preliminary-Audit-Report-external-b70f6c3e0aa843489a1aeefc2b9f4b9b#032a6002e36448eba91d3643b5abf534


test/Removal.t.sol
```diff
@@ -1332,30 +1332,6 @@ contract Removal_safeTransferFrom_reverts_ForbiddenTransfer is
...
-    _removalIds = _seedRemovals({
-      to: _namedAccounts.supplier,
-      count: 1,
-      list: false
-    });
-    _removal.pause();
-  }
-
-  function test() external {
-    vm.expectRevert("Pausable: paused");
```
amiecorso, line 1348: I remove this test because we can't actually test it with `expectRevert` anymore now that we are relying on `_beforeTokenTransfer` to enforce the pause on the contract.  This is because `expectRevert` expects the *next* call to revert, but since the next immediate call no longer reverts and instead waits until `_beforeTokenTransfer` gets called to revert, `vm.expectRevert` doesn't register the revert.  This test is covered by the test for pausability on `_beforeTokenTransfer` directly

jaycenhorton: bummer. Probably should be rewritten to instead have this test just expect that  safeTransferFrom invokes whenNotPaused (no idea what that might look like though I'd imagine there's a cheat code for checking which state was accessed)

jaycenhorton (approved):
Would have loved to see each of these changes broken out into smaller PRs to see how each change impacted gas individually instead of as a whole. Hard to see how much each affected totals

PR #438, opened 2022-11-21
[NO-2406] Add `Removal.updateHoldbackPercentage`

amiecorso:
Adds the ability to update the holdback percentage for a project after removals have been minted.  Emits corresponding event.

### DIFF
--- a/contracts/Removal.sol
+++ b/contracts/Removal.sol
@@ -318,7 +318,10 @@ contract Removal is
       removals: removals,
       projectId: projectId
     });
-    _setHoldbackPercentage(projectId, holdbackPercentage);
+    _setHoldbackPercentage({
+      projectId: projectId,
+      holdbackPercentage: holdbackPercentage
+    });
     _mintBatch({to: to, ids: ids, amounts: amounts, data: ""});
     IRestrictedNORI _restrictedNORI = IRestrictedNORI(
       _market.restrictedNoriAddress()
--- a/contracts/Removal.sol
+++ b/contracts/Removal.sol
@@ -156,6 +156,13 @@ contract Removal is
   event ContractAddressesRegistered(IMarket market, ICertificate certificate);
 
   /**
+   * @notice Emitted when the holdback percentage is updated for a project.
+   * @param projectId The ID of the project.
+   * @param holdbackPercentage The new holdback percentage for the project.
+   */
+  event HoldbackPercentageUpdated(uint256 projectId, uint8 holdbackPercentage);
+
+  /**
    * @notice Emitted on releasing a removal from a supplier, the market, or a certificate.
    * @param id The id of the removal that was released.
    * @param fromAddress The address the removal was released from.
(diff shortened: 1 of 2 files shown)


contracts/Removal.sol
```diff
@@ -155,6 +155,13 @@ contract Removal is
    */
   event ContractAddressesRegistered(IMarket market, ICertificate certificate);
 
+  /**
+   * @notice Emitted when the holdback percentage is updated for a project.
+   * @param projectId The ID of the project.
+   * @param holdbackPercentage The new holdback percentage for the project.
+   */
+  event HoldbackPercentageUpdated(uint256 projectId, uint8 holdbackPercentage);
+
```
amiecorso, line 164: do you think we should also emit this event when the initial holdback percentage is set during `mintBatch` ?

jaycenhorton: Nit: some of our events use past tense for event names, yet all of our inherited events use active tense as well as some of our own events. Id opt out for all events to be active tense since we can't change underlying event names. 

`UpdateHoldbackPercentage`

jaycenhorton: Yes to your comment, but that should happen automatically if the initializer calls the function

contracts/Removal.sol
```diff
@@ -236,6 +243,33 @@ contract Removal is
     });
   }
 
+  /**
+   * @notice Updates the holdback percentage value for a project.
```
jaycenhorton, line 247:
```suggestion
   * @notice Update the holdback percentage value for a project.
```

contracts/Removal.sol
```diff
@@ -236,6 +243,33 @@ contract Removal is
...
+   * @dev Emits a `HoldbackPercentageUpdated` event.
+   * ##### Requirements:
```
jaycenhorton, line 249:
```suggestion
   *
   * ##### Requirements:
```

Nit: missing line break

test/Removal.t.sol
```diff
@@ -1458,3 +1458,72 @@ contract Removal_getOwnedTokenIds is UpgradeableMarket {
     }
   }
 }
+
+contract Removal_updateHoldbackPercentage is UpgradeableMarket {
+  event HoldbackPercentageUpdated(uint256 projectId, uint8 holdbackPercentage);
+
+  uint256 projectId = 1234567890;
```
jaycenhorton, line 1465: Can use immutable priv types for these to improve gas accuracy in reports

test/Removal.t.sol
```diff
@@ -1458,3 +1458,72 @@ contract Removal_getOwnedTokenIds is UpgradeableMarket {
...
+    });
+  }
+
+  function test() external {
+    uint8 newHoldbackPercentage = 20;
+    assertEq(
+      _removal.getHoldbackPercentage(removalTokenId),
+      originalHoldbackPercentage
+    );
+    vm.expectEmit(true, true, false, false);
```
jaycenhorton, line 1498: You can write a higher value test for events using this pattern

https://github.com/nori-dot-eco/contracts/blob/master/test/Removal.t.sol#L1060

contracts/Removal.sol
```diff
@@ -299,12 +318,7 @@ contract Removal is
       removals: removals,
       projectId: projectId
     });
-    if (holdbackPercentage > 100) {
-      revert InvalidHoldbackPercentage({
-        holdbackPercentage: holdbackPercentage
-      });
-    }
-    _projectIdToHoldbackPercentage[projectId] = holdbackPercentage;
+    _setHoldbackPercentage(projectId, holdbackPercentage);
```
jaycenhorton, line 321: Use named params

jaycenhorton (approved):
Left one suggestion to use named params, otherwise lgtm

(Sorry for the nit-- easier to maintain pattern if we don't have to track them down later)

contracts/Removal.sol
amiecorso: no worries this is a good and worthy nit!

PR #440, opened 2022-11-22
[NO-2405] Allow internal setting of a `_priceMultiple` state variable, getter

amiecorso:
Establishes a state variable in the `Market` contract, `_priceMultiple`, which can later be used as a multiple of the underlying base token to determine how many of that token are required to purchase one NRT.

The setter here is internal and so can only be called by this contract.  An event is emitted with the updated price multiple information. Also creates a getter to view the current price multiple.

In a followup PR, a public setter will be exposed that enforces atomic setting of the underlying token and the price multiplier.

### DIFF
--- a/contracts/Market.sol
+++ b/contracts/Market.sol
@@ -358,7 +358,7 @@ contract Market is
   }
 
   /**
-   * @notice Sets the price multiple, which is the number of base tokens required to purchase one NRT.
+   * @notice Set the price multiple, which is the number of base tokens required to purchase one NRT.
    * @param priceMultiple The new price multiple.
    */
   function _setPriceMultiple(uint256 priceMultiple) internal {
@@ -660,7 +660,7 @@ contract Market is
   }
 
   /**
-   * @notice Returns the current value of the price multiple, which is the number of base tokens required to
+   * @notice Get the current value of the price multiple, which is the number of base tokens required to
    * purchase one NRT.
    */
   function getPriceMultiple() external view returns (uint256) {
--- a/contracts/Market.sol
+++ b/contracts/Market.sol
@@ -108,6 +108,11 @@ contract Market is
   RestrictedNORI private _restrictedNORI;
 
   /**
+   * @notice The number of base tokens required to purchase one NRT.
+   */
+  uint256 private _priceMultiple;
+
+  /**
    * @notice Wallet address used for Nori's transaction fees.
    */
   address private _noriFeeWallet;
@@ -154,6 +159,12 @@ contract Market is
    * @param threshold The updated threshold for priority restricted supply.
    */
   event PriorityRestrictedThresholdSet(uint256 threshold);
+
+  /**
+   * @notice Emitted on setting of `_priceMultiple`.
+   * @param priceMultiple The updated price multiple.
+   */
+  event SetPriceMultiple(uint256 priceMultiple);
 
   /**
    * @notice Emitted on updating the addresses for contracts.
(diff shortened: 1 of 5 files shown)


contracts/Market.sol
```diff
@@ -155,6 +160,12 @@ contract Market is
    */
   event PriorityRestrictedThresholdSet(uint256 threshold);
 
+  /**
+   * @notice Emitted on setting of `_priceMultiple`.
+   * @param priceMultiple The updated price multiple.
+   */
+  event SetPriceMultiple(uint256 priceMultiple);
```
scottynomad, line 167: PriceMultipleSet?

deploy/deploy-market.ts
```diff
@@ -20,6 +20,7 @@ export const deploy: DeployFunction = async (environment) => {
     hre,
     feeWallet,
     feePercentage: 15,
+    priceMultiple: 1,
```
scottynomad, line 23: I think we need to have control down to at least a two decimal places.

scottynomad (changes requested):

contracts/Market.sol
```diff
@@ -343,6 +357,15 @@ contract Market is
     });
   }
 
+  /**
+   * @notice Sets the price multiple, which is the number of base tokens required to purchase one NRT.
```
jaycenhorton, line 361: > 
```suggestion
   * @notice Set the price multiple, which is the number of base tokens required to purchase one NRT.
```

contracts/Market.sol
```diff
@@ -636,6 +659,14 @@ contract Market is
     }
   }
 
+  /**
+   * @notice Returns the current value of the price multiple, which is the number of base tokens required to
```
jaycenhorton, line 663:
```suggestion
   * @notice Get the current value of the price multiple, which is the number of base tokens required to
```

test/Market.t.sol
```diff
@@ -861,3 +862,21 @@ contract Market_getRemovalIdsForSupplier is UpgradeableMarket {
...
 }
+
+contract Market__setPriceMultiple is NonUpgradeableMarket {
+  function test() external {
+    vm.recordLogs();
+    uint256 newPriceMultiple = 20;
+    _setPriceMultiple({priceMultiple: newPriceMultiple});
+    Vm.Log[] memory entries = vm.getRecordedLogs();
+    assertEq(entries.length, 1);
+    assertEq(entries[0].topics[0], keccak256("SetPriceMultiple(uint256)"));
```
jaycenhorton, line 873: 💯

deploy/deploy-market.ts
amiecorso: meaning we want to be able to set a price like $20.34 / NRT? 

to achieve this I guess we just need our price multiple to be scaled by 100 such that 1 is represented by 100 and 20.34 would be represented as 2034 and we divide by 100 after doing the calculations, and just make it very clear how this works in the documentation.  sound right?

contracts/Market.sol
amiecorso: so Jaycen had asked me to change the tense of another event to match the active voice used in all the OZ dependencies (for example, `TransferBatch`, and now we have `SetHoldbackPercentage`)... what do you guys think?  I realize that it is not consistent with some of our other event names and personally I find it less intuitive in the active voice than in the past tense, but i'm fine with whatever approach

scottynomad: Hmm, yeah I would rather have our names be internally consistent since these will be with us forever but not a big deal.

scottynomad (approved):

contracts/Market.sol
jaycenhorton: Fwiw active voice seems to be the standard in most popular contracts. Same goes for documentation

PR #442, opened 2022-11-23
[NO-2418] Enforce atomic updating of purchasing token and price multiple

amiecorso:
Public setter for purchasing token and price multiple, requires both variables be set at once, tests.


scottynomad (approved):
🎆

PR #435, opened 2022-11-17
[Q7] [NO-2386] fix uint array slicing

jaycenhorton:
Our slice implementation was completely dysfunctional and was not actually ever slicing anything and was instead just overwriting values. This addresses Q7 and implements the proper behavior for array slicing.

PR #443, opened 2022-11-23
[NO-2407] Incorporate price multiple into market purchases

amiecorso:
This PR incorporates the price multiple (and arbitrary purchasing token - though no changes are needed here for that) into purchases and adds this value as well as the purchasing token contract address to the `ReceiveRemovalBatch` event emitted upon certificate creation.
- existing tests are updated to be robust to an arbitrary price multiple, without assuming that it is 1
- existing tests are updated to account for the additional parameters in a `ReceiveRemovalBatch` event
- the default price multiple used in deployments of the market contract is set to 2000 (i.e. 20) which is what we'll want to use for USDC.... I'm not sure if we want to do this here?
- new tests for testing purchases after having changed the purchasing token and price multiple to an arbitrary ERC20, also tests with a price multiple that acts as a floating point (i.e. $19.95)


contracts/Removal.sol
```diff
@@ -433,7 +433,7 @@ contract Removal is
       to: address(_certificate),
       ids: ids,
       amounts: amounts,
-      data: abi.encode(certificateRecipient, certificateAmount)
+      data: abi.encode(certificateRecipient, certificateAmount, address(0), 0)
```
amiecorso, line 436: since this is the `migrate` function, there is no purchasing token or price multiple... but we have to pass this as data that will be decoded by `Certificate.onERC1155BatchReceived`.... what do we think about 0 address and 0 price multiplier as defaults?

deploy/deploy-market.ts
```diff
@@ -20,7 +20,7 @@ export const deploy: DeployFunction = async (environment) => {
     hre,
     feeWallet,
     feePercentage: 15,
-    priceMultiple: 100,
+    priceMultiple: 2000,
```
amiecorso, line 23: this sets the price multiple to 20.  I don't think it hurts to set that as a default now that we're preparing to use USDC as the purchasing token.  this way tests are using a non-1 multiple value too

test/Market.test.ts
```diff
@@ -343,7 +343,6 @@ describe('Market', () => {
           hre,
           totalAmountOfSupply,
```
amiecorso, line 344: the changes in this file are simply updating existing tests to rely on using the contract logic for calculations and thus abstracting away the impact of the price multiple from hard-coded values in tests -- something that we'd been meaning to do to all our tests anyway

test/Certificate.t.sol
```diff
@@ -77,7 +77,7 @@ contract Certificate_burn is UpgradeableCertificate {
       address(0),
       _removalIds2,
       _removalAmounts,
-      abi.encode(address(this), 1 ether) // todo use diff. named account
+      abi.encode(address(this), 1 ether, address(0), 100) // todo use diff. named account
```
amiecorso, line 80: just updating existing tests with an arbitrary filler value for these new pieces of data -- this is not an integration test

contracts/Certificate.sol
```diff
@@ -335,7 +345,9 @@ contract Certificate is
       certificateId: certificateId,
       certificateAmount: certificateAmount,
       removalIds: removalIds,
-      removalAmounts: removalAmounts
+      removalAmounts: removalAmounts,
+      purchasingTokenAddress: purchasingTokenAddress,
```
jaycenhorton, line 349: I'd also be curious how much additional gas overhead was added due to the number of arguments and the non-tighty packed nature of these arguments

contracts/Market.sol
```diff
@@ -735,7 +741,7 @@ contract Market is
    * @return The amount of the fee for Nori.
    */
   function calculateNoriFee(uint256 amount) external view returns (uint256) {
-    return (amount * _noriFeePercentage) / 100;
+    return (amount * _priceMultiple * _noriFeePercentage) / 100 / 100;
```
jaycenhorton, line 744: Can we use muldiv here to optimize gas?

https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/utils/math/Math.sol#L55

contracts/Market.sol
```diff
@@ -749,21 +755,24 @@ contract Market is
...
    */
   function calculateCertificateAmountFromPurchaseTotal(uint256 purchaseTotal)
     external
     view
     returns (uint256)
   {
-    return (purchaseTotal * 100) / (100 + _noriFeePercentage);
+    uint256 priceWithoutFee = (purchaseTotal * 100) /
+      (100 + _noriFeePercentage);
+    return (priceWithoutFee * 100) / _priceMultiple;
```
jaycenhorton, line 775: Can we use muldiv here?

https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/utils/math/Math.sol#L55

contracts/Market.sol
jaycenhorton: Also, I find it a bit confusing that price multiple is in n*10^2 but the fee percentage is not. That might just be me though so happy to leave this unaddressed if so.

contracts/Market.sol
```diff
@@ -749,21 +755,24 @@ contract Market is
     view
     returns (uint256)
   {
-    return amount + this.calculateNoriFee({amount: amount});
+    return
+      (amount * _priceMultiple) / 100 + this.calculateNoriFee({amount: amount});
```
jaycenhorton, line 759: Use muldiv

jaycenhorton, line 759:
```suggestion
      ((amount * _priceMultiple) / 100) + this.calculateNoriFee({amount: amount});
```
Although not needed, I find the above notation easier to read

contracts/Market.sol
```diff
@@ -749,21 +755,24 @@ contract Market is
...
   }
 
   /**
-   * @notice Calculates the quantity of carbon removals being purchased given the purchase total and the
-   * percentage of that purchase total that is due to Nori as a transaction fee.
-   * @param purchaseTotal The total amount of Nori used for a purchase.
+   * @notice Calculates the quantity of carbon removals being purchased given the purchase total, the price multiple,
+   * and the percentage of that purchase total that is due to Nori as a transaction fee.
+   * @param purchaseTotal The total amount of ERC20 tokens used for a purchase.
```
jaycenhorton, line 765:
```suggestion
   * @param purchaseTotal The total number of `_purchasingToken`s used for a purchase.
```

contracts/Market.sol
jaycenhorton: Also in the line above it

contracts/Market.sol
```diff
@@ -860,7 +869,7 @@ contract Market is
 
   /**
    * @notice Fulfill an order.
-   * @dev This function is responsible for paying suppliers, routeing tokens to the RestrictedNORI contract, paying Nori
+   * @dev This function is responsible for paying suppliers, routing tokens to the RestrictedNORI contract, paying Nori
```
jaycenhorton, line 872: Nice catch

contracts/Market.sol
```diff
@@ -891,7 +900,7 @@ contract Market is
     uint256 restrictedSupplierFee;
     uint256 unrestrictedSupplierFee;
     for (uint256 i = 0; i < countOfRemovalsAllocated; ++i) {
-      unrestrictedSupplierFee = removalAmounts[i];
+      unrestrictedSupplierFee = (removalAmounts[i] * _priceMultiple) / 100;
```
jaycenhorton, line 903: Use muldiv

jaycenhorton: Also wonder if there is value in making this an external view function

contracts/Market.sol
```diff
@@ -927,7 +936,12 @@ contract Market is
         amount: unrestrictedSupplierFee
       });
     }
-    bytes memory data = abi.encode(recipient, certificateAmount);
+    bytes memory data = abi.encode(
+      recipient,
+      certificateAmount,
+      address(_purchasingToken),
```
jaycenhorton, line 942: Sort of wonder if there's a way to detect this without passing it through (either by parsing out this information from transaction data since the token contract address and amount will be included in the originating transaction) or some other way (msg.sender? tx.origin?-- I think both of these are invalid for different reasons but that's the initial direction my mind is drawing me towards).

contracts/Market.sol
```diff
@@ -927,7 +936,12 @@ contract Market is
...
+      _priceMultiple
+    );
```
jaycenhorton, line 944: What's this slither alert 👇?

contracts/Removal.sol
jaycenhorton: Sounds reasonable, though check my other comments about whether we even need these values at all

contracts/Market.sol
jaycenhorton: Also check the slither warning below

contracts/Market.sol
jaycenhorton: Also, is it possible to instead pass a struct as the data argument somehow? I vaguely recall some solidity magic that let me do this before, but I'm not sure what was required to pull it off

contracts/test/MockERC20Permit.sol
```diff
@@ -1,9 +1,9 @@
 // SPDX-License-Identifier: MIT
 pragma solidity ^0.8.17;
-
+import "../IERC20WithPermit.sol";
```
jaycenhorton, line 3: Doesn't OZ have an interface contract for permit tokens? Must've missed where this interface was added to the codebase

test/checkout.int.t.sol
```diff
@@ -361,3 +368,139 @@ contract Checkout_buyingFromTenSuppliers is Checkout {
...
+    );
+    _assertExpectedBalances(address(_market), 0, false, 0);
+    _assertExpectedBalances(_namedAccounts.supplier, 0, false, 0);
+    _assertExpectedBalances(address(_certificate), certificateAmount, true, 1);
+    assertEq(
+      _removal.balanceOf(address(_certificate), _removalIds[0]),
+      certificateAmount
+    );
+    assertEq(_certificate.ownerOf(_certificateTokenId), owner);
+    vm.stopPrank();
```
jaycenhorton, line 435: Would be useful to also test the new event you added here (or elsewhere)

test/checkout.int.t.sol
```diff
@@ -361,3 +368,139 @@ contract Checkout_buyingFromTenSuppliers is Checkout {
...
+      owner,
+      amount,
+      signedPermit.permit.deadline,
+      signedPermit.v,
+      signedPermit.r,
+      signedPermit.s
+    );
+    _assertExpectedBalances(address(_market), 0, false, 0);
+    _assertExpectedBalances(_namedAccounts.supplier, 0, false, 0);
+    _assertExpectedBalances(address(_certificate), certificateAmount, true, 1);
```
jaycenhorton, line 498: useful to also test nori fee wallet and buyer wallet for erc20 balances but maybe that's already happening in this fn?

test/Market.test.ts
```diff
@@ -1028,28 +1026,34 @@ describe('Market', () => {
...
-          purchaseAmount.sub(
-            formatTokenAmount(removalAmount)
-              .mul(project1HoldbackPercentage)
-              .div(100)
-              .add(
-                formatTokenAmount(removalAmount)
-                  .mul(project2HoldbackPercentage)
-                  .div(100)
-              )
+        supplierInitialNoriBalance
```
jaycenhorton, line 1029: Statements like these are prime candidates to check if there's an issue with our big number/floating point usage we discovered on a call with @rlfarman a week or so ago

test/Market.t.sol
```diff
@@ -108,11 +108,12 @@ contract Market_swap_rNori_mint_failure is UpgradeableMarket {
...
     );
-
+    uint256 numberOfNRTsToPurchase = 1 ether;
     uint256 ownerPrivateKey = 0xA11CE;
     owner = vm.addr(ownerPrivateKey);
-    checkoutTotal = _market.calculateCheckoutTotal(1 ether);
-    rNoriToMint = (1 ether * holdbackPercentage) / 100;
+    checkoutTotal = _market.calculateCheckoutTotal(numberOfNRTsToPurchase);
+    uint256 noriFeeAmount = _market.calculateNoriFee(numberOfNRTsToPurchase);
+    rNoriToMint = ((checkoutTotal - noriFeeAmount) * holdbackPercentage) / 100;
```
jaycenhorton, line 116: Might be useful to add an external view function for this logic (I believe the market contract is also internally doing this calculation making it easy for this to become decoupled later-- some DRY code would prevent that)

jaycenhorton (changes requested):
Can you check in the gas snapshot? If you run it locally can you also print the % change total from `yarn snapshot --diff`

Also see online comments

contracts/test/MockERC20Permit.sol
scottynomad: I added it during the removal of circular deps. They do have this [draft-IERC20PermitUpgradeable.sol](https://github.com/OpenZeppelin/openzeppelin-contracts-upgradeable/blob/master/contracts/token/ERC20/extensions/draft-IERC20PermitUpgradeable.sol) Which has the Permit extensions but not the rest of the ERC20 spec.

contracts/Market.sol
scottynomad: They're both represented the same way here by my read -- in hundredths of the utility token.

scottynomad (approved):

contracts/test/MockERC20Permit.sol
jaycenhorton: Makes sense then!

contracts/Certificate.sol
```diff
@@ -91,14 +91,18 @@ contract Certificate is
...
   event ReceiveRemovalBatch(
     address from,
     address indexed recipient,
     uint256 indexed certificateId,
     uint256 certificateAmount,
     uint256[] removalIds,
-    uint256[] removalAmounts
+    uint256[] removalAmounts,
+    address purchasingTokenAddress,
+    uint256 priceMultiple
```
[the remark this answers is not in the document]
amiecorso: so, very weirdly, I cannot seem to generate a new snapshot for this branch.  no errors, all the tests run when I run `forge snapshot`, but there are no differences in the generated .gas-snapshot .... going to have to figure this out, let me know if you have any ideas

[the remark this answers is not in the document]
amiecorso: oh, this is because there was a failing test (which I believe has been failing since[ this PR](https://github.com/nori-dot-eco/contracts/pull/435)) and when I comment out that test the snapshot runs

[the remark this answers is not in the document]
amiecorso: so I tried packing the event args more tightly by moving the addresses to be all together, which means those three args can fit in two words, and the gas actually went up slightly from the original snapshot.  the difference was super tiny.  In what sense can you see the purchasing token in the originating transaction?  there are permit sig components but nothing from which the purchasing token contract address can be derived.  I suppose that contract is going to emit `TransferFrom` events that we could comb through....

I'm fine with removing the purchasing token contract address if you feel confident that there's a straightforward path to reconstructing this information in whatever analytics we will need to build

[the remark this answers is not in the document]
amiecorso: @scott any opinions?

contracts/Certificate.sol
amiecorso: parameter ordering doesn't matter in terms of packing and gas: [Here](https://docs.soliditylang.org/en/v0.8.17/internals/layout_in_storage.html#storage-inplace-encoding) it says "When dealing with function arguments or memory values, there is no inherent benefit because the compiler does not pack these values."

contracts/Market.sol
amiecorso: I think it does make sense because both of them are technically scaled by 100.  For example, a fee percentage of 15% is the value .15 which we represent as 15.  If the price multiple were also .15 (which I presume it never would be because that's less than one, but just for example) it would be represented as 15.  Anything that needs decimal precision has to be scaled by the number of digits of precision needed, and both of these need two digits and are scaled by 100.

contracts/Certificate.sol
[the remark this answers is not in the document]
scottynomad: @amiecorso I'd prefer to keep it.  This will be a critical part of reports built on the market transaction flow and the additional headache of trying to correlate this event to all possible source tokens doesn't seem worth it.

contracts/Market.sol
amiecorso: moot after using muldiv

contracts/Market.sol
amiecorso: you mean a function for `unrestrictedSupplierFee` ?  I'm not sure how much sense that would make out of context or where else we'd want to use it

contracts/Market.sol
amiecorso: I think you may be mixing up our old approach of using ERC777 tokens where the token contract address would be present in the originating transaction but that is no longer the case with the permit mechanism

contracts/Market.sol
amiecorso: ah yeah another divide before multiply... I think I can take care of that

contracts/Removal.sol
amiecorso: yeah maybe scott can weigh in on this but I think we need this because we don't actually have the purchasing token contract address in the originating transaction because we do not use an ERC777 anymore -- it's just opaque permit sig args

test/Market.t.sol
amiecorso: a function like this only makes sense on a removal-specific basis - it doesn't actually apply at the purchase level because individual removals that are used to fulfill a purchase could have different holdback percentages.  The only reason we use this line of logic in the test is because we know that all the removals have the same holdback percentage and we're generating a value to be able to expect against. 

I don't think we can make an external view function that actually makes sense / is wroth it on its own.

test/Market.test.ts
amiecorso: i'm not sure there's anything that interesting here because none of these values are floating point numbers, which was the source of that bug.  we have a ticket in this sprint to audit all of those instances so i'll look at them then.

contracts/Certificate.sol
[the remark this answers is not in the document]
amiecorso: agreed.  I also realized in some later comments that I think jaycen might be thinking of the old ERC777 system which would indeed have very obvious information about the purchasing token because the originating transaction would interact with that contract.  In this case, the only link we'd have to the ERC20 would be the transfer events it would emit as part of that transaction and i'm not even sure those would have the token address.

contracts/Market.sol
amiecorso: oops answered this before refreshing and seeing your comments scott. sounds like we're on the same page -- not trying to overexplain you jaycen haha

test/checkout.int.t.sol
amiecorso: done

test/checkout.int.t.sol
amiecorso: done

PR #444, opened 2022-11-30
[NO-2448] Prevent ERC20 transfer from `Market` to `RestrictedNORI` that are not its wrapped token

amiecorso:
See the[ jira ticket ](https://norinauts.atlassian.net/browse/NO-2448) for more details on the issue we're trying to solve for.

The approach implemented here is to simply check whether the ERC20 that would be deposited into `RestrictedNORI` is the one that `RestrictedNORI` is configured to wrap (i.e. its `underlyingToken`).  If the ERC20 is NOT the one `RestrictedNORI` expects, we emit an event (`RestrictedNORIERC20TransferSkipped`) and instead that balance of ERC20 goes straight to the supplier instead of being restricted.

The idea being that this scenario is an error case that should never arise, and if it does arise we don't want to crash purchases because of it.  Nori takes the hit by foregoing the restriction schedule on the insurance tokens.  Presumably, if the ERC20 was the wrong one, it never should have been sent to `RestrictedNORI` in the first place and so preventing that is the correct behavior (and will be much easier to clean up after if that is needed).

Note that we'll want to have sentinels set up for the occurrence of this event as well as the restricted nori mint failure event so that we are notified if they ever happen, which indicates that something is misconfigured and some action needs to be taken.


scottynomad:
The implementation looks fine but I'm concerned about the complexity this adds to gracefully handle an error that should never happen.

The accounting and reporting implications of this actually happening would get quite involved.

I'd prefer if this just reverted and we monitor for that error but I obviously missed the conversation that led to this work so take my opinion with a grain of salt!

scottynomad (approved):

PR #447, opened 2022-12-02
[NO-2445] Standardize naming of events and getters

amiecorso:
Uses imperative tense for events and prefixes getters with "get" (excludes `LockedNORI`)


rlfarman (approved):
FWIW a cursory glance at [Microsoft's design guidelines](https://learn.microsoft.com/en-us/dotnet/standard/design-guidelines/general-naming-conventions) says that you should include either the present or past tense for your events (IE - `Closed` or `Closing`). [Other results](https://www.google.com/search?q=naming+conventions+events&rlz=1C1ONGR_enUS953US953&oq=naming+conventions+events&aqs=chrome..69i57.2191j0j1&sourceid=chrome&ie=UTF-8) generally seem to suggest following an Object->Action pattern.

I don't have a strong opinion about this other than we should try to be consistent, and I think this PR helps in that effort. But we might want to consider changing these again to follow recommended guidelines for event names.

PR #453, opened 2022-12-08
[NO-2269] add migration validation and summary 

jaycenhorton:
Using a sample: 
image
image

### DIFF
--- a/tasks/migrate-certificates.ts
+++ b/tasks/migrate-certificates.ts
@@ -1,15 +1,16 @@
 /* eslint-disable no-await-in-loop -- need to submit transactions synchronously to avoid nonce collisions */
 import cliProgress from 'cli-progress';
 import { task, types } from 'hardhat/config';
+import type { ContractReceipt } from 'ethers';
 import { BigNumber, FixedNumber } from 'ethers';
 import { readJsonSync, writeJsonSync } from 'fs-extra';
 import type { TransactionReceipt } from '@ethersproject/providers';
 
-import type { Certificate } from '../typechain-types';
+import type { Certificate, Removal } from '../typechain-types';
 
 import { getLogger } from '@/utils/log';
 import { parseTransactionLogs } from '@/utils/events';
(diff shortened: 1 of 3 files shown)


tasks/migrate-certificates.ts
```diff
@@ -30,12 +33,310 @@ interface MigratedCertificate {
...
+    throw new Error(
+      `Removal amounts do not match for certificate ${eventLog.certificateId}. Expected: ${datastoreCertificate.amounts} , Got: ${eventLog.removalAmounts}`
+    );
+  }
+  if (!offChainIdsMatchOnchainIds) {
+    throw new Error(
+      `Removal ids do not match for certificate ${eventLog.certificateId}. Expected: ${datastoreCertificate.ids} , Got: ${eventLog.removalIds}`
+    );
+  }
+  if (eventLog.certificateAmount !== datastoreCertificate.data.gramsOfNrts) {
```
amiecorso, line 150: wouldn't the certificateAmount be represented here as tonnes of carbon, in wei, which shouldn't equal `gramsOfNrts` ?

jaycenhorton: on-chain values are represented as tonnes in wei (5 tonnes === 5 10**18). Above this ([here](https://github.com/nori-dot-eco/contracts/blob/jaycen-migration-validation-and-summary/tasks/migrate-certificates.ts#L127)) line we are formatting on-chain values as grams  so that these two are the same

tasks/migrate-certificates.ts
```diff
@@ -30,12 +33,310 @@ interface MigratedCertificate {
...
+}): void => {
+  validateMigrateEvent({
+    txResult,
+    removalContract,
+    hre,
+    certificateIndex,
+    inputData,
+    recipient,
+  });
+  // Removal.TransferBatch(
```
rlfarman, line 191: RM commented code?

rlfarman (approved):
LGTM, glad we have some extra assurance at these critical steps

PR #456, opened 2022-12-10
[NO-2515] [M6] Correct selection of next supplier in round robin

amiecorso:
Our code was written such that if a removal being used up was a supplier's *last* removal, the current supplier address was incremented twice: once while the last removal was being removed inside `_removeActiveRemoval` > `_removeActiveSupplier` (since the supplier itself is removed from the queue and we need to point at the next one) and then again inside `_allocateSupply`.

Suggested remediation: Consider to only call _incrementCurrentSupplierAddress() if the current supplier hasn’t been already incremented in the _removeActiveRemoval function.

This is implemented most simply by checking whether the call to `_removeActiveRemoval` led to a change in the state variable `_currentSupplierAddress`.  If it *did*, we do not increment again.

Tests check that the next active supplier is being correctly selected in cases where a removal is used up and is a supplier's last removal, and cases where a removal is used up but is not that supplier's last removal.

### DIFF
--- a/test/Market.t.sol
+++ b/test/Market.t.sol
@@ -1148,7 +1148,7 @@ contract Market_setPurchasingTokenAndPriceMultiple_revertsIfNotAdmin is
   }
 }
 
-contract Market_supplier_selection_using_up_suppliers_last_removal is
+contract Market_supplierSelectionUsingUpSuppliersLastRemoval is
   MarketBalanceTestHelper
 {
   address owner;
@@ -1220,7 +1220,7 @@ contract Market_supplier_selection_using_up_suppliers_last_removal is
   }
 }
 
-contract Market_supplier_selection_not_using_up_suppliers_last_removal is
+contract MarketSupplierSelectionNotUsingUpSuppliersLastRemoval is
   MarketBalanceTestHelper
 {
   address owner;
--- a/test/Market.t.sol
+++ b/test/Market.t.sol
@@ -1147,3 +1147,152 @@ contract Market_setPurchasingTokenAndPriceMultiple_revertsIfNotAdmin is
     _market.setPurchasingTokenAndPriceMultiple(IERC20WithPermit(address(0)), 0);
   }
 }
+
+contract Market_supplier_selection_using_up_suppliers_last_removal is
+  MarketBalanceTestHelper
+{
+  address owner;
+  uint256 checkoutTotal;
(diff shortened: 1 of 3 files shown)


test/Market.t.sol
```diff
@@ -1147,3 +1147,152 @@ contract Market_setPurchasingTokenAndPriceMultiple_revertsIfNotAdmin is
     _market.setPurchasingTokenAndPriceMultiple(IERC20WithPermit(address(0)), 0);
   }
 }
+
+contract Market_supplier_selection_using_up_suppliers_last_removal is
```
jaycenhorton, line 1151: nit:
```suggestion
contract Market_supplierSelectionUsingUpSuppliersLastRemoval is
```

test/Market.t.sol
```diff
@@ -1147,3 +1147,152 @@ contract Market_setPurchasingTokenAndPriceMultiple_revertsIfNotAdmin is
...
+    // Nothing of 3rd supplier was taken
+    uint256 supplier3RemovalBalance = _removal.balanceOf(
+      address(_market),
+      removalIds[2]
+    );
+    assertEq(1 ether, supplier3RemovalBalance);
+  }
+}
+
+contract Market_supplier_selection_not_using_up_suppliers_last_removal is
```
jaycenhorton, line 1223: nit:
```suggestion
contract MarketSupplierSelectionNotUsingUpSuppliersLastRemoval is
```

jaycenhorton (approved):

test/Removal.test.ts
```diff
@@ -8,6 +9,19 @@ import { Zero, AddressZero } from '@/constants/units';
...
+  it('should return false for isBignumberish floating point', () => {
+    expect(isBigNumberish(1.1)).to.be.false;
+    expect(() => BigNumber.from(1.1)).throws();
+    expect(() => BigNumber.from('1.1')).throws();
+
+    const nrts = BigNumber.from(5050).div(1000);
+    console.log({ nrts: nrts.toString() });
+    // const isThisZero = BigNumber.from(null);
+    // console.log({ isThisZero: isThisZero.toString() });
+    // expect(isThisZero.isZero()).to.be.true;
```
scottynomad, line 22: rm

scottynomad (commented):
:+1:

PR #457, opened 2022-12-10
[NO-2517] H1 Followup to remove unused param in `Market.release()`

amiecorso:


jaycenhorton (changes requested):
Regen docs; otherwise lgtm

PR #459, opened 2022-12-14
Add unpermissioned approval-based swap endpoint

amiecorso:
NO-2598

Proposal for adding a public (unpermissioned) approval-based purchasing endpoint, `swapByApproval`
- some renaming to make the various endpoints more clear
- tests
- omitted creation of a `swapFromSupplierByApproval` because we're not clear on that use case yet but maybe someday we will want that

[NO-2598]: https://norinauts.atlassian.net/browse/NO-2598?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


scottynomad (commented):
Would it be possible to do one of these approaches @amiecorso ?

* Have the permit version call the approval version after doing the `_permit` check since the code is otherwise identical.  Would probably then call the approval version `swap` to minimize the tortured naming creeping in.
* -or- Stick with a single `swap` function and treat the permit related params as optional.  If they're all non-null then you'd call the `_permit` check otherwise ignore it and move on.

I would be in favor of taking the same approach for the per supplier version now since we're changing our standard interface to support permit or plain approval it would be nice to be consistent about it.

scottynomad:
@amiecorso Can you remind me what the context was for needing this?  Was this related to the relayer interaction?

I'm ok with merging this but I would like us to rethink the proliferation of swap options now exposed that probably don't all need their own methods.
* Permit Y/N?
* Custom Fee?
* Specific Supplier?

contracts/Market.sol
```diff
@@ -539,6 +539,7 @@ contract Market is
    * ##### Requirements:
    *
    * - Can only be used when this contract is not paused.
+   *
```
jaycenhorton, line 542:
```suggestion
```

contracts/Market.sol
```diff
@@ -575,6 +576,45 @@ contract Market is
...
+   * supported ERC20 to complete the purchase. A certificate is minted in the Certificate contract
+   * to the specified recipient and the ERC20 is distributed to the suppliers of the carbon removals,
+   * to the RestrictedNORI contract that controls any restricted tokens owed to the suppliers, and finally
+   * to Nori Inc. as a market operator fee.
+   *
+   * ##### Requirements:
+   *
+   * - Can only be used when this contract is not paused.
+   * - Can only be used if this contract has been granted approval to transfer the sender's ERC20.
+   *
```
jaycenhorton, line 593:
```suggestion
```

contracts/Market.sol
```diff
@@ -632,10 +672,59 @@ contract Market is
...
+   * the purchase. A certificate is issued by the Certificate contract to the specified recipient and the ERC20 is
+   * distributed to the supplier of the carbon removal, to the RestrictedNORI contract that controls any restricted
+   * ERC20 owed to the supplier, and finally to Nori Inc. as a market operator fee.
+   *
+   *
+   * ##### Requirements:
+   *
+   * - Can only be used when this contract is not paused.
+   * - Can only be used if this contract has been granted approval to transfer the sender's ERC20.
+   *
```
jaycenhorton, line 691:
```suggestion
```

contracts/Market.sol
```diff
@@ -575,6 +576,45 @@ contract Market is
     });
   }
 
+  /**
+   * @notice Exchange ERC20 tokens for an ERC721 certificate by transferring ownership of the removals to the
+   * certificate. Relies on pre-approval of this market by the sender to transfer the sender's tokens.
+   * @dev See [here](https://docs.openzeppelin.com/contracts/4.x/api/token/erc20#IERC20-approve-address-uint256-) for more.
+   * The message sender must have granted approval to this contract to authorize this market to transfer the sender's
```
jaycenhorton, line 583:
```suggestion
   * The sender must have granted approval to this contract to authorize this market to transfer the sender's
```

contracts/Market.sol
```diff
@@ -575,6 +576,45 @@ contract Market is
...
+   * supported ERC20 to complete the purchase. A certificate is minted in the Certificate contract
+   * to the specified recipient and the ERC20 is distributed to the suppliers of the carbon removals,
```
jaycenhorton, line 585:
```suggestion
   * to the specified recipient and the ERC20 tokens are distributed to the supplier(s) of the carbon removals,
```

contracts/Market.sol
```diff
@@ -575,6 +576,45 @@ contract Market is
...
+   * to the RestrictedNORI contract that controls any restricted tokens owed to the suppliers, and finally
+   * to Nori Inc. as a market operator fee.
+   *
+   * ##### Requirements:
+   *
+   * - Can only be used when this contract is not paused.
+   * - Can only be used if this contract has been granted approval to transfer the sender's ERC20.
```
jaycenhorton, line 592:
```suggestion
   * - Can only be used if this contract has been granted approval to transfer the sender's ERC20 tokens.
```

contracts/Market.sol
```diff
@@ -632,10 +672,59 @@ contract Market is
     });
   }
 
+  /**
+   * @notice An overloaded version of `swap` that additionally accepts a supplier address and will exchange
+   * ERC20 tokens for an ERC721 certificate token and transfers ownership of removal tokens supplied only
+   * from the specified supplier to that certificate. If the specified supplier does not have enough carbon removals
+   * for sale to fulfill the order the transaction will revert.
+   * @dev See [here](https://docs.openzeppelin.com/contracts/4.x/api/token/erc20#IERC20-approve-address-uint256-) for more.
```
jaycenhorton, line 680:
```suggestion
   * @dev See [here](https://docs.openzeppelin.com/contracts/4.x/api/token/erc20#IERC20-approve-address-uint256-)
   * for more.
```
line length exceeds 120

contracts/Market.sol
```diff
@@ -632,10 +672,59 @@ contract Market is
...
+   * The message sender must have already granted approval to this market to transfer the sender's ERC20 to complete
```
jaycenhorton, line 681:
```suggestion
   * The sender must have already granted approval to this contract in order to transfer the sender's ERC20 tokens to complete
```

contracts/Market.sol
```diff
@@ -632,10 +672,59 @@ contract Market is
...
+   * the purchase. A certificate is issued by the Certificate contract to the specified recipient and the ERC20 is
```
jaycenhorton, line 682:
```suggestion
   * the purchase. A certificate is issued by the Certificate contract to the specified recipient and the ERC20 tokens are
```

contracts/Market.sol
```diff
@@ -632,10 +672,59 @@ contract Market is
...
+   * distributed to the supplier of the carbon removal, to the RestrictedNORI contract that controls any restricted
+   * ERC20 owed to the supplier, and finally to Nori Inc. as a market operator fee.
```
jaycenhorton, line 684:
```suggestion
   * ERC20 tokens owed to the supplier, and finally to Nori Inc. as a market operator fee.
```

contracts/Market.sol
```diff
@@ -632,10 +672,59 @@ contract Market is
...
+   *
+   *
+   * ##### Requirements:
+   *
+   * - Can only be used when this contract is not paused.
+   * - Can only be used if this contract has been granted approval to transfer the sender's ERC20.
```
jaycenhorton, line 690:
```suggestion
   * - Can only be used if this contract has been granted approval to transfer the sender's ERC20 tokens.
```

contracts/Market.sol
```diff
@@ -678,7 +767,7 @@ contract Market is
   }
 
   /**
-   * @notice An overloaded version of `swap` that additionally accepts a supplier address and will exchange supported ERC20
+   * @notice An overloaded version of `swapWithoutFee` that additionally accepts a supplier address and will exchange supported ERC20
```
jaycenhorton, line 768: line length > 120

contracts/Market.sol
```diff
@@ -528,7 +528,7 @@ contract Market is
 
   /**
    * @notice Exchange ERC20 tokens for an ERC721 certificate by transferring ownership of the removals to the
-   * certificate.
+   * certificate. Relies on the EIP-2612 permit extension to facilitate ERC20 token transfer.
    * @dev See [ERC20Permit](https://docs.openzeppelin.com/contracts/4.x/api/token/erc20#ERC20Permit) for more.
```
jaycenhorton, line 532: line length > 120

contracts/Market.sol
```diff
@@ -632,10 +672,59 @@ contract Market is
...
+   * @notice An overloaded version of `swap` that additionally accepts a supplier address and will exchange
+   * ERC20 tokens for an ERC721 certificate token and transfers ownership of removal tokens supplied only
+   * from the specified supplier to that certificate. If the specified supplier does not have enough carbon removals
+   * for sale to fulfill the order the transaction will revert.
+   * @dev See [here](https://docs.openzeppelin.com/contracts/4.x/api/token/erc20#IERC20-approve-address-uint256-) for more.
+   * The message sender must have already granted approval to this market to transfer the sender's ERC20 to complete
+   * the purchase. A certificate is issued by the Certificate contract to the specified recipient and the ERC20 is
+   * distributed to the supplier of the carbon removal, to the RestrictedNORI contract that controls any restricted
+   * ERC20 owed to the supplier, and finally to Nori Inc. as a market operator fee.
+   *
```
jaycenhorton, line 684:
```suggestion
```

contracts/Market.sol
```diff
@@ -575,6 +576,45 @@ contract Market is
     });
   }
 
+  /**
+   * @notice Exchange ERC20 tokens for an ERC721 certificate by transferring ownership of the removals to the
+   * certificate. Relies on pre-approval of this market by the sender to transfer the sender's tokens.
+   * @dev See [here](https://docs.openzeppelin.com/contracts/4.x/api/token/erc20#IERC20-approve-address-uint256-) for more.
```
jaycenhorton, line 581: line length > 120

contracts/Market.sol
jaycenhorton: also line length > 120

contracts/Market.sol
```diff
@@ -575,6 +576,45 @@ contract Market is
...
+   * The message sender must have granted approval to this contract to authorize this market to transfer the sender's
+   * supported ERC20 to complete the purchase. A certificate is minted in the Certificate contract
```
jaycenhorton, line 583: line length > 120

jaycenhorton (approved):

jaycenhorton:
make sure to check in a gas report!

scottynomad:
> @scottynomad So the purpose of introducing this was to offer a public (unpermissioned) endpoint that uses pre-approval instead of of per-purchase-permits so that other potential partners will have the same ability that our relayer needed to get around the risk of race conditions during the creation of permits. Race conditions can occur in the scenario where transactions that require permit creation happen more quickly than a given transaction finally settles and they (like us inside Defender) don't have the ability to force the serialization of each atomic transaction process (permit creation + actual transaction finishing).
> 
> I took your advice and made the permit params optional -- and the way to do that in solidity is by overloading the function. So now our names are back to being simple. I also wanted to improve code reuse and refactor the permit version of `swap` to call the approval based version after calling `_permit` and using the permit. The issue that arises with this is kinda interesting... when `swap` (approval version) is called from `swap` (permit version), `msgSender()` is actually now the _Market_ contract, as the most recent initiator of a call, not the original sender of the transaction. So it fails with an "insufficient balance" error because the market does not itself hold any ERC20 to transfer. There's another value that always has the value of the true originator of a transaction, and that's `tx.origin`. However, this is typically considered bad practice to use in solidity because it can actually be manipulated by an attacker (this makes no sense to me but apparently when you submit a transaction you can actually set this value to any address you want, so it can be dangerous to use for authorization purposes or whatever because it can be spoofed). So `tx.origin` is usually considered bad practice to use in solidity... however for our purposes I don't actually think there's a security vulnerability because `tx.origin` is just being used to determine whose money to spend, and that money can't be spent unless that address has actually granted approval or permit to the market in a secure way.
> 
> So I think our options are:
> 
> 1. don't refactor the code and have two very similar swap/swapFromSupplier functions
>    or
> 2. go ahead and use tx.origin
>    or
> 3. add a parameter to the approval version of swap which is the transaction originator and/or basically whose money to spend (bit cumbersome and seems unnecessary - least favorite option)
> 
> What do you think?

Interesting.  The overload naming is a good improvement for now.  I think the best refactor here would be to have the two variations of the public `swap` function both call an internal `_swap` that takes the sender as a parameter.  This is very similar to how the OZ token contracts are structured.

PR #461, opened 2022-12-16
Removing tenderly as a dependency.

scottynomad:


mrdave-dev:
What's weird is I'm not seeing `tslog` in the lockfiles anywhere. I think we want to do this anyway so it's worth a shot

mrdave-dev (approved):

scottynomad:
@mrdave-dev Huh, yeah I ran `yarn why tslog` after the changes and didn't see it. Assumed it had fallen out as transitive to tenderly...

PR #482, opened 2023-01-11
add event typings to support

jaycenhorton:
- adds contract event typings + improved event parser to support [NO-2317]

[NO-2317]: https://norinauts.atlassian.net/browse/NO-2317?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

Prior to this PR, `parseTransactionLogs` would return event logs that were deeply types as any. This PR adds type support so that events are **automatically** typed based on the contract type you pass the funciton. 

Examples (note the intellisense hover info):
image
image
image
image


utils/events.ts
```diff
@@ -1,13 +1,38 @@
...
+import { ethers } from 'ethers';
 
-export const parseTransactionLogs = ({
+import type {
+  ContractEventInterface,
+  ContractWithEvents,
+} from '../types/events';
+
+/**
+ * Take a contract instance and a transaction receipt, and
```
amiecorso, line 9: missing the rest of this description

utils/events.ts
```diff
@@ -1,13 +1,38 @@
...
+ *
+ * @param args.contractInstance The contract instance to parse logs from.
+ * @param args.txReceipt The transaction receipt to parse logs from.
+ * @returns An array of strongly typed logs with type support based on `args.contractInstance`. Note that all events in
+ * the logs array which use `args.values` in their event signature will have `args.values` remapped to `args.vals` due
+ * to a {@link https://github.com/ethers-io/ethers.js/discussions/3542#discussioncomment-4214097| limitation} of
+ * JavaScript.
```
amiecorso, line 16: nice - i've run into this weirdness before

amiecorso (approved):
nice, just one incomplete documentation but not much else to say this is awesome!

PR #483, opened 2023-01-11
NO-2626 update fireblocks libs, which have transitive dependencies on jsonwebtoken

aqualen:


rlfarman (changes requested):
The nature of maintaining a submodule with its own lockfile as a workspace package means that when we make changes to dependencies in the submodule, we need to run both `yarn install` at the root of the monorepo and `yarn install` in the submodule, so both the monorepo will update its lockfile with the correct dependencies, as well as the lockfile in the submodule. So I'd expect to see an updated lockfile attached to this PR. I've opened a PR against this branch that does this [here](https://github.com/nori-dot-eco/contracts/pull/484)

aqualen:
Thanks @rlfarman I think I'm actually going to abandon this PR (which was part of trying to update all direct dependencies to pick up the transitive update to jsonwebtoken >=-v9.0.0...but that didn't work out as hoped).  Using a top-level resolution resulted in a clean dependency tree afaict.

PR #496, opened 2023-01-19
[NO-2653] Prototype replacement mechanism for released removals

amiecorso:
This PR introduces a new method to the `Market.sol` contract:
```
  function replace(
    address treasury,
    uint256 certificateId,
    uint256 totalAmountToReplace,
    uint256[] memory removalIdsBeingReplaced,
    uint256[] memory removalAmountsBeingReplaced
  ) external whenNotPaused onlyRole(MARKET_ADMIN_ROLE) 
```

This permissioned method purchases new removals from the market and sends them to the certificate contract on behalf of a certificate that has had some portion of its original removals released.  Because the release process for a removal doesn't specifically indicate a certain affected certificate or set of certificates, determination of which certificates have been affected by release needs to occur off-chain based on examining `CreateCertificate` events for the relationship between certificates and removals.  The certificate that is being updated therefore needs to be specified (as an argument to this function) so that this token id can become part of an `UpdateCertificate` event along with removal-related accounting information.

There are some refactors and renames associated with the changes in this PR so that everything makes more sense given the added complexity.

TODO:
- look into all these code scanner warnings...
- final docgen and forge snapshot


contracts/Market.sol
```diff
@@ -1081,94 +1149,126 @@
...
+          amount: this.calculateNoriFee(removalAmounts[i])
+        });
+      }
       _purchasingToken.transferFrom({
         from: from,
         to: suppliers[i],
         amount: unrestrictedSupplierFee
       });
     }
+  }
```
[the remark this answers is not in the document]
amiecorso: I believe that this is the only legit security issue coming up in the code scanning.  We actually touched on this issue in a meeting but need to follow up with another conversation about how to handle the security issue here.  Basically, the issue is that if one address approves the market to spend its funds, ANY other address could originate a purchase that spends that address's funds, but directs the recipient of the certificate to itself the originator or an address of its choosing.

This is fine with our relayer endpoint because it's permissioned and only intended to be used by an address that we control.  However, with an unpermissioned endpoint, this is an issue that we need to figure out how to address.  It's not a HUGE security concern because the only thing that someone can do with the approved funds is buy carbon removals, which the approval would have been intended for anyway, but it's still an issue for the integrity of our market.

amiecorso:
@scottynomad I left a comment on this slither warning identifying that we need to decide how to approach this security vulnerability with our decoupled spender/msgSender in the approval setting.

contracts/Market.sol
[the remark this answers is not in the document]
scottynomad: My recollection was that we were going to disallow an arbitrary payer on the unpermissioned non-pemit swap.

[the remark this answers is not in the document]
amiecorso: gotcha.  yeah I think I got confused because we talked about potentially having some sort of allowlist permissioning around that but I realize now it would be too involved to implement.  Opened a PR to address this security issue and revert the approval based endpoint to how it was before.  Not sure why I didn't catch a slither warning on the PR that introduced it, but at least we're catching it here.
https://github.com/nori-dot-eco/contracts/pull/500

contracts/Certificate.sol
```diff
@@ -206,22 +241,32 @@ contract Certificate is
...
-      recipient: recipient,
-      certificateAmount: certificateAmount,
-      removalIds: removalIds,
-      removalAmounts: removalAmounts,
-      purchasingTokenAddress: purchasingTokenAddress,
-      priceMultiple: priceMultiple
-    });
+    bool isReplacement = abi.decode(data, (bool));
+    if (isReplacement) {
+      // TODO should we be bothering to encode the replacement amount here, or just sum it from the removalAmounts?
```
rlfarman, line 249: Probably safest/least redundant to sum from removalAmounts.

contracts/Market.sol
```diff
@@ -345,6 +365,95 @@ contract Market is
...
+    address treasury,
+    uint256 certificateId,
+    uint256 totalAmountToReplace,
+    uint256[] memory removalIdsBeingReplaced,
+    uint256[] memory amountsBeingReplaced
+  ) external whenNotPaused onlyRole(MARKET_ADMIN_ROLE) {
+    if (_certificate.getPurchaseAmount(certificateId) == 0) {
+      revert CertificateNotYetMinted({tokenId: certificateId});
+    }
+    int256 _guaranteeDiscrepancy = _certificate.getGuaranteeDiscrepancy();
```
rlfarman, line 398: Update usage of `_guaranteeDiscrepancy` based on my comments in your other PR.
https://github.com/nori-dot-eco/contracts/pull/497#pullrequestreview-1266121711

contracts/Certificate.sol
```diff
@@ -206,22 +241,32 @@ contract Certificate is
     bytes calldata data
   ) external returns (bytes4) {
     if (_msgSender() != address(_removal)) {
+      // TODO use a require + string instead of a custom error because custom can't bubble - falls through a try catch
```
rlfarman, line 244: Fix this TODO before merging?

contracts/Market.sol
rlfarman, line 398: This should be updated in accordance to the changes in your other PR.
https://github.com/nori-dot-eco/contracts/pull/497#pullrequestreview-1266121711

contracts/Market.sol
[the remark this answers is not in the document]
amiecorso: fixing this in a separate ticket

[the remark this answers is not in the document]
amiecorso: fixing this in a separate ticket

contracts/Certificate.sol
```diff
@@ -103,7 +123,7 @@ contract Certificate is
    * @param purchasingTokenAddress The address of the token used to purchase the certificate.
    * @param priceMultiple The number of purchasing tokens required to buy one NRT.
    */
-  event ReceiveRemovalBatch(
+  event CreateCertificate(
```
rlfarman, line 126: Don't have a strong opinion on renaming this but we should be careful since @jaycenhorton has some changes in his branch that would break CI (and make it difficult to resolve) if this was merged.
Not saying we shouldn't do it, just make sure to coordinate with him on this.

rlfarman (approved):

contracts/Certificate.sol
```diff
@@ -61,6 +61,26 @@ contract Certificate is
...
+   * @param purchasingTokenAddress The address is the address of the token that was used to purchase the certificate.
+   * @param priceMultiple The number of purchasing tokens required to purchase one NRT.
+   */
+  struct CertificateData {
+    bool isReplacement;
+    address recipient;
+    uint256 certificateAmount;
+    address purchasingTokenAddress;
+    uint256 priceMultiple;
+  }
```
jaycenhorton, line 82: can probably use a smaller uint here which would allow for tighter packing in the struct. Currently at 840 bits I believe (3.2x 256 bit words). 

```suggestion
  struct CertificateData {
    uint256 certificateAmount; // 256
    address purchasingTokenAddress; // 160
    bool isReplacement; // 8
    uint88 priceMultiple;  // 88 (this + prev 2 == 256)
    address recipient; // 160
  }
```

That should make for 3 256 bit chunks instead of 4

jaycenhorton (commented):
I  find it a little weird that the market is responsible for replacements. Aside from working around the permissions of Removal.safeTransferFrom, what's the reasoning for this?

contracts/Certificate.sol
amiecorso: Made a new ticket to capture our intent to rename and will revert this rename for now

amiecorso: we'll want to rename on a branch where we're going to actual do a fresh deploy from

contracts/Certificate.sol
amiecorso: great - implemented this but reordered things to keep the bool as the first byte because we need to decode that for determining what kind of batch received this -- new cert or replacement removals

amiecorso: actually never mind, TWL that struct packing only leads to gas benefits when the structs are being written to or read from storage

PR #497, opened 2023-01-19
[NO-2644] Add on-chain counter for certificate contract guarantee

amiecorso:
Decided for now to go with only one counter that will record any discrepancy between NRTs purchase and the number of NRTs actually held by the certificate contract (instead of keeping track of the total purchase amount and also the current total held amount).

This means that value doesn't need to be updated during certificate creation, because there is already a safeguard that ensures the purchase amount and the assigned removal amounts are equal (no need to update a value with the addition of "0", which is the amount of discrepancy introduced by a new certificate).

We just need to update this value when removals are either burned from or replaced to the certificate contract.  This PR takes care of the burning, and in a followup PR that prototypes the replacement mechanism, we see the replacement-related updating.


contracts/Certificate.sol
```diff
@@ -162,6 +171,19 @@ contract Certificate is
...
+  /**
+   * @notice Used to decrement the discrepancy counter when removals are burned from this contract.
+   */
+  function decrementGuaranteeDiscrepancy(uint256 amount)
+    external
+    whenNotPaused
+  {
+    if (_msgSender() != address(_removal)) {
+      revert SenderNotRemovalContract();
+    }
```
rlfarman, line 183: Could we revert if we decremented below 0? That way we're extra extra certain that we never replace more removals than necessary?

contracts/Certificate.sol
```diff
@@ -73,6 +73,15 @@ contract Certificate is
...
   mapping(uint256 => uint256) private _purchaseAmounts;
 
+  /**
+   * @notice Keeps track of any discrepancy between the total number of NRTs guaranteed by this contract and the
+   * number of NRTs currently held. In other words, the number of NRTs held minus the number of NRTs guaranteed.
+   * @dev This is used to provide a redundant, transparent account of the number of NRTs that may still need to be
+   * replaced in the case of released removals. This number should only be non-zero if removals are in the process of
+   * being replaced.
+   */
+  int256 private _guaranteeDiscrepancy;
```
rlfarman, line 83: WRT to what's below, this should never be above 0 (meaning there should never be more replaced removals than have been released. Given that, we might be able to name this in a clearer way (IE - `numberOfRemovalsReleased`).

rlfarman (changes requested):
LGTM, although I `_guaranteeDiscrepancy` isn't the clearest na

contracts/Certificate.sol
amiecorso: as discussed IRL -- happening at the point of replacement in the followup PR

contracts/Certificate.sol
amiecorso: IRL decision, we're going with `nrtDeficit`

rlfarman (approved):

PR #500, opened 2023-01-21
Fixes security issue in approval based `swap`

amiecorso:
We can't actually decouple the msg sender and the purchaser in our approval-based swap endpoint without introducing the risk of some arbitrary address using an approved spender's funds to buy their own carbon removals.

This address decoupling is ok in the permissioned endpoint because the permissions prevents an arbitrary user from spending approved funds, but without some more elaborate mechanisms we can't safely do this with a public endpoint, so I'm reverting it such that the message sender is always the spender in the approval version of the endpoint.

### DIFF
--- a/contracts/Market.sol
+++ b/contracts/Market.sol
@@ -556,43 +556,6 @@ contract Market is
     bytes32 r,
     bytes32 s
   ) external whenNotPaused {
-    _permit({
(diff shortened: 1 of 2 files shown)


jaycenhorton (approved):
:fearful:

PR #503, opened 2023-01-24
[NO-2648] [L-2 Jan Audit] Check return value of `transferFrom`

amiecorso:
See [Jira ticket](https://norinauts.atlassian.net/browse/NO-2648), which also links to audit report, for more details.

Checks the return value of `transferFrom` since unsafe implementations of ERC20 sometimes return `false` _instead of reverting_ if a transfer fails.  In this case, we should revert.


contracts/Market.sol
```diff
@@ -1119,6 +1119,7 @@ contract Market is
       from: 0,
       to: countOfRemovalsAllocated
     });
+    bool successfulTransfer;
```
rlfarman, line 1122: nit - booleans best prefixed with affirmative phrase (`is`, `can`, `has`)
https://learn.microsoft.com/en-us/dotnet/standard/design-guidelines/names-of-type-members#names-of-properties
https://www.comp.nus.edu.sg/~cs2103/AY1617S1/contents/coding-standards-java.html#:~:text=Boolean%20variables%20should%20be%20prefixed%20with%20'is'&text=This%20is%20the%20naming%20convention,names%20like%20status%20or%20flag.

rlfarman (changes requested):
Approved

contracts/Market.sol
rlfarman: IE - `isTransferSuccessful`

amiecorso: haha I actually thought about this while I was naming the variable, but apparently not that hard. supported nit

rlfarman (approved):

PR #504, opened 2023-01-24
[NO-2630] Add `noriFeePercentage` to event emitted upon certificate creation

amiecorso:
Since the fee percentage can potentially change, we need this information to be emitted in the event emitted upon certificate creation to know what the fee was for a specific transaction.  There are more complicated ways to deduce this without emitting and event, but we has consensus on this approach.


rlfarman (commented):
Feeling something nagging me about the growing responsibility of the RecieveRemovalBatch event. Lots of properties being emitted here, and that limits how much we're able to index. So maybe we should rethink how much info we emit from one event.

At the same time, I'm not _super_ worried about this, but might end up regretting it later if we don't index the right properties. 

We should have a quick chat about this after standup tomorrow. and decide if we want to split up the event or not. Otherwise looks very much good to me.

contracts/Removal.sol
```diff
@@ -406,7 +406,13 @@ contract Removal is
       to: address(_certificate),
       ids: ids,
       amounts: amounts,
-      data: abi.encode(certificateRecipient, certificateAmount, address(0), 0)
+      data: abi.encode(
+        certificateRecipient,
+        certificateAmount,
+        address(0),
+        0,
+        0
```
jaycenhorton, line 414: What is the difference between using 0 as the multiple vs 1?

jaycenhorton (approved):
Left one question inline.

Also, is the slither check a new reentrancy vector?

PR #509, opened 2023-01-27
[NO-2700] Add accounting for tracking the number of `RestrictedNORI` that need to/can be manually minted

amiecorso:
See notes on the [Jira ticket](https://norinauts.atlassian.net/browse/NO-2700) for more thorough background in slack and the audit report.

We're choosing to include this accounting now so that it will be accurate from the start, even if we haven't yet built the ability to manually mint `RestrictedNORI` to suppliers.


contracts/RestrictedNORI.sol
```diff
@@ -167,6 +168,17 @@ contract RestrictedNORI is
...
+  /**
+   * @notice Accounting for the maximum amount of RestrictedNORI that should be mintable manually.
+   * @dev In the case of a non-ERC1155-compatible supplier wallet address, minting RestrictedNORI during a
+   * purchase will fail and cause an event to be emitted. This variable is used to track the maximum amount
+   * of RestrictedNORI that can be redmedially minted to a supplier's compatible address to avoid over-
+   * minting the wrapper token and failing to have enough RestrictedNORI backed by wrapped NORI.
+   * TODO This variable should be used to enforce the maximum number of tokens that can ever be minted manually,
+   * and should be decremented when tokens are minted manually, which is not yet implemented.
+   */
+  uint256 private _maxManualMintable;
```
rlfarman, line 180: For the sake of the strongest possible safeguards against minting Restricted NORI, it might make sense to turn this into a mapping from `(address -> uint256)`, so that we know specifically who is owed how much. That way, when we mint Restricted NORI, we prevent ourselves from accidentally minting too much for any given supplier.

Right now, it becomes difficult to safeguard against that possibility in the event that more than one supplier is responsible for incrementing this counter.

contracts/RestrictedNORI.sol
```diff
@@ -375,7 +406,9 @@ contract RestrictedNORI is
       scheduleOwners: tokenHoldersLocal,
       quantitiesBurned: quantitiesToBurnForHolders
     });
-    _underlyingToken.transfer({to: toAccount, amount: quantityToRevoke});
+    if (!_underlyingToken.transfer({to: toAccount, amount: quantityToRevoke})) {
+      revert ERC20TransferFailed();
+    }
```
jaycenhorton, line 411: Why aren't we using [SafeERC20](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/token/ERC20/utils/SafeERC20.sol)

[This one](https://github.com/Vectorized/solady/blob/main/src/utils/SafeTransferLib.sol#L231) is another great alternative that's highly optimized

jaycenhorton (changes requested):
left some comments. I think I missed the market PR for checking erc20 transfers, but I left a note about safe erc20 libraries inline

https://medium.com/coinmonks/missing-return-value-bug-at-least-130-tokens-affected-d67bf08521ca
```
The calling contract sends an external call to the BadToken, which processes the call, makes the transfer and does not return a Boolean return value. The calling contract now looks up the return value in the memory, but since the token did not write a return value it will take whatever it finds in this memory position as the return value of the external call.

This is already pretty bad: taking some data that happens to be in a memory slot as a return value is not a good idea.

By sheer coincidence this problem didn’t surface in the past because the memory slot in which the caller expected the return value, overlapped with the memory slot, the function selector of the call was stored in. This was interpreted by the EVM as the return value “true”.
```

contracts/RestrictedNORI.sol
amiecorso: yeah good idea.... waffled on this a little bit in favor of simplicity but it's not that useful that way. went with the mapping

amiecorso: note that there's not currently a way to sum across all of the addresses because we're not using an enumerable mapping, so we don't know which entries the mapping contains (though these addresses can be retrieved by events, so I don't think we need to further burden this contract in the case that someday we want to understand the *total* deficit of this contract... not only is this scenario exceedingly rare, but there are other avenues to getting the complete set of affected addresses by events and creating the sum using the getter multiple times)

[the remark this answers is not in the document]
amiecorso: yeah true bad name.  renamed esp now that it's a mapping and not a single uint

contracts/RestrictedNORI.sol
```diff
@@ -167,6 +168,17 @@ contract RestrictedNORI is
    */
   bytes32 public constant TOKEN_REVOKER_ROLE = keccak256("TOKEN_REVOKER_ROLE");
 
+  /**
+   * @notice Accounting for the maximum amount of RestrictedNORI that should be mintable manually.
+   * @dev In the case of a non-ERC1155-compatible supplier wallet address, minting RestrictedNORI during a
+   * purchase will fail and cause an event to be emitted. This variable is used to track the maximum amount
+   * of RestrictedNORI that can be redmedially minted to a supplier's compatible address to avoid over-
```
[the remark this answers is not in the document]
amiecorso: good tip! I see the spellchecking happening in solidity files now 🙏

contracts/RestrictedNORI.sol
amiecorso: First of all, fascinating medium article! I actually found myself wondering about whether there are ERC20 tokens out there that don't return anything at all while adding this check.... Turns out there are but they aren't *actually* ERC20s.

With regard to your actual comment here, I think you're confused about what we're doing here.  We are using a SafeERC20 in the sense that `_underlyingToken` here is literally `NORI` ... we originally decided not to bother checking the return value of `transfer` because we never intend to use `RestrictedNORI` with any underlying token other than `NORI`.   We may still want to take this approach, but given that *technically* the underlying token for `RestrictedNORI` IS configurable.... and by similar logic we recently decided to check the return values from transfers in the market contract, I just decided to 1) shut up Slither and 2) future-proof this a litttttle bit by doing the safest thing and just checking the return value.

We don't ever expect there to be a scenario where we're interacting with on of the BadTokens referenced in the article here, so that's way overkill to take into account.  But for the sake of best practice, I just threw the check in here.  Especially because I believe gas concerns are of lesser concern in this contract which will have far fewer interactions and is not buyer-facing.  Let me know if you'd rather I pull it out though.

jaycenhorton: Makes sense given those constraints not to use the safe lib. Was thinking there might be a future where this also used a non nori token (super unlikely). I'd be in favor of removing the check here as it sets precedent for code style and the check itself is never needed.

jaycenhorton (approved):

contracts/RestrictedNORI.sol
amiecorso: makes sense for me - gonna go ahead and remove the check.  if we're ever using a non-nori token as the underlying here I'm sure this entire contract will receive another round of scrutiny and we'll check return values then

PR #510, opened 2023-01-28
[NO-2658] Multicall for migration, script improvements, and testing on Mumbai

amiecorso:


tasks/migrate-certificates.ts
```diff
@@ -33,7 +33,7 @@ interface MigratedCertificate {
   };
   ids: string[];
   amounts: number[];
-  txReceipt?: TransactionReceipt;
+  txReceipt?: { transactionHash?: string; status?: number };
```
amiecorso, line 36: when including the full transaction receipt for every single certificate, the resulting output object was too big to stringify! So here I'm including the only two pieces of the transaction receipt that are used by the datastore update.  Since you can always retrieve a receipt by its hash if we need more details in the future, the hash should suffice.

tasks/migrate-removals.ts
```diff
@@ -76,33 +77,34 @@ const summarize = async ({
...
+  const multicallData = allRemovalIds.map((tokenId) =>
+    removalContract.interface.encodeFunctionData('totalSupply', [tokenId])
+  );
+  const removalAmountsOnChain = await removalContract.callStatic.multicall(
+    multicallData
+  );
+  const bigNumberRemovalAmounts = removalAmountsOnChain.map((amount) =>
+    BigNumber.from(amount)
+  );
+  const totalOnChainSupply = bigNumberRemovalAmounts.reduce(
```
rlfarman, line 90: This is just fine but you can also avoid needing to write your own reducer with something like `BigNumber.sum.apply(null, bigNumberRemovalAmounts)`

tasks/migrate-removals.ts
```diff
@@ -245,36 +283,84 @@ export const GET_MIGRATE_REMOVALS_TASK = () =>
...
-          }).flatMap((log) =>
-            log.args.ids.map((id: BigNumber) => id.toString())
-          );
-          if (txReceipt.status !== 1) {
-            logger.error(
-              `❌ Transaction ${pendingTx.hash} failed with failure status ${txReceipt.status} - exiting early`
+          } else {
+            // all other cases
+            maybePendingTx = await callWithTimeout(
+              migrationFunction(
```
jaycenhorton, line 322: Are you _**sure**_ this works as expected? When you use a promise, **_even if you don't await it_**, it starts to execute

jaycenhorton: to avoid that, pass it as a callback instead (`()=> fn()`) or use the provider timeout setting I linked before

tasks/migrate-certificates.ts
```diff
@@ -389,106 +423,247 @@ export const GET_MIGRATE_CERTIFICATES_TASK = () =>
...
+            multicallBatches.push(batch);
+          }
+          batch = [certificate];
+          multicallBatches.push(batch);
+          singularBatchCount += 1;
+          batch = [];
+        }
+        batch.push(certificate);
+      }
+      multicallBatches.push(batch);
```
jaycenhorton, line 483: :cold_sweat:

jaycenhorton (approved):
see the note about the provider timeout before merging. The rest of the comments are suggestions/feedback

tasks/migrate-certificates.ts
jaycenhorton: In the future, refer to to the [chunk utility in the mono repo](https://github.com/nori-dot-eco/nori/blob/master/nori-graphql/src/utils.ts#L75) for how to chunk arrays

contracts.json
```diff
@@ -90,22 +90,22 @@
   },
   "mumbai": {
     "BridgedPolygonNORI": {
-      "proxyAddress": "0xB3fe45C08137dD6adACb2918D899e0C0dBB036C8"
+      "proxyAddress": "0xd768Df6475a853194dB4afF2dDbDC0ea29F05c66"
```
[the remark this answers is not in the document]
amiecorso: these may have changed due to Scott's original activity on this branch, in which I'm assuming he correctly redeployed proxies + implementations...
I only ever used one deployment command, which was `hardhat deploy --network mumbai --tags market`

This is actually making me curious about how my deployments would have broken CI, because I would have thought they would have upgraded the new proxy addresses that were created here as opposed to the existing ones that Defender currently relies on.  I'm not sure.

I suppose we also need to decide if we want to merge these proxy addresses or not.

tasks/migrate-certificates.ts
amiecorso: yeah that's a clean implementation... this was harder because it's not necessarily creating equally sized chunks - it's dealing with large certificates in a separate way

README.md
```diff
@@ -38,6 +38,14 @@ This command will do the following:
 2. Seed test wallets with funds.
 3. Run the deployment scripts which seed the test contracts.
 
+### Deploying the market contracts on Mumbai or Polygon
+
+Note that you will need to have .env settings configured for production or the deployment will be prevented.
```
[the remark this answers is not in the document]
amiecorso: I think it's just `SOLC_PROFILE`, so i specified that

tasks/migrate-certificates.ts
```diff
@@ -338,6 +355,23 @@ const printSummary = ({
   });
 };
 
+const callWithTimeout = async (
```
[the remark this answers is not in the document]
amiecorso: I'm having a hard time figuring out how to do this given the hardhat/fireblocks environment that we're in and the fact that the construction of the provider is highly abstracted and bundled with the construction of the signer...

Is there anything wrong with using our own timeout?

contracts.json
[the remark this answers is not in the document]
scottynomad: Correct, I force deployed new proxies since I was trying to start with fresh state.  Wasn't necessary for BPNori but that flag is all or nothing and the dependency graph in the deployment scripts pulled this in.

tasks/migrate-removals.ts
amiecorso: fixed this using the callback approach.  this was working because it was simply meant to make sure that a promise doesn't hang forever, but I see how passing a promise that has already begun execution means that it may end up running longer than the timeout before things actually time out, which could be unexpected behavior in other applications of a function like this.

PR #521, opened 2023-02-08
[NO-2661] Deployment script tweaks

amiecorso:
Deployment artifacts were merged in https://github.com/nori-dot-eco/contracts/pull/529 while we debugged failing tests due to script tweaks.

This PR includes the scripts that were used while deploying with the fixed bug (see latest commit on this branch for bug fix)

Double-checked the polygon contracts to make sure all contracts have correct registered addresses and this appears to be the case.


plugins/index.ts
```diff
@@ -116,47 +137,56 @@ const deployOrUpgradeProxy = async <
...
+      if (typeof fireblocksSigner.setNextTransactionMemo === 'function') {
+        fireblocksSigner.setNextTransactionMemo(
+          `Upgrade contract instance for ${contractName}`
+        );
+      }
+      const deployment = await hre.deployments.get(contractName);
+      const artifact = await hre.deployments.getArtifact(contractName);
+      if (deployment.bytecode !== artifact.bytecode) {
+        contract = await hre.upgrades.upgradeProxy<TContract>(
+          maybeProxyAddress!,
```
rlfarman, line 160: Would be nice to maybe include this in the if statement above  (IE - `if (deployment.bytecode !== artifact.bytecode && maybeProxyAddress !== undefined)` so we don't have to assert that it's defined here and everywhere else it's used

plugins/index.ts
```diff
@@ -95,9 +111,14 @@ const deployOrUpgradeProxy = async <
...
-  if (shouldDeployProxy) {
+  if (
+    shouldDeployProxy &&
+    !['bridgedpolygonnori', 'nori', 'lockednori'].includes(
+      contractName.toLowerCase()
+    )
+  ) {
     hre.trace('Deploying proxy and instance', contractName);
-    const fireblocksSigner = signer as FireblocksSigner;
+    const fireblocksSigner = signer as any as FireblocksSigner;
```
rlfarman, line 121: Possible to fix this before merging?

deploy/configure-assets-after-deployment.ts
```diff
@@ -9,43 +10,86 @@ import {
   getRestrictedNORI,
 } from '@/utils/contracts';
 
+const CONFIRMATIONS = 5;
 export const deploy: DeployFunction = async (environment) => {
   const hre = environment as unknown as CustomHardHatRuntimeEnvironment;
   Logger.setLogLevel(Logger.levels.DEBUG);
   hre.trace(`configure-assets-after-deployment`);
+  // const provider = new ethers.providers.JsonRpcProvider(
```
rlfarman, line 18: If you'd like to keep this commented code for posterity it'd be nice to see a comment above it like you did elsewhere in this PR.

rlfarman (commented):
Will review again when tests are passing - just a few comments for now

plugins/index.ts
amiecorso: yep - it can removed now that our impersonation code isn't active

plugins/index.ts
amiecorso: so the only reason this is showing up as a diff in this PR is because of some large-scale block rearrangements into try/catch statements.  I hesitate to adjust this logic now because I'd like the contents of this file to reflect what they were at the time they were used to successfully create a deployment.  I think if we want to update this in the future we should do so in a separate PR and ideally with a deployment of some sort to verify that they work?

rlfarman (approved):

PR #528, opened 2023-02-14
Foundry invariants prototype

amiecorso:
Check out the docs on invariant testing using foundry: https://book.getfoundry.sh/forge/invariant-testing
TL;DR:  invariant testing allows you to define invariants about your system, such as "a removal token should only ever be held either by the encoded supplier address, the market, the certificate, or the consignor", and then to fuzz your system with lots of randomized inputs while checking continuously that your asserted invariants still hold. 

From the docs:
"Invariant testing allows for a set of invariant expressions to be tested against randomized sequences of pre-defined function calls from pre-defined contracts. After each function call is performed, all defined invariants are asserted.
Invariant testing is a powerful tool to expose incorrect logic in protocols. Due to the fact that function call sequences are randomized and have fuzzed inputs, invariant testing can expose false assumptions and incorrect logic in edge cases and highly complex protocol states."

This PR introduces just one invariant, and sets up a handler contract that is designed to wrap functionalities of the market such as minting, listing, purchasing, releasing and replacing in such a way that the randomness of fuzzing can be incorporated into these behaviors while favoring states that do not result in meaningless reverts. 

In the future, we can expand on this structure to define more market-level invariants, and perhaps to define more single-contract-specific handlers (and invariants) that could test more granular invariants in our contracts around things like permissions, configurations, etc.

See function documentation and PR comments for more information on the specifics of what's happening here.


foundry.toml
```diff
@@ -30,4 +30,10 @@ mainnet = "${ETHEREUM_RPC_URL}"
 goerli = "${GOERLI_RPC_URL}"
 polygon = "${POLYGON_RPC_URL}"
 
+
+[invariant]
+runs = 100
+depth =1
```
amiecorso, line 36: runs is the number of times a randomized call sequence will be generated
depth is the number of random calls that will be made

This was set to 1 to debug a handler function but we should put it back to its default of 15 or so when ready

foundry.toml
```diff
@@ -30,4 +30,10 @@ mainnet = "${ETHEREUM_RPC_URL}"
...
+fail_on_revert = true
```
amiecorso, line 37: determines whether reverts during the fuzzing are acceptable.  usually, you probably want this set to false so that reverts ARE allowed... after all, given randomized inputs, reverts are expected and often the correct behavior of contracts.  However, certain tests could be designed in such a way and inputs can be bounded/assumed (forge cheatcodes) or manipulated such that reverts are NOT expected.  it just depends.  Handlers should usually be designed so that reverts aren't happening all of the time -- otherwise no actual state changes are being meaningfully tested.

test/Invariants.t.sol
```diff
@@ -0,0 +1,56 @@
+/* solhint-disable contract-name-camelcase, func-name-mixedcase, reason-string */
```
amiecorso, line 1: This is the file that actually defines the invariants (so the contracts defined in here are what would be called like `forge test --mc MyInvariantContract`).  The name `Invariants.t.sol` is arbitrary -- I should probably call this `MarketInvariants.t.sol`

test/helpers/MarketHandler.sol
```diff
@@ -0,0 +1,227 @@
+/* solhint-disable contract-name-camelcase, func-name-mixedcase, not-rely-on-time */
```
amiecorso, line 1: this contract is what will be called and fuzzed by the fuzzing engine (its external functions at least).  randomized call sequences will be made with equal probability among external functions up to the number of calls in the `depth` setting in config. (currently this handler only has 1 external function and my depth was set to 1 so each run will simply call `endToEnd` once, but I'm going to change that)

"handler" contracts like this are used to customize and arrange logic in a more complex system so that fuzzing inputs can be meaningful and not just a bunch of meaningless reverts.  sky's the limit in here.

test/Invariants.t.sol
```diff
@@ -0,0 +1,56 @@
...
+  function setUp() external {
+    marketHandler = new MarketHandler();
+    _certificate = Certificate(marketHandler.getCertificateAddress());
+    _removal = Removal(marketHandler.getRemovalAddress());
+    targetContract(address(marketHandler)); // only target the handler
+    // _selectors = [MarketHandler.endToEnd.selector];
+    // targetSelector(FuzzSelector(address(marketHandler), _selectors));
+  }
+
+  function invariant_sumOfPurchaseAmounts() external {
```
amiecorso, line 29: like `test` in forge tests that we're familiar with, invariant tests are indicated by prefixing the test name with `invariant_` 

here I have just one invariant defined -- which asserts that the sum of purchase amounts in the market should always be equal to the sum of the certificate contracts balances of all removal ids that it owns any amount of, less its own reported nrtDeficit.

this invariant relies on a "ghost variable" in the handler contract, which keeps track of which removals the certificate has any ownership of, so that a sum can be performed while checking the invariant.

test/checkout.int.t.sol
```diff
@@ -13,7 +13,7 @@ abstract contract Checkout is UpgradeableMarket {
   uint256[] internal _removalIds;
   uint256 internal _certificateTokenId;
 
-  bytes32 constant RECEIVE_REMOVAL_BATCH_EVENT_SELECTOR =
+  bytes32 constant CREATE_CERTIFICATE_EVENT_SELECTOR =
```
amiecorso, line 16: caught a need for rename

foundry.toml
```diff
@@ -30,4 +30,10 @@ mainnet = "${ETHEREUM_RPC_URL}"
 goerli = "${GOERLI_RPC_URL}"
 polygon = "${POLYGON_RPC_URL}"
 
+
+[invariant]
+runs = 400
+depth = 15
+fail_on_revert = false
```
amiecorso, line 37: runs is the number of times a randomized call sequence will be generated
depth is the number of random calls that will be made for a given run
fail_on_revert = false means that we expect some of the randomized calls to revert, and that's ok as long as our defined invariants hold in all cases

test/MarketInvariants.t.sol
```diff
@@ -0,0 +1,87 @@
+/* solhint-disable contract-name-camelcase, func-name-mixedcase, reason-string */
```
amiecorso, line 1: This is the file that actually defines the invariants (so the contracts defined in here are what would be called like forge test --mc MarketInvariantTest).

test/MarketInvariants.t.sol
```diff
@@ -0,0 +1,87 @@
...
+  Removal internal _removal;
+
+  function setUp() external {
+    marketHandler = new MarketHandler();
+    _certificate = Certificate(marketHandler.getCertificateAddress());
+    _removal = Removal(marketHandler.getRemovalAddress());
+    targetContract(address(marketHandler)); // only target the handler
+  }
+
+  function invariant_sumOfPurchaseAmounts() external {
```
amiecorso, line 25: like test in forge tests that we're familiar with, invariant tests are indicated by prefixing the test name with invariant_

here I have just one invariant defined -- which asserts that the sum of purchase amounts in the market should always be equal to the sum of the certificate contracts balances of all removal ids that it owns any amount of, less its own reported nrtDeficit.

this invariant relies on a "ghost variable" in the handler contract, which keeps track of which removals the certificate has any ownership of, so that a sum can be performed while checking the invariant.

test/helpers/MarketHandler.sol
```diff
@@ -0,0 +1,293 @@
...
+   * - the vintage is between 1990 and 2019
+   * - the subIdentifier is between 0 and 100 million
+   * - the supplier is chosen randomly from a predefined set of possible suppliers
+   * - the amounts are chosen randomly between 0 and 1 ether-
+   * - the projectId is between 1 and 1 billion
+   * - the scheduleStartTime is between 1 and intMax minus 10 years (avoid overflow when calculating end time for schedules)
+   * - the holdbackPercentage is between 0 and 100
+   * - other ID fields are hardcoded
+   */
+  function mintRemovalBatchMinimizeReverts(uint256 fuzz) external {
```
rlfarman, line 58: I might be wrong here, but would it be easier to set up functions like this one by using bounded args instead of using a generic fuzz across a bunch of different variables?
So instead of one arg, (fuzz), you'd have (batchSize, supplierIndex, vintage, subIdentifier)... And then subIdentifier, for example, would look like `subIdentifier = bound(1, 100_000_000)`.

rlfarman (approved):
Very fancy! Left one comment, mostly just a curiosity about how we might be able to simplify writing some of these functions inside `MarketHandler`. Would love to sort through our list of todos and pick out what we feel like we should prioritize first.

foundry.toml
scottynomad: My hunch is we'll want more depth to really push on our assumptions since doing anything useful with the market requires ~5 steps.

scottynomad (approved):
Super cool, thanks for laying down some patterns for this!  I want to build some invariant tests for LockedNORI now!

test/helpers/MarketHandler.sol
amiecorso: so I thought carefully about this one and the reasons I ended up parameterizing these functions in the simplest possible way and then transforming the entropy internally were twofold:
1) most importantly, these functions may also be used as building blocks for other more complicated functions (such as how they're called in `endToEnd`).  when they're called indirectly like that, the fuzzer doesn't actually have access to their parameters, the calling function itself has to pass args.  so it's easier to pass around a single source of entropy than it is to have to parameterize the calling function with all the fuzzed params that might be required by its helpers.
2) some of these inputs require more complicated transformation than is offered by "bound" (such as modulation to achieve an index or something like that) so it seemed simpler to do all of those transformations internally.

PR #532, opened 2023-02-17
NO-2810 Working through some config before launch and fixing USDC decimals.

scottynomad:
# TLDR

This started out as Amie and I trying to get all the final configs in place for the contracts but ended up becoming a bit of a rabbit hole due to the discovery that USDC uses 6 decimals rather than 18.

## Adding NoriUSD test contract

I added a `NoriUSDC` contract that is intentionally not BPNori with decimals() set to six like the real USDC.  This is intended only for local development.  The deploy scripts are set to configure the market with the official USDC bridged contracts on polygon and mumbai.

## Issue: Removals use 18 decimals, USDC uses 6

We have a couple of options here, removals are mostly 18 decimals by convention without it being strictly enforced anywhere.  If we instead made it more like 4, then we could use price multiple to cope with pricing now for USDC and later for NORI while still allowing two decimal places for fees to be accurately applied.  This seems like kind of a big sacrifice that we don't have a clear path away from after it's out there.

The other option (prototyped here) is to compensate for the differing decimals when calculating `checkoutTotal` and whenever we do transfers in or out from the purchase token.  This works but adds another layer of complexity therefore risk.  I've tried to prototype it by considering it an interface concern and keeping the conversion as close as possible to where the code interacts with the ERC20 contract while trying to cleanly enforce that all internal representations and logic always operates on the Removal standard of 18 decimals. 

In the process I was compelled to tweak the semantics of the `swap` functions so that the `amount` parameters means the quantity of Removals purchased rather than the amount of money paid.  As we've gotten further and further away from the 1:1 ideal the convention of passing in a payment amount there felt very awkward.  I think the change has actually simplified things somewhat.  So now the integration contract works as follows:

1. (optional) call `calculateCertificateTotalFromCheckoutAmount` (conceptually, I've got $180, how many NRTs can I buy?)
2. call `calculateCheckoutTotal(#NRTsToBuy)` to get the total amount to pay in the purchaseToken with decimals accounted for.
3. sign the permit with that amount
4. call `swap` where `amount` is `#NRTsToBuy` as used in step 2.

The swap functions all use `calculateCheckoutTotal` internally so they can easily recreate the correct amount to pass to the permit check with the same truncation of decimals beyond 6.

# Status

- [x] Foundry tests updated
- [x] Hardhat tests updates
- [ ] Scrub through docstrings for naming
- [ ]  (later) improve tests for naming, repetitive setup, composability, configurability (i.e. USDC vs NORI)


contracts/Market.sol
```diff
@@ -126,6 +126,12 @@ contract Market is
    */
   IERC20WithPermit private _purchasingToken;
 
+  /**
+   * @notice The number of decimals configure on the _purchasingToken contract.
+   * Used to adapt between Removal amounts and payment token amounts.
+   */
+  uint8 private _purchasingTokenDecimals;
```
amiecorso, line 133: there might be something valuable about the explicitness of storing this here and requiring it up on configuration of the purchasing token even if you were able to rely on calling `_purchasingToken.decimals()` to get that where necessary

contracts/Market.sol
```diff
@@ -670,8 +688,8 @@ contract Market is
    * - Can only be used when this contract is not paused.
    * @param recipient The address to which the certificate will be issued.
    * @param permitOwner The address that signed the EIP2612 permit and will pay for the removals.
-   * @param amount The total purchase amount in ERC20 tokens. This is the combined total price of the removals being
-   * purchased and the fee paid to Nori.
+   * @param amount The total amount iof Removals being purchased. This is the combined total price of the removals being
```
amiecorso, line 691:
```suggestion
   * @param amount The total amount of Removals being purchased. This is the combined total price of the removals being
```

contracts/Market.sol
```diff
@@ -686,8 +704,7 @@ contract Market is
     bytes32 r,
     bytes32 s
   ) external whenNotPaused {
-    uint256 certificateAmount = this
-      .calculateCertificateAmountFromPurchaseTotal({purchaseTotal: amount});
+    uint256 certificateAmount = amount;
```
amiecorso, line 707: looks like we could get away with just not assigning this variable at all and passing `amount` directly as the argument in below where `certificateAmount` is passed

contracts/Market.sol
```diff
@@ -735,12 +752,11 @@ contract Market is
    * - Can only be used when this contract is not paused.
    * - Can only be used if this contract has been granted approval to transfer the sender's ERC20 tokens.
    * @param recipient The address to which the certificate will be issued.
-   * @param amount The total purchase amount in ERC20 tokens. This is the combined total price of the removals being
+   * @param amount The total amount of Removals to purchase. This is the combined total price of the removals being
    * purchased and the fee paid to Nori.
```
amiecorso, line 756:
```suggestion
   * @param amount The total amount of Removals to purchase.
```

contracts/Market.sol
```diff
@@ -670,8 +688,8 @@ contract Market is
    * - Can only be used when this contract is not paused.
    * @param recipient The address to which the certificate will be issued.
    * @param permitOwner The address that signed the EIP2612 permit and will pay for the removals.
-   * @param amount The total purchase amount in ERC20 tokens. This is the combined total price of the removals being
-   * purchased and the fee paid to Nori.
+   * @param amount The total amount iof Removals being purchased. This is the combined total price of the removals being
+   * purchased and the fee paid to Nori. (See calculateCheckoutTotal())
```
amiecorso, line 692:
```suggestion
   * @param amount The total amount iof Removals being purchased.
```

contracts/Market.sol
```diff
@@ -735,12 +752,11 @@ contract Market is
...
    * - Can only be used if this contract has been granted approval to transfer the sender's ERC20 tokens.
    * @param recipient The address to which the certificate will be issued.
-   * @param amount The total purchase amount in ERC20 tokens. This is the combined total price of the removals being
+   * @param amount The total amount of Removals to purchase. This is the combined total price of the removals being
    * purchased and the fee paid to Nori.
    */
   function swap(address recipient, uint256 amount) external whenNotPaused {
-    uint256 certificateAmount = this
-      .calculateCertificateAmountFromPurchaseTotal({purchaseTotal: amount});
+    uint256 certificateAmount = amount;
```
amiecorso, line 759: same thinking on potentially not assigning this variable here and in a few other places

contracts/Market.sol
```diff
@@ -1066,23 +1074,49 @@ contract Market is
...
+
+  function convertRemovalAmountToPurchasingTokenAmount(uint256 removalAmount)
+    public
+    view
+    returns (uint256)
+  {
+    if (_purchasingTokenDecimals == 18) {
+      return removalAmount;
+    }
+    int8 decimalDelta = 18 - int8(_purchasingTokenDecimals);
```
amiecorso, line 1091: is there a specific reason for converting this value to a uint8?  gas savings for smaller data types only apply when these things are actually being stored in storage memory (and you can pack them tightly) and it actually can cost more gas to cut something down to a uint8 when handling in-memory since the EVM word size is 256 bits

amiecorso: oh I see, it's because the `_purchasingTokenDecimals` variable is a uint8.  I believe that unless this value is part of a struct or an array and is being packed tightly with other values that are sub-256 bytes, there is still no savings on making this a smaller data type than 256 even though it is a storage variable

contracts/Market.sol
```diff
@@ -1066,23 +1074,49 @@ contract Market is
...
+  }
+
+  function convertPurchasingTokenAmountToRemovalAmount(
+    uint256 purchasingTokenAmount
+  ) public view returns (uint256) {
+    if (_purchasingTokenDecimals == 18) {
+      return purchasingTokenAmount;
+    }
+    int8 decimalDelta = 18 - int8(_purchasingTokenDecimals);
+    return purchasingTokenAmount * 10**uint8(decimalDelta);
```
amiecorso, line 1102: same notes on uint8 being unnecessary

contracts/Market.sol
```diff
@@ -1336,7 +1374,9 @@ contract Market is
       isTransferSuccessful = _purchasingToken.transferFrom({
         from: from,
         to: suppliers[i],
-        amount: unrestrictedSupplierFee
+        amount: convertRemovalAmountToPurchasingTokenAmount(
+          unrestrictedSupplierFee
+        )
```
amiecorso, line 1379: same thoughts on perhaps doing this conversion earlier so the naming is less confusing

deploy/configure-assets-after-deployment.ts
```diff
@@ -91,10 +118,34 @@ export const deploy: DeployFunction = async (environment) => {
...
   }
+
+  if (
+    (await market.getPurchasingTokenAddress()) !== purchaseTokenAddress ||
+    (await market.getPriceMultiple()) !== priceMultiple
+  ) {
+    txn = await market.setPurchasingTokenAndPriceMultiple(
+      purchaseTokenAddress,
+      priceMultiple
+    );
```
amiecorso, line 129: looks like we need the decimals passed here

deploy/configure-assets-after-deployment.ts
```diff
@@ -91,10 +118,34 @@ export const deploy: DeployFunction = async (environment) => {
...
+    await txn.wait(CONFIRMATIONS);
+    hre.trace('Set USDC as purchase token with price multiple of 2000');
+  }
+
+  if ((await market.getNoriFeePercentage()) !== BigNumber.from(25)) {
+    txn = await market.setNoriFeePercentage(25);
+    await txn.wait(CONFIRMATIONS);
+    hre.trace('Set fee percentage to 25');
```
amiecorso, line 137: I wonder if we should define this fee percentage somewhere more central, like at least as a global variable in this script but maybe as an environment variable or something

amiecorso: this might also be helpful for the price multiple values used above

amiecorso:
Thanks for doing this!! I like what you've prototyped here... it makes the most conceptual sense to me to leave removals with 18 decimals, especially because this has become convention for tokens, and removals are tokens too!  It feels straightforward to me to perform decimal-related conversions just around the currency values, and it doesn't feel like there is an extreme number of them. (with the restricted and unrestricted supplier fees, I also think that we can do this conversion earlier and therefore in even fewer places).  I think it also makes sense for this to be part of this same PR even though it's big -- we need that `NoriUSDC` infrastructure for testing decimals, and we need the decimal work you introduced here to correctly implement `NoriUSDC`

jaycenhorton:
RE: Issue 1
- USDC addresses on various testnets: https://developers.circle.com/developer/docs/usdc-on-testnet
- official usdc testnet faucet: https://usdcfaucet.com/ (there are a handful of others as well)
- you can also bridge usdc to mumbai from goerli (I have a handful of goETH that I can swap for a significant amount of testnet USDC)

contracts/Market.sol
scottynomad: It was mostly doing it to be consisent with the return value of `decimals()` but happy to change it if you think it improves gas.

test/Market.t.sol
```diff
@@ -1595,3 +1623,102 @@ contract MarketSupplierSelectionNotUsingUpSuppliersLastRemoval is
...
+    _market.swap(
+      owner,
+      owner,
+      numberOfNRTsToPurchase,
+      signedPermit.permit.deadline,
+      signedPermit.v,
+      signedPermit.r,
+      signedPermit.s
+    );
+    vm.stopPrank();
```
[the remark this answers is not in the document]
scottynomad: definitely, unfinished thought ...

deploy/configure-assets-after-deployment.ts
scottynomad: agreed, we're kindof lacking an environment to contract config mapping somewhere more organized.

contracts/Market.sol
scottynomad: Totally, I'd rather do that but couldn't get it to work.  Would love it if someone can!

scottynomad: Fixed, thanks!

contracts/Market.sol
```diff
@@ -1284,7 +1312,9 @@ contract Market is
           address(_purchasingToken)
         ) {
           emit SkipRestrictedNORIERC20Transfer({
-            amount: restrictedSupplierFee,
+            amount: convertRemovalAmountToPurchasingTokenAmount(
+              restrictedSupplierFee
+            ),
```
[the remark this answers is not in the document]
scottynomad: Fixed.

scottynomad:
> RE: Issue 1
> 
> * USDC addresses on various testnets: https://developers.circle.com/developer/docs/usdc-on-testnet
> * official usdc testnet faucet: https://usdcfaucet.com/ (there are a handful of others as well)
> * you can also bridge usdc to mumbai from goerli (I have a handful of goETH that I can swap for a significant amount of testnet USDC)

Thanks.  I updated the deploy scripts to assume staging uses the real USDC.  It's easy enough to switch to our own if we decide we want to for some specific test.

contracts/Market.sol
```diff
@@ -1267,24 +1297,29 @@ contract Market is
...
           _restrictedNORI.getUnderlyingTokenAddress() !=
           address(_purchasingToken)
         ) {
           emit SkipRestrictedNORIERC20Transfer({
-            amount: restrictedSupplierFee,
+            // This assumes that rnori will only ever be backed by NORI or a 1:1 18 decimal token ...
+            // Not ideal to have to convert back to Removal units here but it minimizes the code changes elsewhere.
+            amount: convertPurchasingTokenAmountToRemovalAmount(
+              restrictedSupplierFee
+            ),
```
scottynomad, line 1322: @amiecorso I realized I'm not clear on what the units are supposed to be for this amount parameter.  I've configured it to be removals at the moment but I guess since the purchasing token is an argument it could just as well be the purchase token amount. 

The docs just say `The amount of RestrictedNORI in the transfer attempt.` which is ambiguous in this fun new world.  :/

amiecorso: so the `restrictedSupplierFee` and `unrestrictedSupplierFee` are currency amounts, not removal amounts, since the RestrictedNORI contract handles currency and these values are generated from removal amounts while incorporating the price multiple.   I'm thinking that RestrictedNORI will work fine as-is, assuming the underlying token that it wraps is never changed after schedules are already in action (something that wouldn't make sense anyway, even if there wasn't this additional precision specter).  That's because it's always working 1:1 with amounts of underlying token and amounts of RestrictedNORI that get minted to represent that underlying token.  So when going to make a withdrawal, if there was a schedule holding $1 there would have been 1_000_000 RestrictedNORI minted there in the first place.  

So this is correct to apply decimal-related conversions to any currency amounts being passed to RestrictedNORI functions.

scottynomad: Ok, so amount in the error event is in _purchasingToken currency units.  I will remove the change I made then.

amiecorso: yeah should always be in currency, everything about restrictedNORI 👍

scottynomad: Reverted my change to this event so amount is in _paymentToken units.

contracts/Market.sol
```diff
@@ -1111,18 +1129,27 @@ contract Market is
...
+      convertPurchasingTokenAmountToRemovalAmount(purchaseTotal).mulDiv(
+        10000,
+        (100 + _noriFeePercentage) * _priceMultiple
+      );
   }
 
   /**
-   * @notice Calculates the quantity of carbon removals being purchased given the purchase total and the price multiple.
+   * @notice Calculates the quantity of carbon removals that can be purchased given
+   * some payment amount taking into account NRT price but excludign fees.
```
amiecorso, line 1140:
```suggestion
   * some payment amount taking into account NRT price but excluding fees.
```

rlfarman (approved):

deploy/configure-assets-after-deployment.ts
```diff
@@ -30,20 +38,38 @@ export const deploy: DeployFunction = async (environment) => {
...
   const market = await getMarket({ hre, signer });
   const certificate = await getCertificate({ hre, signer });
   const rNori = await getRestrictedNORI({ hre, signer });
   const removal = await getRemoval({ hre, signer });
   const bpNori = await getBridgedPolygonNori({ hre, signer });
+
+  // SW: Leaving the default local configuration as bridged polygon NORI
+  // for the purchase token to minimize test breakage.
+  let purchaseTokenAddress = bpNori.address;
+  let priceMultiple = BigNumber.from(100);
```
jaycenhorton, line 52: @scottynomad why is this 100?

deploy/deploy-market.ts
```diff
@@ -34,6 +34,9 @@ deploy.dependencies = [
   'BridgedPolygonNORI',
   'RestrictedNORI',
 ];
+if (hre.network.name !== 'polygon') {
+  deploy.dependencies = [...deploy.dependencies, 'NoriUSDC'];
```
jaycenhorton, line 38: @scottynomad is this still needed?

jaycenhorton (approved):

deploy/configure-assets-after-deployment.ts
scottynomad: Because that's the correct price multiple for NORI (price X 100).  I didn't push the usdc + 20x multiplier through all the tests because it broke too many of them and this is still the long termp lan.

deploy/deploy-market.ts
scottynomad: It's needed for local only now but should get updated to exclude mumbai as well.

jaycenhorton: gotcha. I think we can drop this entirely in the future actually by just using a fork of mumbai as the base block in tests

jaycenhorton: yeah using account impersonation

PR #535, opened 2023-02-22
Use ERC20 method to get decimals instead of state variable

rlfarman:
Updates `IERC20WithPermit` to use `IERC20MetadataUpgradeable` (which, in turn, uses `IERC20Upgradeable`, so we can get rid of it there) to allow us to look up the decimals for _purchasingToken instead of needing to explicitly store it as a state variable.

Some Qs after talking with Jaycen: Requires redeploy of LockedNori? Should we use two different interfaces to avoid redeploy?

### DIFF
--- a/.github/workflows/slither.yml
+++ b/.github/workflows/slither.yml
@@ -26,7 +26,7 @@ jobs:
           restore-keys: |
             ${{ runner.os }}-yarn-
       - name: Run Slither
-        uses: crytic/slither-action@v0.2.0
+        uses: crytic/slither-action@v0.3.0
         id: slither
         with:
           sarif: results.sarif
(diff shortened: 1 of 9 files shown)


scottynomad:
> Some Qs after talking with Jaycen: Requires redeploy of LockedNori? Should we use two different interfaces to avoid redeploy?

Redeploying LockedNORI is not an option.  I will look at splitting the interface so we don't disrupt LockedNORI.

scottynomad:
> > Some Qs after talking with Jaycen: Requires redeploy of LockedNori? Should we use two different interfaces to avoid redeploy?
> 
> Redeploying LockedNORI is not an option. I will look at splitting the interface so we don't disrupt LockedNORI.

Actually I practiced deploying the contracts from master locally then switching to this branch and LockedNORI doesn't get upgraded, only Market does (plus new deployment of NoriUSDC).  This makes sense to me since LockedNORI is a consumer of the changed interface not an implementer of it and the contract still compiles down to the same bytecode with the interface change.

PR #543, opened 2023-02-23
Deployment artifacts from polygon bugfix deploy.

scottynomad:
Salient change here is deploying Market with the decimals fix but it seems like some of the other artifacts may have been out of date.  All the market-related contracts also got upgrades but there shouldn't have been any functional changes in them.


jaycenhorton:
Guessing all the contracts got upgrades due to the optimizer change

jaycenhorton (approved):

scottynomad:
> Guessing all the contracts got upgrades due to the optimizer change

Makes sense, I just noticed that the optimizer settings had changed.

PR #546, opened 2023-02-25
make buyNRTs agnostic to purchasing token and clean up misc. permit utils

jaycenhorton:
- tweaks `buyNRTs` to make the script agnostic to the underlying purchasing token
- uses the [`parseJson`](https://book.getfoundry.sh/cheatcodes/parse-json?highlight=parsejson#parsejson) and `readFile` cheat codes to parse `contracts.json` so that we can load contract addresses dynamically in scripts
- tweaks buyNRTs so that it accepts a mnemonic instead of a private key
- removes/refactors a few redundant permit utils


rlfarman (approved):

script/BuyNRT.s.sol
```diff
@@ -1,123 +1,61 @@
...
- * export BUYER_PRIVATE_KEY="<0x00>"
+ * export PURCHASING_TOKEN_ADDRESS="0x0FA8781a83E46826621b3BC094Ea2A0212e71B23" # USDC
+ * export TONNES=1
+ * export MNEMONIC="..."
  * export RECIPIENT_ADDRESS="<0x00>"
  * export MUMBAI_RPC_URL="https://..."
+ * export NETWORK_NAME="mumbai"
  *
- * forge script script/BuyNRT.s.sol:BuyNRT --rpc-url "$MUMBAI_RPC_URL" --private-keys "$BUYER_PRIVATE_KEY" --broadcast -vvvv
+ * forge script script/BuyNRT.s.sol:BuyNRT --rpc-url="$NETWORK_NAME" --mnemonics="$MNEMONIC" --broadcast -vvvv
```
scottynomad, line 21: `--rpc-url="$NETWORK_NAME"` This doesn't make sense...

script/BuyNRT.s.sol
```diff
@@ -1,123 +1,61 @@
...
-  bytes32 public constant PERMIT_TYPEHASH =
-    0x6e71edae12b1b97f4d1f60370fef10105fa2faae0126114a169c64845d6126c9;
-
-  BridgedPolygonNORI private _bridgedPolygonNORI;
-  Market private _market;
-
-  function run() external {
-    uint256 buyerPrivateKey = vm.envUint("BUYER_PRIVATE_KEY");
-    address buyerAddress = vm.addr(buyerPrivateKey);
+  SignatureUtils private immutable _signatureUtils = new SignatureUtils();
```
scottynomad, line 24: fwiw I was spelling out the permit implementation explicitly because I was trying to make this script as free of dependencies as possible since it serves as a simple implementation for partners who might want to integrate.  I think this is probably still a good cleanup, they can always pull in that helper code.

script/BuyNRT.s.sol
```diff
@@ -1,123 +1,61 @@
...
+  Market private immutable _market =
+    Market(
+      vm.parseJsonAddress(
+        vm.readFile(string.concat(vm.projectRoot(), "/contracts.json")),
+        ".mumbai.Market.proxyAddress"
+      )
+    );
 
-    vm.startBroadcast(buyerAddress);
+  uint256 private immutable _tonnes = vm.envUint("TONNES") * 1 ether;
```
scottynomad, line 34: Can we use the terminology NRTs rather than tonnes?

scottynomad (changes requested):

PR #559, opened 2023-03-02
add task for validating proxy

jaycenhorton:
### Example usage

```
yarn hardhat upgrade Removal --network mumbai 
```

> Warning: Potentially unsafe deployment of contracts/Removal.sol:Removal
You are using the `unsafeAllow.delegatecall` flag.                

Would have loved to do this with foundry somehow


amiecorso (approved):
nice

PR #560, opened 2023-03-02
use named imports

jaycenhorton:
I've see dramatic improvements in vscode + solidity using explicit imports. Verified that this does not impact upgradeability using 

```
yarn hardhat upgrade Removal --network mumbai --validate
```


amiecorso (approved):
nice -- IIRC this was noted as a best practice in the foundry document for other reasons too so thanks for doing this

PR #561, opened 2023-03-02
validate removal transfer amount

jaycenhorton:
**tl;dr: Enforce that removals can never be created/transferred/etc in a granularity smaller than 1e14**

Quoting/paraphrasing a comment in slack from @amiecorso 

> A Macro audit recently pointed out some issues with out initial changes related to handling the differences in precision between removals and USDC
> 
> Initially, we introduced a barrier to enforce that purchase amounts are never too small to be able to handle with the purchasing token. For example, if the purchasing token is USDC and we only have 6 decimals to work with (as well as performing some fractional math around fees and potentially holdback percentages that need two digits of precision of their own to use), then the smallest possible purchase amount can only use 4 possible decimals (such as 0.0001 NRTs) in order to perform all payouts and calculations without any loss of precision.
> 
> However, Macro noted that when fulfilling an order, it is possible that the order will be fulfilled with more than one removal - potentially pulling from multiple removals (and multiple suppliers) while assembling the total purchase amount. This means that the payouts and calculations — which happen on a per-removal basis since these removals potentially each belong to different suppliers — may be applied to amounts that are actually smaller than the enforced smallest safe purchase amount.
> 
> For a little deeper explanation: There is nothing in place enforcing that removal tokens don’t have information in low-order digits. So for example, let’s say that we mint/list a removal with a balance of 1,123,456,789,000,000,000 (which is how we’d represent 1.1234…… NRTs) and a second removal with a balance of 100 NRTs (100,000,000,000,000,000,000). An order comes through for 1 NRT (a beautifully valid amount!). It will use some of that first removal, leaving it with an available balance of 123,456,789,000,000,000 (0.1234…. NRTs). Then we get another order for 1 NRT. This order will use up the amount in the first removal, and then start to dip into the second removal, so it will have amounts of 123,456,789,000,000,000 and 876,543,210,000,000,000. Note that there are 8 significant digits in these values, and when they are used to make USDC-related calculations, we will lose precision and won’t be able to represent token amounts that are granular enough to accurately represent what is owed to the various parties — information will be chopped off and lost forever.
>
> An initial possible solution we thought of for this issue is to represent removal tokens (an ERC1155 token) with 6 decimals to match the USDC purchasing token in precision. However, unlike ERC20s, ERC1155s don’t have decimals as part of their interface, and the standard uses 18 decimals across the board. Even if we could take this approach, which we can’t, there was a thought that we may not want to limit our ability to represent super fractional amounts of NRTs in the future, when the super fractional NORI token is used for purchases, because it would preclude our ability to sell micro amounts of NRTs as part of certain platform integrations — a use case that has been talked about with potential partners.
>
> The approach we are currently prototyping is this: Even though removal tokens can’t be represented with 6 decimals in a formal code-enforced way like an ERC20, it turns out that we don’t actually currently work with removal amounts that use such granular precision. We represent removal amounts as grams in our datastore, which effectively use 6 decimals, and even then, I believe they effectively use 4 decimals since we’re never representing NRT amounts below something like 1.1234 NRTs (it might be even less precise?) which would translate to something like 1,123,400 grams (4 decimals of precision — the two zeros on the end are not significant). If listed removal balances never effectively use more than 4 decimals of precision, and if purchase amounts are always limited to amounts that do not use more than 4 decimals of precision, then while eating through supply in the round robin data structure to fulfill an order, we simply will never actually encounter a scenario where we have unsafely fractional removal amounts being used to fulfill an order.
>
> We propose adding safeguards to the entry points through which removals enter the market for sale to ensure that balances that are listed for sale are always treated as if the removal token uses 4 decimals of precision — the same granularity as valid purchase amounts themselves. Since order fulfillment allocation never uses division or other operations that would further subdivide amounts, it means that we would be able to fulfill an order without ever creating unsafely fractional amounts, as long as the balances are safely imprecise in the first place.


test/RestrictedNORI.test.ts
```diff
@@ -143,32 +143,6 @@ describe('RestrictedNORI', () => {
           )
       ).to.be.reverted;
     });
-    it('should revert if a restriction schedule is being created for a methodology/version that does not have a duration set', async () => {
```
jaycenhorton, line 146: Test failed so I just migrated it to forge instead

test/RestrictedNORI.t.sol
```diff
@@ -60,6 +60,16 @@ contract RestrictedNORI_createSchedule is UpgradeableMarket {
       methodologyVersion: 1
     });
   }
+
+  function test_RevertWhen_MethodologyVersionHasNoDurationSet() external {
```
jaycenhorton, line 64: Migrated from hardhat

contracts/Removal.sol
```diff
@@ -945,4 +945,30 @@ contract Removal is
...
+   *
+   * - If the recipient is the Market or the Certificate, the amount must be divisible by 1e14 (100,000,000,000,000).
+   * - If the recipient is neither the Market nor the Certificate the amount may also be zero.
+   */
+  function _isValidTransfer(uint256 amount, address to)
+    internal
+    view
+    returns (bool)
+  {
+    return
```
jaycenhorton, line 969: This ended up being a bit more complex than the original as it also accounts for 0 value removals

contracts/Removal.sol
```diff
@@ -945,4 +945,30 @@ contract Removal is
       revert InvalidData();
     }
   }
+
+  /**
+   * @notice Check if the amount and recipient constitute a valid transfer.
+   * @dev Ensure that the amount of tokens in circulation always multiples of 1e14.
+   *
+   * ##### Examples:
```
rlfarman, line 953: Thanks for including these examples!

contracts/Removal.sol
```diff
@@ -945,4 +945,30 @@ contract Removal is
...
+   * - `_isValidTransfer({amount: 1e14, to: address(1)}) == true`
+   * - `_isValidTransfer({amount: 0, to: address(_certificate)}) == true`
+   * - `_isValidTransfer({amount: 1, to: address(1)}) == false`
+   * - `_isValidTransfer({amount: 1e14 - 1, to: address(_market)}) == false`
+   *
+   * ##### Requirements:
+   *
+   * - If the recipient is the Market or the Certificate, the amount must be divisible by 1e14 (100,000,000,000,000).
```
amiecorso, line 961:
```suggestion
   * - If the recipient is the Market or the Certificate, the amount must be divisible by 1e14 (100,000,000,000,000) and non-zero.
```

rlfarman (approved):
LGTM

amiecorso (approved):

scottynomad (approved):

PR #562, opened 2023-03-02
fix certificate amount validation logic

jaycenhorton:
Fixes a bug in the market where it changes the smallest allowed amount for NORI from 100 to 1000, opposite of the comment

image


amiecorso:
I wonder if a good test case that would have detected this issue is to make sure we have a test case for the smallest valid amount, and then also an amount that is too small by exactly one significant figure in the low order direction

jaycenhorton:
> I wonder if a good test case that would have detected this issue is to make sure we have a test case for the smallest valid amount, and then also an amount that is too small by exactly one significant figure in the low order direction

totally-- that's what shot me into the rabbit hole for #561 where I wanted a better pattern for internal function tests. Was going to fuzz it as well.

PR #571, opened 2023-03-03
bump 721a dep

jaycenhorton:
supersedes #511

- [ ] run upgradeability check
- [ ] bump gas report


contracts/Certificate.sol
```diff
@@ -286,6 +286,18 @@ contract Certificate is
...
+   * @dev Override to disable ERC721 operator approvals, since certificate tokens are non-transferable.
+   */
+  function approve(address, uint256)
+    public
+    payable
+    override(ERC721AUpgradeable, IERC721AUpgradeable)
+  {
+    revert FunctionDisabled();
+  }
+
```
amiecorso, line 300: is this related to the upgrade, or something that was an oversight by us originally?

amiecorso (approved):

contracts/Certificate.sol
jaycenhorton: I think there function signature changed from pure to payable in the latest version which in also necessitated a reordering for solhint

amiecorso: oh I see - I actually hadn't noticed the deleted version of this same function, that totally makes sense

PR #575, opened 2023-03-06
add full audit history to readme

jaycenhorton:


amiecorso (approved):
nice thanks for doing this.  I clicked all the links and they all look good

PR #578, opened 2023-03-17
Update README.md

tsaopei:


amiecorso (approved):
this looks good, thanks Patrick! a lot out of date in here wow

jaycenhorton (changes requested):
you need to delete the [footnotes at the end of the file as well ](https://github.com/nori-dot-eco/contracts/blob/master/docs/README.md?plain=1#L166)

docs/README.md
```diff
@@ -107,7 +108,7 @@ A contract managing supplier insurance holdbacks.
 Operates similarly to _LockedNORI_ by acting as a wrapper token that governs the scheduled release of the underlying _BridgedPolygonNori_ asset. Implemented as an ERC1155, each token ID has its own schedule parameters that control the linear release of the underlying assets over the duration of the schedule. It is possible to create more than one schedule per owner address. It is not currently
 possible to transfer restricted balances between addresses, though this feature may be implemented in a future release.
 
-### Support Libraries
+## Support Libraries
```
jaycenhorton, line 111: I think this was intended to live under `## Contracts Overview`
```suggestion
### Support Libraries
```
:point_up: would need to relocate for this suggestion to be valid

PR #582, opened 2023-03-21
[NO-2817] Enable contracts to specify custom fee in permissioned endpoint

amiecorso:
Overloads `swapWithoutFee` and `swapFromSupplierWithoutFee` to allow the specification of a custom fee collected off-chain, for reporting purposes in the `CreateCertificate` event.


contracts/Market.sol
```diff
@@ -908,6 +914,60 @@ contract Market is
...
+   * @param recipient The address to which the certificate will be issued.
+   * @param purchaser The address that will pay for the removals and has granted approval to this contract
+   * to transfer their ERC20 tokens.
+   * @param amount The total purchase amount in ERC20 tokens. This is the total number of removals being
+   * purchased, scaled by the price multiple.
+   * @param customFee The custom fee percentage that was paid to Nori, as an integer, specified here for
+   * inclusion in emitted events.
+   */
+  function swapWithoutFee(
+    // TODO should we rename this or allow `swapWithoutFee` to be overloaded? (same question applies to swapWithoutFeeFromSupplier)
```
rlfarman, line 954: I'd say rename

contracts/Market.sol
```diff
@@ -908,6 +914,60 @@ contract Market is
...
+        recipient: recipient,
+        countOfRemovalsAllocated: countOfRemovalsAllocated,
+        ids: ids,
+        amounts: amounts,
+        suppliers: suppliers
+      })
+    });
+  }
+
+  /**
```
jaycenhorton, line 929: regen docs

test/checkout.int.t.sol
```diff
@@ -961,3 +961,141 @@ contract Checkout_buyingWithAlternateERC20_floatingPointPriceMultiple is
...
+  function test() external {
+    vm.startPrank(owner);
+    vm.recordLogs();
+    _market.swapWithoutFee(owner, owner, certificateAmount, customFee);
+    vm.stopPrank();
+
+    Vm.Log[] memory entries = vm.getRecordedLogs();
+    bool containsCreateCertificateEventSelector = false;
+    for (uint256 i = 0; i < entries.length; ++i) {
+      if (entries[i].topics[0] == CREATE_CERTIFICATE_EVENT_SELECTOR) {
```
jaycenhorton, line 996: this test is problematic because it can pass without ever testing anything if this selector isn't found. The entry index should be deterministic. Can you locate which one it is and use it statically?

test/checkout.int.t.sol
```diff
@@ -961,3 +961,141 @@ contract Checkout_buyingWithAlternateERC20_floatingPointPriceMultiple is
...
+      certificateAmount,
+      _namedAccounts.supplier,
+      customFee
+    );
+    vm.stopPrank();
+
+    Vm.Log[] memory entries = vm.getRecordedLogs();
+    bool containsCreateCertificateEventSelector = false;
+    for (uint256 i = 0; i < entries.length; ++i) {
+      if (entries[i].topics[0] == CREATE_CERTIFICATE_EVENT_SELECTOR) {
```
jaycenhorton, line 1068: this test is problematic because it can pass without ever testing anything if this selector isn't found. The entry index should be deterministic. Can you locate which one it is and use it statically?

test/checkout.int.t.sol
```diff
@@ -961,3 +961,141 @@ contract Checkout_buyingWithAlternateERC20_floatingPointPriceMultiple is
...
+      owner,
+      owner,
+      certificateAmount,
+      _namedAccounts.supplier,
+      customFee
+    );
+    vm.stopPrank();
+
+    Vm.Log[] memory entries = vm.getRecordedLogs();
+    bool containsCreateCertificateEventSelector = false;
```
jaycenhorton, line 1066: never used fyi

contracts/Market.sol
```diff
@@ -102,6 +103,7 @@ contract Market is
    */
   struct FulfillOrderData {
     bool chargeFee;
+    uint256 feePercentage;
```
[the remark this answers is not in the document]
amiecorso: we can't because `chargeFee` is used to predicate logic in the order fulfillment, but we often have scenarios (cash and special orders) where we need to specify what fee percentage was used off-chain (like the custom fee option we're introducing here) but _don't actually charge that fee during the on-chain purchase_ 

so we often have a scenario where `chargeFee` is false but the fee we're reporting in the emitted event is non-zero.  kinda weird, but this is how patrick and business folks want it so that we can have accurate on-chain accounting for fees that were collected

test/checkout.int.t.sol
amiecorso: oh yeah... the test I took this from makes assertions after the loop that the certificate event selector flag was tripped but I must have deleted that. i'll try the static index approach though as it's more clear

test/checkout.int.t.sol
```diff
@@ -1063,39 +1070,41 @@ contract Checkout_buyingFromSingleSupplierWithCustomFee is Checkout {
...
-        assertEq(priceMultiple, _market.getPriceMultiple());
-        assertEq(noriFeePercentage, customFee);
-        assertEq(removalIds.length, 1);
-        assertEq(removalAmounts.length, 1);
-        assertEq(removalIds[0], _removalIds[0]);
-        assertEq(removalAmounts[0], certificateAmount);
-      }
-    }
+    uint256 createCertificateEventIndex = 6;
+    assert(
```
jaycenhorton, line 1074: same question here: why assert vs asserteq?

amiecorso: no reason just wasn't paying attention! fixed

jaycenhorton (commented):
can you add a new gas report? `yarn  snapshot` otherwise lgtm

jaycenhorton (approved):
can you add a new gas report? `yarn  snapshot` otherwise lgtm

PR #589, opened 2023-03-31
diff for audit 3

jaycenhorton:


contracts/Market.sol
```diff
@@ -454,15 +454,14 @@ contract Market is
...
-    );
     _removal.safeBatchTransferFrom({
       from: address(this),
       to: address(_certificate),
       ids: removalIds,
       amounts: removalAmounts,
-      data: data
+      data: abi.encode(
+        true // isReplacement
+      )
```
amiecorso, line 464: inlined this to avoid stack too deep error

contracts/Market.sol
```diff
@@ -799,21 +788,20 @@ contract Market is
     bytes32 r,
     bytes32 s
   ) external whenNotPaused {
-    uint256 certificateAmount = this
-      .calculateCertificateAmountFromPurchaseTotal({purchaseTotal: amount});
```
amiecorso, line 803: Previously, the swap endpoints received the total amount of currency needed for the purchase, in units of the purchasing token, and would back out the certificate amount from this value based on the price multiple and the marketplace fee.  Now, for simplicity and to avoid errors/misunderstandings on the part of the buyer, the `amount` value passed to any purchasing endpoint is the number of NRTs to buy, and all other values/amounts owed to the various parties are calculated from that.

contracts/Market.sol
```diff
@@ -1063,26 +1041,72 @@ contract Market is
...
+
+  /**
+   * @notice Convert an amount of removals into an equivalent amount expressed in the purchasing token's decimals.
+   * @dev If the purchasing token's decimals is not 18, we need to convert the `removalAmount` (which is expressed with
+   * 18 decimals) to a unit that is expressed in the purchasing token's decimals. For example, if `removalAmount` is
+   * 1 ether (18 decimals) and the purchasing token's decimals is 6, the return value would be 1,000,000.
+   * @param removalAmount The amount of removals to express in the purchasing token's decimals.
+   * @return The amount of purchasing tokens required to purchase the specified amount of removals.
+   */
+  function convertRemovalDecimalsToPurchasingTokenDecimals(
```
jaycenhorton, line 1064: @amiecorso

PR #599, opened 2023-04-03
Upgrade Market contract on Mumbai with special orders endpoints

amiecorso:


config/solidity.ts
```diff
@@ -28,7 +28,7 @@ const PRODUCTION_SOLC_CONFIG: SolcUserConfig = {
     viaIR: true,
     optimizer: {
       enabled: true,
-      runs: 4980,
+      runs: 3000,
```
amiecorso, line 31: had to reduce optimizer runs to make the Market contract bytecode, which now has additional endpoints for special orders, small enough to be deployable.

this change in optimizer settings is also why the other contracts were upgraded -- because their bytecode changed too

deploy/deploy-market.ts
```diff
@@ -35,7 +35,7 @@ deploy.dependencies = [
   'RestrictedNORI',
 ];
 if (hre.network.name !== 'polygon') {
-  deploy.dependencies = [...deploy.dependencies, 'NoriUSDC'];
```
amiecorso, line 38: mumbai relies on real testnet USDC now

jaycenhorton (approved):

plugins/index.ts
```diff
@@ -184,10 +184,8 @@ const deployOrUpgradeProxy = async <
         }) as InstanceOfContract<TContract>;
       }
     } catch (error) {
-      hre.trace(`Failed to upgrade ${contractName} with error:`, error);
-      contract = hre.deployments.get(
-        contractName
-      ) as InstanceOfContract<TContract>;
+      hre.log(`Failed to upgrade ${contractName} with error:`, error);
+      throw new Error(`Failed to upgrade ${contractName} with error: ${error}`);
```
amiecorso, line 188: Throwing here is actually important because otherwise a failed upgrade/deployment will still lead to deployments being saved/written using the artifacts from the failed upgrade, thus subsequent attempts to upgrade will think that there has been no change and therefore no need to upgrade.  We shouldn't write deployment info if something went wrong during the deployment. @scottynomad  just flagging you here in case you are aware of a reason why it would be problematic to throw here instead of swallowing this error.

PR #602, opened 2023-04-04
Update monorepo

amiecorso:


jaycenhorton:
@amiecorso why is this PR needed?

amiecorso:
this fixes tests that started failing after we removed a deployment dependency on `NoriUSDC` and failed to commit the resulting `contracts.json` -- adding this updated file back to the contracts repo seems to fix the tests that were failing and will allow us to bump the contracts repo commit to master in the monorepo

jaycenhorton:
I see

jaycenhorton (approved):

PR #606, opened 2023-04-14
refactor removal before transfer hook

jaycenhorton:
Refactors the beforeTransfer hook to make things a little bit more readable.


contracts/Removal.sol
```diff
@@ -836,20 +836,18 @@ contract Removal is
...
+        (!isValidTransfer &&
+          to != RemovalIdLib.supplierAddress({removalId: id}))
+      ) {
         revert ForbiddenTransfer();
       }
       if (to == market) {
         _currentMarketBalance += amount;
-      }
-      if (from == market) {
+      } else if (from == market) {
```
jaycenhorton, line 848: this is the only thing that should be functionally different. It shouldn't ever be possible to have both the to and from address be the market contract. If that's not the case I'll need to change this

rlfarman (approved):

contracts/Removal.sol
amiecorso: even if to and from were able to be the same address and a transfer successfully executed in that way, would that ever functionally matter? it would still never be able to create an overly granular balance because no balance would change

amiecorso (approved):

PR #627, opened 2023-05-11
[NO-2654] upgrade to node 18.15

jaycenhorton:


.github/workflows/foundry.yml
```diff
@@ -25,7 +25,7 @@ jobs:
             ${{ runner.os }}-yarn-
       - uses: actions/setup-node@v3
         with:
-          node-version: 14
+          node-version-file: '.nvmrc'
```
kyleclark171, line 28: 👍

test/Market.t.sol
```diff
@@ -1909,7 +1903,6 @@ contract Market_validates_certificate_amount is UpgradeableUSDCMarket {
         numberOfNRTsToPurchase,
         _namedAccounts.supplier
       );
-      vm.stopPrank();
```
kyleclark171, line 1912: Are these being removed b/c these `stopPrank()` is being used unnecessarily? 

ex if like 1901|1895 used `vm.startPrank(...)` we would need this line, but since it doesn't, the call to `vm.stopPrank()` is doing nothing?

jaycenhorton: Correct, and in fact it failed the test when in clouded because there was no prank running. Not sure if this is due to a new forge version or something else, but either way it was always doing nothing.

PR #636, opened 2023-05-31
fix invariant test

jaycenhorton:
Not sure why, but having the invariant handler in a separate file broke invariant tests


amiecorso (approved):
that's very weird -- these tests definitely used to run, but I was able to repro the error without these changes and this does seem to fix it -- thanks!

PR #646, opened 2023-06-03
bump ts version and misc fixes

jaycenhorton:


.eslintrc.js
```diff
@@ -1,5 +1,8 @@
-const { parserOptions } = require('@nori-dot-com/eslint-config-nori/rules');
+const {
+  parserOptions,
+} = require('@nori-dot-com/eslint-config-nori/dist/settings/parser-options');
```
rlfarman, line 3: Copying comment from PR on monorepo:
> Why is this being imported from dist? Can we not expose this in through the @nori-dot-com/eslint-config-nori index file so we can import it properly?

rlfarman (changes requested):

.eslintrc.js
jaycenhorton: You can't import from a nested path like this AFAIK without using either:  tsconfig paths (which have to be set up for each importing project-- but can become problematic for ts-node as more recent versions don't have full support) or by using native esm (which we don't use yet). Further, this is a JS file, which can't use the tsconfig paths approach.

If there's a magical way to do this that I'm missing, I'd love to know about it

jaycenhorton: I've spent many hours on this problem before, but it always boils down to tsconfig paths or esm, and esm is the only stable solution but we're far away from supporting

jaycenhorton: https://josselinbuils.me/blog/stop-using-index-files

PR #669, opened 2023-08-08
`Market.sol` add option for custom price in special order transaction

amiecorso:
address [NO-2766]

Adds the option to specify a custom price in the special order endpoint that we use to fulfill special orders.

[NO-2766]: https://norinauts.atlassian.net/browse/NO-2766?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


test/checkout.int.t.sol
```diff
@@ -991,12 +992,13 @@ contract Checkout_buyingWithCustomFee is Checkout {
...
       certificateAmount,
-      customFee
+      customFee,
+      customPriceMultiple
     );
     vm.stopPrank();
 
     Vm.Log[] memory entries = vm.getRecordedLogs();
-    uint256 createCertificateEventIndex = 6;
+    uint256 createCertificateEventIndex = 7;
```
amiecorso, line 1001: now there's also a "SetPriceMultiple" event (or whatever it's actually called) happening before the creation of the certificate

contracts/Market.sol
```diff
@@ -949,12 +943,14 @@ contract Market is
    * purchased, scaled by the price multiple.
    * @param customFee The custom fee percentage that was paid to Nori, as an integer, specified here for
    * inclusion in emitted events.
+   * @param customPrice The custom price that will be charged for this transaction.
    */
   function swapWithoutFeeSpecialOrder(
```
amiecorso, line 948: real changes are in this function -- otherwise it's all linting

.gas-snapshot
```diff
@@ -19,42 +19,48 @@ Certificate_supportsInterface:test() (gas: 5159)
...
-Checkout_buyingFromTenRemovals_singleSupplier_byApproval:test() (gas: 1963977)
-Checkout_buyingFromTenRemovals_singleSupplier_withoutFee:test() (gas: 1708460)
-Checkout_buyingFromTenRemovals_withoutFee:test() (gas: 1719464)
-Checkout_buyingFromTenSuppliers:test() (gas: 2855260)
-Checkout_buyingWithAlternateERC20:test() (gas: 544655)
-Checkout_buyingWithAlternateERC20_floatingPointPriceMultiple:test() (gas: 544655)
-Checkout_buyingWithCustomFee:test() (gas: 547446)
-Checkout_swapWithDifferentPermitSignerAndMsgSender:test() (gas: 726838)
+Checkout_buyingFromOneRemoval:test() (gas: 725948)
+Checkout_buyingFromOneRemoval_byApproval:test() (gas: 684373)
```
jaycenhorton, line 23: Was the original gas report stale? Is that why we're seeing so many changes here?

If so, can you update the gas report on master first?

rlfarman (dismissed):
I like it!

contracts/Market.sol
```diff
@@ -1078,6 +1079,8 @@ contract Market is
         certificateAmount: amount,
         supplier: supplier
       });
+    uint256 currentPrice = _priceMultiple;
+    _priceMultiple = customPriceMultiple;
```
jaycenhorton, line 1083: I think we should move this to the top of the function to reduce future risk of the prior functions introducing changes that rely on this value

contracts/Market.sol
```diff
@@ -963,6 +959,8 @@ contract Market is
       uint256[] memory amounts,
       address[] memory suppliers
     ) = _allocateRemovals({purchaser: purchaser, certificateAmount: amount});
+    uint256 currentPrice = _priceMultiple;
```
jaycenhorton, line 962: I think we should move this to the top of the function to reduce future risk of the prior functions introducing changes that rely on this value

contracts/Market.sol
amiecorso: good idea

.gas-snapshot
```diff
@@ -19,48 +19,48 @@ Certificate_supportsInterface:test() (gas: 5159)
...
-MarketGasExplorer:test_4_singleSupplierGranularAmounts() (gas: 3629646788)
-MarketGasExplorer:test_5_uniqueSuppliersGranularAmounts() (gas: 5791445583)
-MarketInvariantTest:invariant_callSummary() (runs: 400, calls: 6000, reverts: 4273)
-MarketInvariantTest:invariant_sumOfPurchaseAmounts() (runs: 400, calls: 6000, reverts: 4273)
-MarketSupplierSelectionNotUsingUpSuppliersLastRemoval:test() (gas: 922133)
-Market_ALLOWLIST_ROLE:test() (gas: 12799)
-Market_USDC_swap_respects_decimal_mismatch:test() (gas: 716909)
-Market_USDC_swap_withholds_restricted_nori:test() (gas: 892028)
+MarketGasExplorer:test_0_bestPossibleCaseOneRemoval() (gas: 29769122)
+MarketGasExplorer:test_1_mostRealisticSeveralSuppliersModerateAmounts() (gas: 49820714)
```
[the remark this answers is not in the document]
amiecorso: dude.  I have no idea.  there are small changes and improvements all over the place, including for tests that should not have been affected by these changes (since we only made changes to two specific endpoints that only have one test each) 

this makes me think that gas costs are affected by the overall layout of the contract's bytecode, which will have changed a little bit by the change in the implementation of those two endpoints.  that's all I can guess.  maybe chatGPT will know more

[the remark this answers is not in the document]
jaycenhorton: 1) what command did you run to generate the report?
2) Is the last point in this [readme](https://github.com/Rubilmax/foundry-gas-diff#%EF%B8%8F-known-limitations) relevant? I dont think it is, but I'm not positive.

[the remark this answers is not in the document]
jaycenhorton: I think you're right about the layout. See [this article](https://medium.com/joyso/solidity-how-does-function-name-affect-gas-consumption-in-smart-contract-47d270d8ac92)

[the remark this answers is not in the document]
amiecorso: i'm not sure.... this sentence from the last point in the readme is confusing to me "Average & median gas costs for each function is estimated based on the test suite, which means they are easily impacted by small changes in the tests."  (it's just so abstract)

I just used `forge snapshot` to generate the report

.gas-snapshot
```diff
@@ -19,48 +19,48 @@ Certificate_supportsInterface:test() (gas: 5159)
...
+Checkout_buyingFromSingleSupplierWithCustomFeeAndPrice:test() (gas: 553566)
+Checkout_buyingFromTenRemovals:test() (gas: 1975477)
+Checkout_buyingFromTenRemovals_singleSupplier:test() (gas: 1964590)
+Checkout_buyingFromTenRemovals_singleSupplier_byApproval:test() (gas: 1962590)
+Checkout_buyingFromTenRemovals_singleSupplier_withoutFee:test() (gas: 1707603)
+Checkout_buyingFromTenRemovals_withoutFee:test() (gas: 1718584)
+Checkout_buyingFromTenSuppliers:test() (gas: 2853764)
+Checkout_buyingWithAlternateERC20:test() (gas: 544459)
+Checkout_buyingWithAlternateERC20_floatingPointPriceMultiple:test() (gas: 544459)
+Checkout_buyingWithCustomFeeAndPrice:test() (gas: 563667)
```
jaycenhorton, line 33: would have been nice to have this test defined as the function name inside the market contract tests. Had to lookup which one of the tests was relevant manually.

diff: +~3%, (at gas price of 100: 0.001622 MATIC, $0.001134). Not as bad as I thought it would be, though still something to call out.

.gas-snapshot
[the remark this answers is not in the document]
jaycenhorton: I think you might have some compiler options set in your env. I get a different report on this branch when running `yarn snapshot`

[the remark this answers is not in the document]
amiecorso: Screenshot 2023-08-10 at 1 12 28 PM
hmm, so I have this in my .env currently -- is this different from what you're working with? (presumably I should be running with the production values for this report?)

rlfarman (dismissed):

test/tasks/vesting.test.ts
```diff
@@ -35,16 +35,24 @@ const MOCK_VESTING_HEADER = [
   'lastQuantityRevoked',
 ];
 
+const currentDate = new Date();
```
jaycenhorton, line 38: were these changes still needed?

test/tasks/vesting.test.ts
```diff
@@ -35,16 +35,24 @@ const MOCK_VESTING_HEADER = [
...
+);
+const currentDatePlusOne = new Date(currentDate.getTime() + 1);
+
 const MOCK_GITHUB_VESTING_GRANTS = [
   [
     '0xDD66B46910918B2F442D6b75C6E55631ad678c99',
     '4c728e75-0aef-42f6-8ee3-3a831734e048',
     '46425.5886',
-    '2023-07-14T00:00:00Z',
+    dateLessThanOneYearAgoFromToday.toISOString(),
```
jaycenhorton, line 51: could probably do toISOString once when initializing the value so we dont need it in each test

test/tasks/vesting.test.ts
```diff
@@ -121,11 +129,11 @@ const MOCK_GITHUB_VESTING_GRANTS = [
     '0xBd6E6A75c7A51cfdf08DDf2f538ceB221835839b',
     'daf2922b-c96d-442e-a267-e977758961f1',
     '50000',
-    '2024-03-03T00:00:00Z',
-    '2028-03-03T00:00:00Z',
+    '2023-03-03T00:00:00Z',
+    '2027-03-03T00:00:00Z',
```
jaycenhorton, line 133: why dont these ones line up?

test/tasks/vesting.test.ts
```diff
@@ -373,8 +381,8 @@ describe('vesting task', () => {
         it('should pass if the argument passes isValidEvmMoment validation', () => {
           for (const v of [
             utcToEvmTime(moment()),
-            utcToEvmTime('2023-03-21T20:45:29.496Z'),
-            utcToEvmTime('2023-03-21'),
+            utcToEvmTime('2022-03-21T20:45:29.496Z'),
```
jaycenhorton, line 384: why was this needed?

test/tasks/vesting.test.ts
```diff
@@ -873,12 +944,21 @@ describe('vesting task', () => {
...
           '0x6b9d03759E9F14a641f0703fBD84F1F726159B6B': {
-            vestEndTime: { __old: 0, __new: 1_689_292_800 },
-            startTime: { __old: 0, __new: 1_689_292_800 },
-            originalAmount: { __old: '0', __new: '20000000000000000000000000' },
-            unlockEndTime: { __old: 0, __new: 1_847_145_600 },
-            cliff1Time: { __old: 0, __new: 1_720_915_200 },
-            cliff2Time: { __old: 0, __new: 1_736_812_800 },
+            vestEndTime: {
+              __old: 0,
+              __new: utcToEvmTime(currentDate.getTime()),
```
jaycenhorton, line 949: seems like these could also be converted once and assigned as a variable

contracts/Market.sol
```diff
@@ -1078,6 +1081,7 @@ contract Market is
         certificateAmount: amount,
         supplier: supplier
       });
+
```
jaycenhorton, line 1084:
```suggestion
```

test/tasks/vesting.test.ts
amiecorso: I just wanted to update these tests so that they're more robust against time passing and had already made the changes on my branch so figured incorporate them

amiecorso: I just wanted to update these tests so that they're more robust against time passing and had already made the changes on my branch so figured incorporate them

test/tasks/vesting.test.ts
amiecorso: merge conflict between the bumping of dates that happened on your node20 PR and the way I went about fixing the tests -- this shouldn't matter because we're just testing that the grant schema is valid with these, and it still is with these dates as well because they are in the future from now

test/tasks/vesting.test.ts
amiecorso: also a merge conflict issue with bumped dates -- switched these up to 2023

test/Market.t.sol
```diff
@@ -1842,6 +1850,7 @@ contract Market_validates_certificate_amount is UpgradeableUSDCMarket {
 
   function test() external {
     uint256 ownerPrivateKey = 0xA11CE;
+    uint256 priceMultiple = _market.getPriceMultiple();
```
amiecorso, line 1853: this has to be set here as a variable otherwise the  `expectRevert` gets applied to the `getPriceMultiple` call as it is the next call that occurs

test/checkout.int.t.sol
```diff
@@ -962,26 +964,28 @@ contract Checkout_buyingWithAlternateERC20_floatingPointPriceMultiple is
...
+    uint256 purchaseAmount = _market.calculateCheckoutTotalWithoutFee({
+      amount: certificateAmount,
+      priceMultiple: customPriceMultiple
+    });
     _market.grantRole({role: _market.MARKET_ADMIN_ROLE(), account: owner});
     vm.prank(_namedAccounts.admin);
     _bpNori.deposit(owner, abi.encode(purchaseAmount));
     vm.prank(owner);
-    _bpNori.approve(address(_market), MAX_INT); // infinite approval for Market to spend owner's tokens
+    _bpNori.approve(address(_market), purchaseAmount);
```
amiecorso, line 988: I wanted to tighten up this logic and discovered that when the approval amount is `MAX_INT` then the approval balance isn't updated upon token transfer.  this means that in the case that the approval amount is `MAX_INT` there are two fewer events that are emitted in the flow of one of our purchases, because two `Approval` events that otherwise would have been emitted (one on transfer to the supplier and one on transfer to the rNORI contract.... and in a standard swap scenario a THIRD approval event upon transfer to the fee wallet) do not get emitted.  this changes the index of the event signatures we care about.  this one was a bit of a stumper for a while so figured I'd share here

test/checkout.int.t.sol
```diff
@@ -991,12 +995,13 @@ contract Checkout_buyingWithCustomFee is Checkout {
...
       certificateAmount,
-      customFee
+      customFee,
+      customPriceMultiple
     );
     vm.stopPrank();
 
     Vm.Log[] memory entries = vm.getRecordedLogs();
-    uint256 createCertificateEventIndex = 6;
+    uint256 createCertificateEventIndex = 8;
```
amiecorso, line 1004: the `Approval` event note above is why this index changed when I approved a specific amount instead of MAX_INT, both here and in the other test

test/checkout.int.t.sol
```diff
@@ -962,26 +964,28 @@ contract Checkout_buyingWithAlternateERC20_floatingPointPriceMultiple is
   }
 }
 
-contract Checkout_buyingWithCustomFee is Checkout {
+contract Checkout_swapWithoutFeeSpecialOrder is Checkout {
   uint256 ownerPrivateKey = 0xA11CE;
   address owner = vm.addr(ownerPrivateKey);
   uint256 customFee = 5;
   uint256 certificateAmount = 1 ether;
+  uint256 customPriceMultiple = 1800; // $18.00 -- test below the default price multiple of $20.00
```
amiecorso, line 972: made sure one test with a custom price was deviating from default price multiple with a lower price, and the other test deviates higher

.gas-snapshot
```diff
@@ -19,42 +19,48 @@ Certificate_supportsInterface:test() (gas: 5159)
...
-MarketSupplierSelectionNotUsingUpSuppliersLastRemoval:test() (gas: 922133)
-Market_ALLOWLIST_ROLE:test() (gas: 12799)
-Market_USDC_swap_respects_decimal_mismatch:test() (gas: 716909)
-Market_USDC_swap_withholds_restricted_nori:test() (gas: 892028)
+MarketGasExplorer:test_0_bestPossibleCaseOneRemoval() (gas: 29769122)
+MarketGasExplorer:test_1_mostRealisticSeveralSuppliersModerateAmounts() (gas: 49820714)
+MarketGasExplorer:test_2_uniqueSuppliersBulkAmounts() (gas: 78231132)
+MarketGasExplorer:test_3_uniqueSuppliersVariableAmounts() (gas: 1808717194)
+MarketGasExplorer:test_4_singleSupplierGranularAmounts() (gas: 3985218350)
+MarketGasExplorer:test_5_uniqueSuppliersGranularAmounts() (gas: 6532333334)
```
amiecorso, line 46: I merged up with master and I thought these "MarketGasExplorer" tests would be omitted after those changes so i'm not sure why they're appearing in this snapshot

contracts/Market.sol
```diff
@@ -1232,17 +1234,17 @@ contract Market is
...
    */
-  function calculateCheckoutTotalWithoutFee(uint256 amount)
-    external
-    view
-    returns (uint256)
-  {
+  function calculateCheckoutTotalWithoutFee(
+    uint256 amount,
+    uint256 priceMultiple
+  ) external view returns (uint256) {
```
amiecorso, line 1243: this is a new change -- I realized that externally to the contract (but nowhere internally!) this function is used to calculate the total purchase amount in ERC-20 tokens that is needed for an `approve` call.  Therefore this needs to be parameterized with the price being used for the transaction since we need to be able to calculate total purchase amount for the special order scenario (currently the only scenario in which this `withoutFee` calculation is relevant).  It's a simple enough calculation to do without the help of a contract function, because there is no fee involved, so what you're really asking is "how many NRTs am I buying and at what price?" and multiplying those two values together.  So it's almost worth removing this code from the contract entirely.  However, the reason it should stay and it's a little more complicated than NRTs * price is because of the purchasing token decimals conversion.... ultimately, the output of this function gives you the purchase total in ERC20 _in the correct decimal expression_ for the token being used.... so I think this has to stay and it has to allow specification of the price multiple directly.

test/Market.t.sol
```diff
@@ -1580,10 +1583,15 @@ contract Market_convertPurchasingTokenDecimalsToRemovalDecimals is
 
 contract Market_calculates_prices_using_decimal is UpgradeableUSDCMarket {
   function test() external {
-    // expectRevert(_market.calculateCheckoutTotal(1), 0);
```
rlfarman, line 1583: Why is this commented out?

rlfarman (approved):
Uncomment or remove comment in test and then LGTM

test/Market.t.sol
jaycenhorton: doesnt seem like it was ever valid anyways

test/Market.t.sol
jaycenhorton: probably belongs in the setup function anyways

test/checkout.int.t.sol
jaycenhorton: that's interesting. We should track that behavior in a test somewhere. (not for this pr)

jaycenhorton (approved):

PR #670, opened 2023-08-10
Update gas snapshot

amiecorso:
snapshot was stale -- this can serve as a baseline for comparing upcoming contracts changes' effects on gas


.github/workflows/foundry.yml
```diff
@@ -29,7 +29,7 @@ jobs:
       - name: Install Foundry
         uses: foundry-rs/foundry-toolchain@v1
         with:
-          version: nightly
+          version: nightly-d4f626bb7f96d46358997d4b27f79358cb2b3401
```
jaycenhorton, line 33: Can you create a jira issue for this?

jaycenhorton (approved):

.github/workflows/foundry.yml
amiecorso: good idea

PR #677, opened 2023-08-11
[NO-3155] Fix usage of pranks in  functions to make suite work with foundry@nightly

jaycenhorton:
These changes fixes our integration with the latest versions of foundry. It seems something changed within the foundry stack that broke some of our test suites. The problematic test suites were all starting pranks in `setUp` functions without also calling `stopPrank` before the `setUp` function ended. It's unclear exactly why this broke (i suspect it's related to the changes in the revm dependency that was recently updated, but that's just a guess). Nonetheless it's a lot cleaner and easier to debug if we are using prank in more limited contexts (otherwise it can be hard to figure out what address is the currently active address, and this is especially true given that setup and test functions are initiated from two separate contexts IIRC).

The primary fix is outlined in [this comment](https://github.com/nori-dot-eco/contracts/pull/677#discussion_r1290898622)

I created a new issue in the foundry repo [here](https://github.com/foundry-rs/foundry/issues/5598)

### DIFF
--- a/test/Market.t.sol
+++ b/test/Market.t.sol
@@ -97,7 +97,7 @@ contract Market_swap_revertsWhenUnsafeERC20TransferFails is UpgradeableMarket {
   }
 
   function test() external {
-    vm.startPrank(owner);
+    vm.prank(owner);
     vm.expectRevert(ERC20TransferFailed.selector);
     _market.swap(
       owner,
@@ -108,7 +108,6 @@ contract Market_swap_revertsWhenUnsafeERC20TransferFails is UpgradeableMarket {
       signedPermit.r,
       signedPermit.s
     );
-    vm.stopPrank();
   }
 }
 
@@ -216,7 +215,6 @@ contract Market_replace is MarketReplaceTestHelper {
   function setUp() external {
     _listRemovals();
     _createCertificate();
-
     _market.grantRole({
       role: _market.MARKET_ADMIN_ROLE(),
       account: _namedAccounts.admin
@@ -227,13 +225,14 @@ contract Market_replace is MarketReplaceTestHelper {
     });
     uint256 amount = _market.calculateCheckoutTotal(_amountToReplace);
     vm.startPrank(_namedAccounts.admin);
-    _removal.release(_removalIds[0], _originalCertificateAmount);
-    assertEq(_certificate.getNrtDeficit(), _originalCertificateAmount);
     _bpNori.deposit(_namedAccounts.admin, abi.encode(amount));
     _bpNori.approve(address(_market), amount);
+    _removal.release(_removalIds[0], _originalCertificateAmount);
+    vm.stopPrank();
   }
 
   function test() external {
+    assertEq(_certificate.getNrtDeficit(), _originalCertificateAmount);
     vm.expectEmit(true, true, false, true);
     emit UpdateCertificate(
       _certificateTokenId,
@@ -244,6 +243,7 @@ contract Market_replace is MarketReplaceTestHelper {
       address(_bpNori),
       _market.getPriceMultiple()
     );
+    vm.prank(_namedAccounts.admin);
     _market.replace({
       treasury: _namedAccounts.admin,
       certificateId: _certificateTokenId,
@@ -255,7 +255,6 @@ contract Market_replace is MarketReplaceTestHelper {
       _certificate.getNrtDeficit(),
       _originalCertificateAmount - _amountToReplace
     );
-    vm.stopPrank();
   }
 }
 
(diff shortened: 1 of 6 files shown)


test/Market.t.sol
```diff
@@ -227,13 +225,14 @@ contract Market_replace is MarketReplaceTestHelper {
...
-    _removal.release(_removalIds[0], _originalCertificateAmount);
-    assertEq(_certificate.getNrtDeficit(), _originalCertificateAmount);
     _bpNori.deposit(_namedAccounts.admin, abi.encode(amount));
     _bpNori.approve(address(_market), amount);
+    _removal.release(_removalIds[0], _originalCertificateAmount);
+    vm.stopPrank();
   }
 
   function test() external {
+    assertEq(_certificate.getNrtDeficit(), _originalCertificateAmount);
```
jaycenhorton, line 235: I moved this from the `setUp` function because assertions in the `setUp` use generic error messages if they fail, whereas within a test function you get the context of the failure (e.g., expected true to be false, etc).

test/Market.t.sol
```diff
@@ -265,37 +264,22 @@ contract Market_replace_reverts_ReplacementAmountExceedsNrtDeficit is
...
-    });
-    uint256 amount = _market.calculateCheckoutTotal(_amountToReplace);
-    vm.startPrank(_namedAccounts.admin);
-    assertEq(_certificate.getNrtDeficit(), 0); // no discrepancy
-    _bpNori.deposit(_namedAccounts.admin, abi.encode(amount));
-    _bpNori.approve(address(_market), amount);
   }
 
   function test() external {
-    // using low level call to get around limitation of vm.expectRevert
```
jaycenhorton, line 285: @amiecorso what was the original context behind why this was needed here?

test/Market.t.sol
```diff
@@ -227,13 +225,14 @@ contract Market_replace is MarketReplaceTestHelper {
     });
     uint256 amount = _market.calculateCheckoutTotal(_amountToReplace);
     vm.startPrank(_namedAccounts.admin);
-    _removal.release(_removalIds[0], _originalCertificateAmount);
-    assertEq(_certificate.getNrtDeficit(), _originalCertificateAmount);
     _bpNori.deposit(_namedAccounts.admin, abi.encode(amount));
     _bpNori.approve(address(_market), amount);
+    _removal.release(_removalIds[0], _originalCertificateAmount);
+    vm.stopPrank();
```
jaycenhorton, line 231: To reproduce the original issue, simply remove this line of code and you'll get a panick from revm

```
The application panicked (crashed).                                                                                                                        
Message:  called `Option::unwrap()` on a `None` value                        
Location: ~/.cargo/git/checkouts/revm-bf744d8ffabcbad9/bc4d203/crates/revm/src/evm_impl.rs:267  
```

test/Market.t.sol
```diff
@@ -265,37 +264,22 @@ contract Market_replace_reverts_ReplacementAmountExceedsNrtDeficit is
...
     });
-    _removal.grantRole({
-      role: _removal.RELEASER_ROLE(),
-      account: _namedAccounts.admin
-    });
-    uint256 amount = _market.calculateCheckoutTotal(_amountToReplace);
-    vm.startPrank(_namedAccounts.admin);
-    assertEq(_certificate.getNrtDeficit(), 0); // no discrepancy
-    _bpNori.deposit(_namedAccounts.admin, abi.encode(amount));
-    _bpNori.approve(address(_market), amount);
```
jaycenhorton, line 281: removed several of these blocks of code as they were not needed in all test contexts

test/Market.t.sol
```diff
@@ -97,7 +97,7 @@ contract Market_swap_revertsWhenUnsafeERC20TransferFails is UpgradeableMarket {
   }
 
   function test() external {
-    vm.startPrank(owner);
+    vm.prank(owner);
```
jaycenhorton, line 100: switched to using `vm.prank` in several places to minimize the number of places that the msg.sender deviates from the expected global default value.

test/Market.t.sol
amiecorso: interesting.... so the limitation of `vm.expectRevert` is that it applies to the _very next function call_ -- so if you're wanting to know whether `replace` is going to revert, that becomes a problem if there are other function calls that happen in the execution of `replace` before the revert you expect (and if those *don't* revert, then `vm.expectRevert` will fail).  the approach here uses the low-level call to get a success/failure code, and then checks that directly -- so it's just a hackier way to ask the question you're trying to ask without relying on `vm.expectRevert`

I'm not sure what changed to allow it to work with `expectRevert` now -- maybe they fixed the limitation of `vm.expectRevert`? (that'd be cool) maybe something in our code changed?

amiecorso: maybe here?: https://github.com/foundry-rs/foundry/pull/4945

amiecorso (approved):
nice work figuring this out

amiecorso:
I resolved all the conversations because I wanted this to automerge so I can pull it in to my priceMultiple PR but I left some responses

test/Market.t.sol
jaycenhorton: problem with the hacky workaround was that we weren't actually testing what caused the revert which made the test fragile and the title misleading. Useful hack, but we should be careful if we need to rely on anything like it in the future and make sure we aren't getting a revert for other reasons.

PR #679, opened 2023-08-11
Require market admin role for swapFromSupplier

rlfarman:
Needs updated snapshot


contracts/Market.sol
```diff
@@ -791,7 +785,7 @@ contract Market is
     uint8 v,
     bytes32 r,
     bytes32 s
-  ) external whenNotPaused {
+  ) external whenNotPaused onlyRole(MARKET_ADMIN_ROLE) {
```
rlfarman, line 788: This is the only change

jaycenhorton: what's the purpose of allowlist_role when we permission based on this too?

rlfarman: Allowlist will still allow you to buy from the restricted threshold pool.
If we decide to keep this functionality and allow buyers direct access to it, then obviously we'd use a role other than MARKET_ADMIN_ROLE. But I also think that role should be distinct from the ALLOWLIST_ROLE.

rlfarman: Relatedly, we could consider removing `_validatePrioritySupply` from `_allocateRemovalsFromSupplier` since it's now safe to assume we only ever allocate removals from a specific supplier when the caller already has the MARKET_ADMIN role. Would reduce gas cost. That's also assuming we don't ever want to enforce the `ALLOWLIST_ROLE` check on someone who has the `MARKET_ADMIN_ROLE`, which seems unnecessary.

But again, at some point in the future I anticipate introducing a third role for allowing buyers to buy from specific suppliers, and then we'd have to put this check back in if we removed it here and now.

jaycenhorton: RE roles: I think it's useful to have more granular roles so we can, for example, allow third party wallet/contract interpretations.  

RE _validatePrioritySupply: if it makes sense to remove that, that would be great. Not sure if my prior point negates the feasibility of that or not. I do find it odd/suboptimal that we check supply + priority supply in different ways across different functions (e.g., `_allocateRemovals` vs `_allocateSupplySingleSupplier`)

rlfarman: [Here's a draft PR ](https://github.com/nori-dot-eco/contracts/pull/680)with the changes needed.
It has the added benefit of not making me to fix the tests so that we have an admin account who does not also have the `ALLOWLIST_ROLE` (if I'm correctly understanding why that test is failing)!

jaycenhorton: There's also a second `swapFromSupplier` you need to permission, I believe.

jaycenhorton (changes requested):
https://github.com/nori-dot-eco/contracts/pull/679/files#r1291990675

test/Certificate.test.ts
```diff
@@ -35,13 +35,10 @@ describe('Certificate', () => {
     await expect(
       market
         .connect(buyer)
-        [
-          'swapFromSupplier(address,address,uint256,address,uint256,uint8,bytes32,bytes32)'
-        ](
+        ['swap(address,address,uint256,uint256,uint8,bytes32,bytes32)'](
```
jaycenhorton, line 38: you can just delete this test since we have one in foundry covering it [here](https://github.com/nori-dot-eco/contracts/blob/7046e8446201ffb20664b74dc152ea709308a17f/test/checkout.int.t.sol#L806)

test/Market.test.ts
```diff
@@ -1137,50 +1137,49 @@ describe('Market', () => {
...
-                },
-              ],
-            },
+    const { bpNori, market, totalAmountOfSupply } = await setupTest({
+      userFixtures: {
+        supplier: {
+          removalDataToList: {
+            removals: [
+              {
+                amount: formatTokenAmount(10),
```
jaycenhorton, line 1146: you can also delete this one I believe (see [here](https://github.com/nori-dot-eco/contracts/blob/7046e8446201ffb20664b74dc152ea709308a17f/test/checkout.int.t.sol#L409))

.gas-snapshot
```diff
@@ -86,12 +86,12 @@ Market_setPriorityRestrictedThreshold:test() (gas: 157404)
...
+Market_supplierSelectionUsingUpSuppliersLastRemoval:test() (gas: 918444)
 Market_swapWithoutFee_emits_and_skips_transfer_when_transferring_wrong_erc20_to_rNori:test() (gas: 419376)
-Market_swap_emits_and_skips_transfer_when_transferring_wrong_erc20_to_rNori:test() (gas: 519931)
-Market_swap_emits_event_and_skips_mint_when_minting_rNori_to_nonERC1155Receiver:test() (gas: 597111)
-Market_swap_revertsWhenUnsafeERC20TransferFails:test() (gas: 216998)
-Market_validates_certificate_amount:test() (gas: 645513)
+Market_swap_emits_and_skips_transfer_when_transferring_wrong_erc20_to_rNori:test() (gas: 519930)
+Market_swap_emits_event_and_skips_mint_when_minting_rNori_to_nonERC1155Receiver:test() (gas: 597110)
+Market_swap_revertsWhenUnsafeERC20TransferFails:test() (gas: 216997)
+Market_validates_certificate_amount:test() (gas: 664841)
```
jaycenhorton, line 94: Was surprised to see this big of a diff here. I'm guessing it's related to the fact that we've stuffed a bunch of setup logic into the test function. Emphasizes that we need too make sure we are putting as much of that kind of thing in `setUp` as possible, otherwise this report becomes much less useful


diff: 19328 (2.994%)

contracts/Market.sol
jaycenhorton: Add this to the list of requirements in the natspec
 ```
- Can only be used when the caller has the `MARKET_ADMIN_ROLE` role.
```

jaycenhorton (approved):
left one comment about updating the natspec, otherwise lgtm

jaycenhorton (approved):

PR #680, opened 2023-08-11
Remove priority supply validation check

rlfarman:


jaycenhorton (changes requested):
https://github.com/nori-dot-eco/contracts/pull/679/files#r1291990675

contracts/Market.sol
```diff
@@ -853,7 +852,6 @@ contract Market is
       uint256[] memory amounts,
       address[] memory suppliers
     ) = _allocateRemovalsFromSupplier({
-        purchaser: _msgSender(),
```
jaycenhorton, line 856: double checking: does this break the changes introduced [here](https://github.com/nori-dot-eco/contracts/pull/493#issue-1537320601)

PR #682, opened 2023-08-12
Fix coverage reporting for `ArrayLib`

jaycenhorton:
This PR fixes coverage reporting for `ArrayLib`. Similar changes are required to fix reports for other libraries.

See [here](https://github.com/foundry-rs/foundry/pull/3128) for why this was needed

### DIFF
--- a/test/helpers/array-lib.sol
+++ b/test/helpers/array-lib.sol
@@ -1,0 +1,73 @@
+/* solhint-disable contract-name-camelcase, func-name-mixedcase */
+// SPDX-License-Identifier: MIT
+pragma solidity =0.8.17;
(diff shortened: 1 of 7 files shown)


test/helpers/array-lib.sol
amiecorso, line 1: I wonder if we shouldn't name this file with some notion of "harness" ?

amiecorso (approved):

test/helpers/array-lib.sol
jaycenhorton: keeping it as is for now to follow the existing pattern (e.g., we don't call `market.sol` `UpgradeableMarket.sol`)

PR #688, opened 2023-08-18
Fulfill special orders from set of valid vintages and/or specific supplier

amiecorso:
Addresses [NO-3136]

Allows passing an array of vintages from which to select removals when fulfilling a special order. If an empty array is passed, the regular fulfillment algorithm is used. Similarly, a specific supplier can be specified, but if the zero address is passed here, any supplier can be used.

This PR consolidates special order endpoints because treating all possible combinations of specification with different overloaded endpoints with different args will lead to an unmanageable proliferation of code (and bloat the contract size). Specifically, our permissioned endpoints need to accommodate the possibility of specifying a custom price, a custom fee, a valid set of vintages and a specific supplier.  Even just dealing with vintages and supplier already, we could be in position of having 4 different permissioned swap endpoints: regular fulfillment, constrained vintage fulfillment, constrained supplier fulfillment, or both constrained vintages and supplier fulfillment.  I propose here to consolidate these endpoints into one endpoint that we use for special orders in which all fields must be specified, but an empty array of vintages and the zero address can be used to indicate unconstrained fulfillment in terms of vintages and supplier, respectively.  Similarly, with price and fee specifications, we weren't using the permissioned endpoints that didn't allow specification of these values.  To fulfill a special order with our permissioned endpoint, we should expect to provide an intentional value for every single one of these options.  The different allocation flows required based on what is specified are implemented in their own respective functions and the right allocation to use is selected based on the arguments supplied to the swap function.

[NO-3136]: https://norinauts.atlassian.net/browse/NO-3136?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


slither.config.json
```diff
@@ -10,6 +10,5 @@
   ],
   "exclude_dependencies": true,
   "compile_force_framework": "hardhat",
-  "ignore_compile": true,
-  "skip_clean": true
```
amiecorso, line 14: slither doesn't recognize this key

slither.db.json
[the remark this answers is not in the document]
jaycenhorton: amazing-- this was actually on my todo list as well

slither.config.json
jaycenhorton: are you sure? I explicitly remember adding this. It's passed through to cryptic cimpile if im not mistaken and removes a redundant step.

https://github.com/crytic/crytic-compile/blob/master/crytic_compile/crytic_compile.py#L552

jaycenhorton: https://github.com/crytic/slither/discussions/1866#discussioncomment-5740034

contracts/Market.sol
```diff
@@ -1610,6 +1369,89 @@ contract Market is
...
+    uint256 availableSupply = _removal.getMarketBalance();
+    _validateSupply({
+      certificateAmount: certificateAmount,
+      availableSupply: availableSupply
+    });
+    _validatePrioritySupply({
+      purchaser: purchaser,
+      certificateAmount: certificateAmount,
+      availableSupply: availableSupply
+    });
```
rlfarman, line 1442: This step is unnecessary so long as this function is only exposed for the `MARKET_ADMIN_ROLE`

contracts/Market.sol
```diff
@@ -1610,6 +1369,89 @@ contract Market is
...
+      uint256[] memory ids,
+      uint256[] memory amounts,
+      address[] memory suppliers
+    )
+  {
+    uint256 availableSupply = _removal.getMarketBalance();
+    _validateSupply({
+      certificateAmount: certificateAmount,
+      availableSupply: availableSupply
+    });
```
rlfarman, line 1437: Any reason why this is the only `_allocateRemovals...` function that calls `validateSupply`?

contracts/Market.sol
```diff
@@ -941,144 +772,72 @@ contract Market is
...
         supplier: supplier
       });
+    }
+    // case no restrictions on fulfillment
+    else {
+      (countOfRemovalsAllocated, ids, amounts, suppliers) = _allocateRemovals({
+        purchaser: purchaser,
+        certificateAmount: amount
+      });
+    }
```
rlfarman, line 840: Similar to your approach to combining these multiple endpoints into a singular function, you might clean up this function (and help unify shared logic between each of the `_allocateRemovals()` functions) by using a singular `_allocateRemovals` function that takes a supplier and vintages as optional arguments, performs shared logic (IE - `validateSupply`, handles the `if... else if... else` logic, and then returns the list of removal IDs. Just a thought.

contracts/Market.sol
```diff
@@ -1929,6 +1771,168 @@ contract Market is
...
+    amounts = new uint256[](countOfListedRemovals);
+    suppliers = new address[](countOfListedRemovals);
+    countOfRemovalsAllocated = 0;
+    for (uint256 i = 0; i < countOfListedRemovals; ++i) {
+      uint256 removalId = _listedSupply[_currentSupplierAddress]
+        .getNextRemovalForSaleFromVintages({vintages: vintages});
+      // if removalId is 0 here, this supplier doesn't have any removals from the vintages requested
+      // at that point we should simply move on to the next supplier
+      if (removalId == 0) {
+        _incrementCurrentSupplierAddress();
```
rlfarman, line 1810: Won't this push the pointer of the round robin order to the next supplier? I don't think purchasing from specific vintages (and jumping outside of the queue as ordered) should modify the state of the queue. Can we store this value and reset it at the end of the function (unless we otherwise detect that we legitimately need to increment the supplier)?

.gas-snapshot
```diff
@@ -20,103 +20,106 @@ Certificate_supportsInterface:test() (gas: 5159)
...
-Checkout_buyingFromTenSuppliers:test() (gas: 2853763)
-Checkout_buyingWithAlternateERC20:test() (gas: 545223)
-Checkout_buyingWithAlternateERC20_floatingPointPriceMultiple:test() (gas: 545223)
-Checkout_swapFromSupplierWithoutFeeSpecialOrder:test() (gas: 559047)
-Checkout_swapWithDifferentPermitSignerAndMsgSender:test() (gas: 727284)
-Checkout_swapWithoutFeeSpecialOrder:test() (gas: 564508)
+Checkout_buyingFromOneRemoval:test() (gas: 726740)
+Checkout_buyingFromOneRemoval_byApproval:test() (gas: 685153)
+Checkout_buyingFromTenRemovals:test() (gas: 1975024)
+Checkout_buyingFromTenRemovals_singleSupplier:test() (gas: 1712779)
```
jaycenhorton, line 26: interesting that this went down (significantly)

contracts/Market.sol
```diff
@@ -941,144 +772,72 @@ contract Market is
...
-    _fulfillOrder({
-      params: FulfillOrderData({
-        chargeFee: false,
-        feePercentage: customFee,
+    uint256 countOfRemovalsAllocated;
+    uint256[] memory ids;
+    uint256[] memory amounts;
+    address[] memory suppliers;
+    // case vintage-specific fulfillment only
+    if (vintages.length > 0 && supplier == address(0)) {
```
jaycenhorton, line 799: possible optimization: store vintages.length as a variable above this line

contracts/Market.sol
```diff
@@ -941,144 +772,72 @@ contract Market is
...
-      uint256 countOfRemovalsAllocated,
-      uint256[] memory ids,
-      uint256[] memory amounts,
-      address[] memory suppliers
-    ) = _allocateRemovals({purchaser: purchaser, certificateAmount: amount});
-    _fulfillOrder({
-      params: FulfillOrderData({
-        chargeFee: false,
-        feePercentage: customFee,
+    uint256 countOfRemovalsAllocated;
```
jaycenhorton, line 794: why can this be different than ids.length?

contracts/Market.sol
```diff
@@ -941,144 +772,72 @@ contract Market is
...
-    _priceMultiple = customPriceMultiple;
-    _validateCertificateAmount({amount: amount});
-    (
-      uint256 countOfRemovalsAllocated,
-      uint256[] memory ids,
-      uint256[] memory amounts,
-      address[] memory suppliers
-    ) = _allocateRemovalsFromSupplier({
+      // case vintage-specific fulfillment and supplier-specific fulfillment
+    } else if (vintages.length > 0 && supplier != address(0)) {
```
jaycenhorton, line 822: It seems like this could have been a ternary in the first if statement if you remove the address(0) check. Did you try that?

contracts/Market.sol
```diff
@@ -941,144 +772,72 @@ contract Market is
...
+        suppliers
+      ) = _allocateRemovalsFromSupplierSpecificVintages({
         certificateAmount: amount,
+        vintages: vintages,
         supplier: supplier
       });
+    }
+    // case no restrictions on fulfillment
+    else {
+      (countOfRemovalsAllocated, ids, amounts, suppliers) = _allocateRemovals({
```
jaycenhorton, line 836: consider making this the first possible block of code if it's the happy path to reduce checking multiple if conditions

contracts/Market.sol
```diff
@@ -1929,6 +1771,168 @@ contract Market is
...
+        id: removalId
+      });
+      if (remainingAmountToFill < removalAmount) {
+        /**
+         * The order is complete, not fully using up this removal, don't check about removing active supplier.
+         */
+        ids[countOfRemovalsAllocated] = removalId;
+        amounts[countOfRemovalsAllocated] = remainingAmountToFill;
+        suppliers[countOfRemovalsAllocated] = _currentSupplierAddress;
+        remainingAmountToFill = 0;
```
jaycenhorton, line 1824: These if/else blocks seem like the could share more code, no? Especially if you can just do something like removalAmount - min(remainingAmountToFill, removalAmount)

contracts/Market.sol
```diff
@@ -1929,6 +1771,168 @@ contract Market is
...
+   * @param supplier The supplier from which to purchase carbon removals.
+   * @param vintages The vintages from which to purchase carbon removals.
+   * @return countOfRemovalsAllocated The number of distinct removal IDs used to fulfill this order.
+   * @return ids An array of the removal IDs being drawn from to fulfill this order.
+   * @return amounts An array of amounts being allocated from each corresponding removal token.
+   */
+  function _allocateSupplySingleSupplierSpecificVintages(
+    uint256 amount,
+    address supplier,
+    uint256[] memory vintages
```
jaycenhorton, line 1864: Double check that any of these new params can't be changed to calldata

contracts/Market.sol
```diff
@@ -1929,6 +1771,168 @@ contract Market is
...
+  )
+    private
+    returns (
+      uint256 countOfRemovalsAllocated,
+      uint256[] memory ids,
+      uint256[] memory amounts
+    )
+  {
+    RemovalsByYear storage supplierRemovalQueue = _listedSupply[supplier];
```
jaycenhorton, line 1873: Remind me why this needs to be storage?

contracts/Market.sol
```diff
@@ -1929,6 +1771,168 @@ contract Market is
...
+    uint256 countOfSuppliersListedRemovals = 0;
+    for (
+      uint256 vintage = supplierRemovalQueue.earliestYear;
```
jaycenhorton, line 1876: Assign variable outside of loop

contracts/Market.sol
```diff
@@ -1929,6 +1771,168 @@ contract Market is
...
+      vintage <= supplierRemovalQueue.latestYear;
```
jaycenhorton, line 1877: Assign variable outside of loop

contracts/Market.sol
```diff
@@ -1929,6 +1771,168 @@ contract Market is
...
+      ++vintage
+    ) {
+      countOfSuppliersListedRemovals += supplierRemovalQueue
```
jaycenhorton, line 1880: Could it be optimized if we used array.fill (filled with lengths) + array.sum. 

Not sure if this is possible with our current implementation of the two array methods as this may require a callback definition

E.g. something like this  https://github.com/siromivel/lambdeth (note: don't use this exact library if you go this route as I believe I've seen an audited/optimized variant elsewhere and I'm happy to dig it up if you need it, just let me know that you're going to go that route so I can post it here as a reply)

contracts/RemovalsByYearLib.sol
```diff
@@ -140,26 +139,44 @@ library RemovalsByYearLib {
...
 
+  /**
+   * @notice Gets the next removal in the collection for sale that comes from any of the specified vintage years.
+   * If no removal exists on the valid range, 0 is returned.
+   * @param collection the collection from storage.
+   * @param vintages the valid set of years from which to pull the removal ID.
+   * @return The next removal to sell, or 0 if no removal exists on that vintage range.
+   */
+  function getNextRemovalForSaleFromVintages(
+    RemovalsByYear storage collection,
```
jaycenhorton, line 156: Missing coverage

jaycenhorton: Might be because you need to refactor the lib tests like I did for arraylib?

contracts/Market.sol
```diff
@@ -1929,6 +1771,168 @@ contract Market is
...
+         * We will use up this removal while completing the order, move on to next one.
+         */
+        ids[countOfRemovalsAllocated] = removalId;
+        amounts[countOfRemovalsAllocated] = removalAmount; // this removal is getting used up
+        remainingAmountToFill -= removalAmount;
+        supplierRemovalQueue.remove({removalId: removalId});
+        /**
+         * If the supplier is out of supply, remove them from the active suppliers.
+         */
+        if (supplierRemovalQueue.isEmpty()) {
```
jaycenhorton, line 1921: Probably worth making sure this missing coverage is hit

jaycenhorton (changes requested):

contracts/Market.sol
jaycenhorton: Big +1 here

contracts/Market.sol
```diff
@@ -941,144 +772,72 @@ contract Market is
...
    */
   function swapWithoutFeeSpecialOrder(
     address recipient,
     address purchaser,
     uint256 amount,
     uint256 customFee,
-    uint256 customPriceMultiple
+    uint256 customPriceMultiple,
+    address supplier,
+    uint256[] memory vintages
```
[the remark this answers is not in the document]
amiecorso: I don't understand

contracts/Market.sol
amiecorso: at this point in the code `ids` is a really long array that as instantiated with as many possible removals as could be used to fulfill. it gets sliced later

contracts/Market.sol
amiecorso: yeah good idea -- can do that in the vintage>0 case and also the else clause to that

contracts/Market.sol
```diff
@@ -941,144 +772,72 @@ contract Market is
...
+        amounts,
+        suppliers
+      ) = _allocateRemovalsFromSupplierSpecificVintages({
         certificateAmount: amount,
+        vintages: vintages,
         supplier: supplier
       });
+    }
+    // case no restrictions on fulfillment
+    else {
```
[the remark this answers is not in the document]
amiecorso: yes actually -- and the only place I needed to add this was in the implementation of `allocateRemovalsSpecificVintages` ... in the happy path case this is taken care of by `validateSupply`, and in the supplier specific cases that logic already exists in the allocation logic because there is some pre-loop overhead that looks at the supplier's available removals ahead of time.  In this way, each of the 4 possible fulfillment paths uses the information that is efficiently available to it to abort as early as possible without doing a bunch of extra work.

contracts/Market.sol
```diff
@@ -1610,6 +1369,89 @@ contract Market is
     return (countOfRemovalsAllocated, ids, amounts, suppliers);
   }
 
+  /**
+   * @notice Allocates removals from a specific supplier to be fulfilled.
+   * @dev This function is responsible for validating and allocating the supply from a specific supplier.
+   * @param certificateAmount The total amount of NRTs for the certificate.
+   * @param supplier The only supplier address from which to purchase carbon removals in this transaction.
+   * @param vintages A set of valid vintages from which to allocate removals.
+   * @return countOfRemovalsAllocated The number of distinct removal IDs used to fulfill this order.
```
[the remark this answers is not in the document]
amiecorso: this function will revert if it's 0

contracts/Market.sol
amiecorso: yeah it's because validate supply is just looking at whether there is enough total supply in the market to fulfill the order. that's not very useful when looking at a per-supplier or vintage-specific fulfillment since you can only fulfill from a subset of the total market supply.  in those cases the checks for insufficient supply happen as early in the process as information permits without doing a bunch of extra work at the outset to come to the same conclusion that would be arrived at during the allocation anyway

contracts/Market.sol
```diff
@@ -1610,6 +1369,89 @@ contract Market is
...
+    internal
+    returns (
+      uint256 countOfRemovalsAllocated,
+      uint256[] memory ids,
+      uint256[] memory amounts,
+      address[] memory suppliers
+    )
+  {
+    uint256 availableSupply = _removal.getMarketBalance();
+    _validateSupply({
```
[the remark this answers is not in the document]
amiecorso: I added tests for the cases both in which there is no such supplier in the market, and where the specified vintages don't exist.  They both revert with `InsufficientSupply`.  The reason I don't think we should add validation before the actual fulfillment loop is that we will repeat work.  In both of these cases, we learn quite early in the fulfillment process whether the supplier has removals and/or whether there are removals available in the requested vintages.  For example, if there is no supplier or none of the requested vintages, we basically have some loops that never run (because they're iterating over 0 things) and then we conclude insufficient supply after doing just as much work as we would have to do to make this determination ahead of the fulfillment algorithm. A check ahead of time would simply duplicate the work of looking into those data structures immediately before doing it anyway.

contracts/Market.sol
amiecorso: love it

contracts/Market.sol
```diff
@@ -1929,6 +1771,168 @@ contract Market is
...
+    });
+    ids = new uint256[](countOfListedRemovals);
+    amounts = new uint256[](countOfListedRemovals);
+    suppliers = new address[](countOfListedRemovals);
+    countOfRemovalsAllocated = 0;
+    for (uint256 i = 0; i < countOfListedRemovals; ++i) {
+      uint256 removalId = _listedSupply[_currentSupplierAddress]
+        .getNextRemovalForSaleFromVintages({vintages: vintages});
+      // if removalId is 0 here, this supplier doesn't have any removals from the vintages requested
+      // at that point we should simply move on to the next supplier
```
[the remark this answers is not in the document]
amiecorso: I've explored this heavily in thinking about this algorithm, and I don't believe there is a better approach here for much the same reason that we have to loop in our regular fulfillment with the upper bound on iterations being the count of all available removal tokens for sale. our data structure is indexed first by supplier, then by vintage. there is overhead we could perform to figure out which suppliers have any of the valid vintages and how much they have which would be a tighter bound on our loop iterations, but it would require doing work during the overhead that we can actually potentially avoid by being lazier about it.  i.e. if we were to try to be smarter about this fulfillment, we still have to walk through our suppliers and examine their data structures to understand who has how much of which vintages we care about. but if we're lazier about it, we can basically start the same walk through the data structure with a high likelihood of NOT having to do the full walk that we would do if we were going to do some organizational overhead before performing fulfillment.  Here, we basically say, "maximum case is that all removal sin our marketplace meet the criteria and could be part of this fulfillment, so that's how we'll bound our fulfillment loop, but we'll look as efficiently as possible at each supplier and if they don't have any removal sfrom the vintages we care about, we'll move right on".  That's my best explanation.  If you have a specific idea for improving this I'm open it but I've thought about it a lot

contracts/Market.sol
amiecorso: yeah that's a good idea

amiecorso: i'm going to do this in a slightly different way... if all we do is keep track of the `_currentSupplierAddress` before the fulfillment algorithm so that we can reset it later, this supplier may be removed during fulfillment as well as any number of other intervening suppliers (but not necessarily all of them as this fulfillment pulls only from supplier with certain vintages).  I think it's easier to reason about it if we have a local `localCurrentSupplier` that starts as the global `_currentSupplierAddress` and then serves the same role.  the only tricky case here is that if we remove the `localCurrentSupplier` while removing the removal, we have to be responsible for making sure we move on to the next supplier locally by checking if the supplier we were on is now empty.  the `removeActiveRemoval` function does what we need it to in terms of keeping the global data structure and `_currentSupplierAddress` consistent which is great.

contracts/Market.sol
```diff
@@ -1929,6 +1771,168 @@ contract Market is
...
+      if (removalId == 0) {
+        _incrementCurrentSupplierAddress();
+        continue;
+      }
+      uint256 removalAmount = _removal.balanceOf({
```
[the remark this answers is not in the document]
amiecorso: we can't batch this call because we need to know how much we can take from this removal in the flow of fulfillment -- same reason we don't already batch this in the similar fulfillment flow for regular fulfillment

contracts/Market.sol
amiecorso: because this is a storage variable -- we need any mutations to this variable to persist in state

contracts/Market.sol
amiecorso: are you sure this would actually save gas to use an array when we don't actually need one in the first place?  I think it's more efficient to use array.sum if your data is in array in the first place and you need to sum it up, but otherwise we'll be performing unnecessary variable declarations and filling the array when really we're just adding a few numbers together and we are never going to be adding more than a small handful of numbers together (not more than the number of potential vintages supplied)

test/helpers/removal.sol
```diff
@@ -241,6 +241,35 @@ abstract contract UpgradeableRemoval is Upgradeable {
     return _removalIds;
   }
 
+  function _seedAndListRemoval(
```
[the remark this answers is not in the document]
amiecorso: yeah we have a lot of variations of this that have cropped up to serve various needs without eventually being revisiting and simplified.  I think we could simplify this by creating a helper function that has a few more parameters, but it is also the case that some of these overrides are helping to obscure some complexity while offering control over the few things that matter.

the new one that I added isn't so much about the listing (which yeah could be parameterized with a "to") but rather offering control over the exact vintage and amount with which a removal will be listed.

I think we could clean these up but given the time sensitivity of the audit, the complexity that's already present in this PR, and the fact that this is a test helper, I'm gonna push for doing this in a followup PR.

contracts/Market.sol
amiecorso: calldata only applies to external functions

contracts/Market.sol
amiecorso: this removes a lot of code, very satisfying

amiecorso: I'm going to leave one version of `_allocateRemovals` that is used in the basic fulfillment and then introduce `_allocateRemovalsSpecialOrder` to handle this complex case

contracts/Market.sol
jaycenhorton: Right, but this is passed in from an external function definition, no?

contracts/Market.sol
jaycenhorton: Nope, just calling it out as a possibility (specifically because if it's a similar or equal gas cost it could enable a functional abstraction/pattern for this kind of thing in many places in our code-- and that abstraction can be optimized in assembly like the rest of arraylib). Totally fine if it doesn't make sense, but wanted to draw attention to it in case it's worth thinking about while there's relevant context for it.

contracts/Market.sol
```diff
@@ -1610,6 +1369,89 @@ contract Market is
...
+    internal
+    returns (
+      uint256 countOfRemovalsAllocated,
+      uint256[] memory ids,
+      uint256[] memory amounts,
+      address[] memory suppliers
+    )
+  {
+    (
+      countOfRemovalsAllocated,
```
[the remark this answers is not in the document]
amiecorso: doin ittttt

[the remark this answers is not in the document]
amiecorso: it's actually a lot less verbose in sooo many places

contracts/Market.sol
amiecorso: interesting... I think this and the missing coverage above for `getNextRemovalForSaleFromVintages` both have to do with the fact that these are library functions that are part of the diff in this PR and thus getting picked up by codecov (whereas the many other library calls we use are not).  I believe that because I know there are tests that hit both of these lines for sure.  I wonder what the difference is between calling these functions from a test harness v.s. as part of actual contract code in regular tests? do you have any sense for that?  it's annoying

contracts/Market.sol
```diff
@@ -1929,6 +1771,168 @@ contract Market is
...
+      account: address(this)
+    });
+    ids = new uint256[](countOfListedRemovals);
+    amounts = new uint256[](countOfListedRemovals);
+    suppliers = new address[](countOfListedRemovals);
+    countOfRemovalsAllocated = 0;
+    for (uint256 i = 0; i < countOfListedRemovals; ++i) {
+      uint256 removalId = _listedSupply[_currentSupplierAddress]
+        .getNextRemovalForSaleFromVintages({vintages: vintages});
+      // if removalId is 0 here, this supplier doesn't have any removals from the vintages requested
```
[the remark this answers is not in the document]
amiecorso: this comment is actually part of the natspec for that function so i'm going to delete this comment that was just part of my pseudocoding

contracts/Market.sol
jaycenhorton: Yeah. I don't remember the exact reason why one is picked up and the other not, but there are several foundry coverage GitHub issues explaining it

contracts/Market.sol
amiecorso: Screenshot 2023-08-24 at 4 34 58 PM

contracts/Market.sol
jaycenhorton: - https://github.com/foundry-rs/foundry/pull/3128
- https://github.com/foundry-rs/foundry/issues/2567
Both of the above contain several related linked issues describing the issue and workarounds. It's how I learned to fix arraylib coverage. 

Not suggesting we add new tests and fix this issue here (assuming this is not net new code that is also uncovered by a correct test case using a harness), just documenting it.

contracts/Market.sol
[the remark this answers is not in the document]
jaycenhorton: Still could be useful to refactor to an internal function since there's a lot happening here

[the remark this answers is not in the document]
amiecorso: the only other activity here is setting the `localCurrentSupplier` to the next in the case that this supplier is out of removals... adding another internal function will add useless overhead and leave the same number of lines here because we still have to assign this local variable

contracts/Market.sol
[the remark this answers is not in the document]
jaycenhorton: If I'm understanding right, there's a special order purchase function and a non special order purchase function. Is it possible to also merge these two into just one? I realize the special order one uses this modifier to prevent unauthorized access for zero fee purchases, but perhaps we can validate/handle/determine that by using an if/else inside a single swap function

Does that make sense?

contracts/Market.sol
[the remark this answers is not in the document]
jaycenhorton: Oh, maybe I'm looking at a stale diff? This diff looks like several variables are updated

contracts/Market.sol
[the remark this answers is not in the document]
amiecorso: oh yeah that makes sense... interesting -- like basically just policing it based on whether the sender has the required role instead of using a modifier on the signature itself which gives us the same permission but more granular control.

the issue I see with this is that the signature of this mega-function would have to look like the special order purchase function signature, with all these extra parameters that have special meanings when they have special values such as the zero address or empty array (not to mention the complexity of specifying your price and fee!).  The complicated signature works for us because we know how to provide an intentional value for each field, but I don't believe the public-facing swap endpoint should have most of these parameters.

maybe one way to do it could be to refactor so that there's one internal swap function that basically looks like the current special order case, and then the public facing function wraps that and provides the arguments to the internal function:

```
...
customFee: getNoriFeePercentage();
priceMultiple: getPriceMultiple();
supplier: address(0);
vintages: new uint256[](0);
})
```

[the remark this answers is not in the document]
amiecorso: and the you also have a special order wrapper that has all of the parameters.  that still leaves 2 different functions though and honestly doesn't do much to simplify the guts and would increase gas costs for the public endpoint

contracts/Market.sol
jaycenhorton: https://github.com/nori-dot-eco/contracts/pull/691

contracts/Market.sol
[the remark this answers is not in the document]
jaycenhorton: Yeah basically what you're saying is what I was thinking (although, alternatively, I don't see why the public facing endpoint couldnt just require zero values as well since they're easy enough to specify-- especially if this at some point gets further simplified to handle things off chain using removal ID specification somehow instead as all you'd need to zero out is a single array).

Using your suggestion might still simplify the readability/audit ability as the two entry points simply pipe into the same internal function, and do nothing else otherwise other than set default values.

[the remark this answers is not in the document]
jaycenhorton: Likewise you could also combine the two swap functions in a similar way using either approach by inferring the path based on whether the value was a zero value or not.

But you might be right when you say we might not gain much if all we end up doing is create several if/else forks later on. Though maybe that's addressable with well written cases with enums (e.g., pseudocode: `checkoutType =  sender === admin ? Swap.SpecialOrder : v,r,s === 0 ? Swap.WithoutPermit : ... ` followed by if statements for each swap enums value... 🤷Super rough idea, not suggesting we do it now, but worth considering when thinking about where all this seems to be heading)

test/helpers/removal.sol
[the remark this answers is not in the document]
jaycenhorton: Some of the other ones also already have todos that suggest removing redundancies here 🙈

Directionally, and related to [discussions](https://github.com/nori-dot-eco/contracts/pull/688/files#r1305026609) on this PR about minimizing swap entry points, there should be one or two variations at most (one with all params and maybe another which has one param which uses a predefined fixture (e.g. an enum perhaps)). 

I'm fine with the decision to not address this now, but I'll also call out that this is the kind of thing that's easier to fix before it gets added into code than it is when ignored and taken as debt just to be removed later (as future code will likely continue adding dependencies to multiple variations, and rewriting tests across suites can be a delicate operation that requires context switching between test context)


Tl;Dr: a stitch in time saves nine

contracts/Market.sol
```diff
@@ -1574,40 +1334,14 @@ contract Market is
       certificateAmount: certificateAmount,
       availableSupply: availableSupply
     });
-    (countOfRemovalsAllocated, ids, amounts, suppliers) = _allocateSupply({
+    uint256 countOfRemovalsAllocated;
+    uint256[] memory ids;
+    uint256[] memory amounts;
+    address[] memory suppliers;
```
jaycenhorton, line 1340: These appear unused

contracts/Market.sol
```diff
@@ -1547,23 +1264,66 @@ contract Market is
...
-   * @return amounts An array of amounts being allocated from each corresponding removal token.
-   * @return suppliers The address of the supplier who owns each corresponding removal token.
+   * @param supplier The only supplier address from which to purchase carbon removals in this transaction,
+   * or zero address if any supplier is valid.
+   * @param vintages A set of valid vintages from which to allocate removals, empty if any vintage is valid.
+   * @return SupplyAllocationData The removals, amounts, suppliers and count data returned
+   * from the supply allocation algorithm.
+   */
+  function _allocateRemovalsSpecialOrder(
+    address purchaser,
```
jaycenhorton, line 1274:
```suggestion
    address,
```

contracts/Market.sol
```diff
@@ -428,20 +437,17 @@ contract Market is
...
-
-    uint256[] memory removalIds = ids.slice({
+    SupplyAllocationData memory supplyAllocationData = _allocateSupply({
+      amount: totalAmountToReplace
+    });
+
+    uint256[] memory removalIds = supplyAllocationData.ids.slice({
       from: 0,
-      to: countOfRemovalsAllocated
+      to: supplyAllocationData.countOfRemovalsAllocated
```
jaycenhorton, line 446: Save 20 gas by caching this. Not sure if it's worth it though

contracts/Market.sol
```diff
@@ -1509,21 +1221,26 @@ contract Market is
    * @param params The order fulfillment data.
    */
   function _fulfillOrder(FulfillOrderData memory params) internal {
-    uint256[] memory removalIds = params.ids.slice({
-      from: 0,
-      to: params.countOfRemovalsAllocated
-    });
-    uint256[] memory removalAmounts = params.amounts.slice({
+    uint256[] memory removalIds = params.supplyAllocationData.ids.slice({
```
jaycenhorton, line 1224: Maybe rename params to order?

contracts/Market.sol
```diff
@@ -1547,23 +1264,66 @@ contract Market is
...
+        });
+      }
+    } else {
+      // case vintage-specific fulfillment only
+      if (supplier == address(0)) {
+        supplyAllocationData = _allocateSupplySpecificVintages({
+          amount: certificateAmount,
+          vintages: vintages
+        });
+        // case vintage-specific fulfillment and supplier-specific fulfillment
```
jaycenhorton, line 1303: Consider moving these inside the blocks they relate to instead of at the end of the previous unrelated block

contracts/Market.sol
```diff
@@ -1776,30 +1510,20 @@ contract Market is
...
-  )
-    private
-    returns (
-      uint256 countOfRemovalsAllocated,
-      uint256[] memory ids,
-      uint256[] memory amounts,
-      address[] memory suppliers
-    )
-  {
+  ) private returns (SupplyAllocationData memory) {
```
jaycenhorton, line 1518: Consider making any private functions internal so we can test them in foundry. Rule of thumb applies globally

test/checkout.int.t.sol
```diff
@@ -455,7 +364,7 @@ contract Checkout_buyingFromTenRemovals_singleSupplier is Checkout {
   }
 }
 
-contract Checkout_buyingFromTenRemovals_singleSupplier_byApproval is Checkout {
```
jaycenhorton, line 458: Can't tell from this diff, but did this permit. Asked test just get moved elsewhere?

test/checkout.int.t.sol
```diff
@@ -1119,8 +1171,150 @@ contract Checkout_swapFromSupplierWithoutFeeSpecialOrder is Checkout {
...
+      owner,
+      certificateAmount,
+      customFee,
+      _priceMultiple,
+      _namedAccounts.supplier2,
+      vintages
+    );
+
+    Vm.Log[] memory entries = vm.getRecordedLogs();
+    // In this test scenario, the `CreateCertificate` event index is 18 because multiple suppliers are getting
```
jaycenhorton, line 1273: I think we should just use 18 statically instead of looping. If the test breaks because we added more events or changed logic otherwise, it's probably a good thing to be alerted about. 

Worried about this test not catching multiple certificate creation events, using 0 as the index,  etc otherwise. 

Thoughts?

test/checkout.int.t.sol
```diff
@@ -1119,8 +1171,150 @@ contract Checkout_swapFromSupplierWithoutFeeSpecialOrder is Checkout {
...
+      address from,
+      uint256 eventCertificateAmount,
+      uint256[] memory removalIds,
+      uint256[] memory removalAmounts,
+      uint256 priceMultiple,
+      uint256 noriFeePercentage
+    ) = abi.decode(
+        entries[createCertificateEventIndex].data,
+        (address, uint256, uint256[], uint256[], uint256, uint256)
+      );
```
jaycenhorton, line 1308: Should we just use expect emit instead of recorded logs for this test?

jaycenhorton: Get recorded logs would be more useful if this test tested more than just the one event

test/checkout.int.t.sol
```diff
@@ -1039,13 +946,157 @@ contract Checkout_swapWithoutFeeSpecialOrder is Checkout {
...
+  function setUp() external {
+    /**
+     * Supplier 1: 2018 vintage
+     * Supplier 2: 2018, 2019 and 2020 vintage
+     * Supplier 3: 2019 vintage
+     * All removals are 1 tonne
+     * Certificate amount is 2.5 tonnes
+     * Fulfillment should result in all of supplier2's removals being used, and .5 tonne of supplier3's removals
+     */
+    _removalIds.push(
```
jaycenhorton, line 972: Looking at how this new seed function is being used, it seems like this should have could have been more consistent if it extended this pattern, but also points out how there are too many ways to seed removals creating test setup to vary between tests. 🙈

https://github.com/nori-dot-eco/contracts/blob/a96d8d6413907b674a119cc1d53afe589831470f/test/checkout.int.t.sol#L195

test/checkout.int.t.sol
```diff
@@ -1119,8 +1171,150 @@ contract Checkout_swapFromSupplierWithoutFeeSpecialOrder is Checkout {
...
+      vintages: new uint256[](0)
+    });
+  }
+}
+
+contract Checkout_swapWithoutFeeSpecialOrder_specificVintagesSpecificSupplier is
+  Checkout
+{
+  uint256 ownerPrivateKey = 0xA11CE;
+  address owner = vm.addr(ownerPrivateKey);
```
jaycenhorton, line 1198: Consider using _namedAccounts instead. It has the added benefit of labeling the account with "owner" due to how they are set up when looking at foundry logs. Doing it this way is just going to render the address with no label and because you don't have any indication what Alice's public key is here it'll be near impossible to debug. If you are doing this elsewhere in this PR change that too.

test/checkout.int.t.sol
```diff
@@ -1119,8 +1171,150 @@ contract Checkout_swapFromSupplierWithoutFeeSpecialOrder is Checkout {
...
+  uint256 customFee = 5;
```
jaycenhorton, line 1199: Not: private, constants

jaycenhorton, line 1199: Nit: private, constant, leading underscore

test/checkout.int.t.sol
```diff
@@ -1119,8 +1171,150 @@ contract Checkout_swapFromSupplierWithoutFeeSpecialOrder is Checkout {
...
+  function test_revertsWhenSupplierDoesNotExistInMarket() external {
+    vm.prank(owner);
+    vm.expectRevert(InsufficientSupply.selector);
+    _market.swapWithoutFeeSpecialOrder({
+      recipient: owner,
+      purchaser: owner,
+      amount: certificateAmount,
+      customFee: customFee,
+      customPriceMultiple: customPriceMultiple,
+      supplier: _namedAccounts.supplier2, // not listed in market
```
jaycenhorton, line 1188: Given the shared setup function, and the ambiguity of insufficient supply errors (iirc correctly it can also revert with this error if there isn't enough supply vs if there isn't supply) it might be useful to either assert that it doesn't exist in the market first or use a non supplier named account (e.g. namedAccounts.buyer, or _account("non-existent supplier", address(f00)) to prevent accidentally adjusting the shared setup in the future

test/checkout.int.t.sol
```diff
@@ -1039,13 +946,157 @@ contract Checkout_swapWithoutFeeSpecialOrder is Checkout {
...
+    assertEq(removalIds[1], _removalIds[3]);
+    assertEq(removalIds[2], _removalIds[4]);
+    assertEq(removalAmounts[0], 1 ether);
+    assertEq(removalAmounts[1], 1 ether);
+    assertEq(removalAmounts[2], 0.5 ether);
+  }
+
+  function test_revertsWhenNoRemovalsFromSpecifiedVintages() external {
+    uint256[] memory nonexistentVintages = new uint256[](2);
+    vintages[0] = 2023; // no such vintages available in market
```
jaycenhorton, line 1084: Nit: Consider using a safer vintage (e.g., 1,9999, etc) that won't accidentally be listed (e.g. with an amount that partially can fill the order thus creating the same error but no test failure)

test/checkout.int.t.sol
jaycenhorton: This applies to multiple tests you've added to this file

test/checkout.int.t.sol
```diff
@@ -1039,13 +946,157 @@ contract Checkout_swapWithoutFeeSpecialOrder is Checkout {
...
+        entries[createCertificateEventIndex].data,
+        (address, uint256, uint256[], uint256[], uint256, uint256)
+      );
+    assertEq(from, address(_removal));
+    assertEq(eventCertificateAmount, certificateAmount);
+    assertEq(priceMultiple, _priceMultiple);
+    assertEq(noriFeePercentage, customFee);
+    assertEq(removalIds.length, 3);
+    assertEq(removalAmounts.length, 3);
+    assertEq(removalIds[0], _removalIds[2]);
```
jaycenhorton, line 1074: Is there no asserEq fn signature to compare array, e.g., assertEq(removalIds.slice(...), _removalIds.slice(...))

test/checkout.int.t.sol
```diff
@@ -1039,13 +946,157 @@ contract Checkout_swapWithoutFeeSpecialOrder is Checkout {
...
+      (certificateAmount * customPriceMultiple) / 100 / 2 // divide to account for price multiple scale and holdback percentage of 50%
     );
     assertEq(_bpNori.balanceOf(_namedAccounts.feeWallet), 0);
   }
 }
 
-contract Checkout_swapFromSupplierWithoutFeeSpecialOrder is Checkout {
+contract Checkout_swapWithoutFeeSpecialOrder_specificVintages is Checkout {
+  uint256 ownerPrivateKey = 0xA11CE;
+  address owner = vm.addr(ownerPrivateKey);
```
jaycenhorton, line 957: Use named accounts

test/checkout.int.t.sol
jaycenhorton: Same question for the previous tests listed that the diff appears to indicate were deleted

test/checkout.int.t.sol
```diff
@@ -1119,8 +1171,150 @@ contract Checkout_swapFromSupplierWithoutFeeSpecialOrder is Checkout {
     assertEq(removalAmounts[0], certificateAmount);
     assertEq(
       _bpNori.balanceOf(_namedAccounts.supplier),
-      (certificateAmount * customPriceMultiple) / 100 / 2 // divide to account for price multiple scale and then holdback percentage of 50%
+      (certificateAmount * customPriceMultiple) / 100 / 2 // divide to account for price multiple scale and holdback percentage of 50%
```
jaycenhorton, line 1174: Nit: use variable names to self document (e.g., holdbackPercent, priceMultipleScale)

contracts/Market.sol
amiecorso: good idea

test/checkout.int.t.sol
amiecorso: there are a few tests that were removed because our contracts' functionality was actually reduced. for example, you can't buy from a single supplier with a permit anymore -- we removed that endpoint because it's not public and now it's part of the special order flow which is all approval based.  This particular test was just renamed I believe to drop the `byApproval` because there's no longer a permit v.s. approval distinction for specific supplier purchases.  I didn't delete any tests I could rewrite to accommodate the new contract interface -- no test functionality should have been reduced except for cases where our contract's functionality is actually being reduced and the test no longer applies and couldn't be converted to an analogous test using the new special order swap function.

contracts/Market.sol
amiecorso: I was able to get better code reuse in all 4 allocation algorithms with this.  ran into some stack too deep errors and refactored how we manage `ids`, `amounts`, `suppliers`, and `countOfRemovalsAllocated` by creating the struct that contains them when they are initialized and referencing the struct throughout the algorithm.  This saves us a few lines of code, got rid of the stack too deep error, and I did a snapshot --diff to make sure the change didn't affect gas and there were only a few tests affected with changes less than 0.05% in all cases, with equal numbers of tests improving in gas as worsening.

contracts/Market.sol
amiecorso: this was fixed by the variable reorganization in the code reuse effort and i'm able to switch to calldata, saving a few gas (snapshot committed along with calldata commit)

jaycenhorton (dismissed):

test/checkout.int.t.sol
amiecorso: oo there is

test/checkout.int.t.sol
amiecorso: makes sense -- this might have been an artifact of adapting permit-based tests which need access to the private key for signing

test/checkout.int.t.sol
amiecorso: switched to using `_namedAccounts.buyer` everywhere where we don't need access to the private key for permit sigs

jaycenhorton: https://book.getfoundry.sh/reference/forge-std/derive-remember-key?highlight=Derivekey#see-also

Additional "see also" links are also relevant

amiecorso: very cool.  trying to figure out how that works with our `_namedAccounts` given that they don't use a mnemonic but `deriveKey` requires a mnemonic... unless the mnemonic is equivalent to `uint160(uint256((keccak256(abi.encodePacked(name)))))` which is being fed into `address`.  at any rate, did a huge test hygiene overhaul in this PR including a bunch of stuff that wasn't otherwise affected here so I'm gonna leave this one for another day

jaycenhorton: Do we not use a mnemonic? Thought I had configured that, but maybe didn't get to it.

amiecorso: no it looks like this: 

```
    NamedAccounts({
      // todo use .env: https://book.getfoundry.sh/tutorials/solidity-scripting?highlight=.env#deploying-our-contract
      deployer: 0x465d5a3fFeA4CD109043499Fa576c3E16f918463,
      _this: account("_this"),
      admin: account("admin"),
      supplier: account("supplier"),
      supplier2: account("supplier2"),
      supplier3: account("supplier3"),
      buyer: account("buyer"),
      feeWallet: account("feeWallet")
    });
```

where `account` looks like this:

```
  function account(string memory name) internal returns (address) {
    address addr = keccak256(bytes(name)) == keccak256("_this")
      ? address(this)
      : address(uint160(uint256((keccak256(abi.encodePacked(name))))));
    vm.label(addr, name);
    return addr;
  }
```

contracts/Market.sol
```diff
@@ -681,12 +687,10 @@ contract Market is
...
   ) external whenNotPaused {
     _validateCertificateAmount({amount: amount});
-    (
-      uint256 countOfRemovalsAllocated,
-      uint256[] memory ids,
-      uint256[] memory amounts,
-      address[] memory suppliers
-    ) = _allocateRemovals({purchaser: _msgSender(), certificateAmount: amount});
+    SupplyAllocationData memory allocationData = _allocateRemovals({
+      purchaser: _msgSender(),
```
jaycenhorton, line 691: Does this param still make sense? Looks like it's always set to msgSender now.

contracts/Market.sol
```diff
@@ -941,155 +768,38 @@ contract Market is
...
    * purchased, scaled by the price multiple.
-   * @param customFee The custom fee percentage that was paid to Nori, as an integer, specified here for
-   * inclusion in emitted events.
-   * @param customPriceMultiple The custom price that will be charged for this transaction.
+   * @param customFee The fee percentage that was paid to Nori off chain, as an integer, specified here
+   * for inclusion in emitted events.
+   * @param customPriceMultiple The price that will be charged for this transaction.
+   * @param supplier The only supplier address from which to purchase carbon removals in this transaction, or
+   * zero address if any supplier is valid.
+   * @param vintages The valid set of vintages from which to fulfill this order, empty if any vintage is valid.
```
jaycenhorton, line 776:
```suggestion
   * @param vintages The valid set of vintages from which to fulfill this order, or an empty array if any vintage is valid.
```

jaycenhorton (approved):
Left one suggestion about considering removing a purchaser param. Worth considering, otherwise lgtm

contracts/Market.sol
amiecorso: good catch

jaycenhorton (approved):

PR #702, opened 2023-09-20
Audit 4 remediations M-2, Q-1, Q-2, Q-3, Q-4

amiecorso:
NO-3237
Here's a link to the audit report: https://www.notion.so/0xmacro/Nori-4-Preliminary-Audit-Report-external-private-de8ef704be0c4ced85bd8eee8f5764d5

Remediations for minor audit feedback -- does not address permit-based vulnerability or the risky gas optimization which I will prototype in another PR


contracts/Market.sol
```diff
@@ -1101,10 +1108,13 @@ contract Market is
...
    * @dev This value is scaled by 100 to allow for decimal precision. For example, a value of 100 means
    * that 1 base token is required to purchase 1 NRT, while a value of 1995 means that 19.95 base tokens
-   * purchase 1 NRT.
+   * purchase 1 NRT. The minimum value for the price multiple is 100, to avoid loss of decimal precision.
    * @param priceMultiple The new price multiple.
    */
   function _setPriceMultiple(uint256 priceMultiple) internal {
+    if (priceMultiple < 100) {
+      revert InvalidPriceMultiple();
+    }
```
amiecorso, line 1117: There actually is a test that covers this line of code ... I think there's an issue with codecov picking this up correctly perhaps related to our forge testing patterns with helper contracts?

rlfarman (dismissed):
Wahoo! LGTM

contracts/Market.sol
jaycenhorton: Perhaps a slightly different error, though I'm surprised it wasnt counted as covered in this case

contracts/Market.sol
```diff
@@ -1404,14 +1414,13 @@ contract Market is
    * ##### Requirements:
    *
    * - Amount is not zero.
-   * - Amount is divisible by 10^(18 - `_purchasingToken.decimals()` + 2). This requirement means that the smallest
-   * purchase amount for a token with 18 decimals (e.g., NORI) is 100, whilst the smallest purchase amount for a token
-   * with 6 decimals (e.g., USDC) is 100,000,000,000,000.
+   * - Amount is divisible by 10^(18 - `_purchasingToken.decimals()` + FEE_DECIMALS). This requirement means that the
```
jaycenhorton, line 1417:
```suggestion
   * - Amount is divisible by 10^(18 - `_purchasingToken.decimals()` + `FEE_DECIMALS`). This requirement means that the
```

contracts/Market.sol
```diff
@@ -1404,14 +1414,13 @@ contract Market is
...
+   * smallest purchase amount for a token with 18 decimals (e.g., NORI) and 2 FEE_DECIMALS is 100, whilst the smallest
```
jaycenhorton, line 1418:
```suggestion
   * smallest purchase amount for a token with 18 decimals (e.g., NORI) and 2 `FEE_DECIMALS` is 100, whilst the smallest
```

contracts/Market.sol
```diff
@@ -1404,14 +1414,13 @@ contract Market is
...
+   * purchase amount for a token with 6 decimals (e.g., USDC) and 2 FEE_DECIMALS is 100,000,000,000,000.
```
jaycenhorton, line 1419:
```suggestion
   * purchase amount for a token with 6 decimals (e.g., USDC) and 2 `FEE_DECIMALS` is 100,000,000,000,000.
```

contracts/Market.sol
jaycenhorton: possible it's [this](https://github.com/foundry-rs/foundry/issues/4316) bug?

jaycenhorton (changes requested):

contracts/Market.sol
amiecorso: interesting... that seems possible since the test is in a test contract that inherits `NonUpgradeableMarket` which inherits `Market` -- this is different from the pattern in most of our tests that inherit `UpgradeableMarket` which does not inherit the target contracts but rather has instances of them I believe.... (and for which coverage seems to be working more as expected)

jaycenhorton (approved):

PR #703, opened 2023-09-20
Audit 4 gas optimization -- inline assembly ⚠️

amiecorso:
NO-3237

Addresses G-4 from [Macro audit 4](https://www.notion.so/0xmacro/Nori-4-Preliminary-Audit-Report-external-private-de8ef704be0c4ced85bd8eee8f5764d5) 

This does appear to provide some gas savings and we just need to decide if it's worth the risk of using inline assembly to modify the original arrays in place.  From what I can tell, the biggest risk would be if there is any case where the original array's contents remain important for other code despite being cut down by different code.  The original memory slots of the longer array remain untouched, but I think this is ok because not only are they all the default 0 value anyway, but we don't need them for any other reason in the logic in this contract.


contracts/ArrayLib.sol
```diff
@@ -129,6 +129,34 @@ library UInt256ArrayLib {
...
+   *
+   * ```solidity
+   * new uint256[](100).fill(1).cutToLength(50); // resizes the original array to length 50
+   * ```
+   * -
+   * @param values The array to shorten.
+   * @param length The desired length of the array.
+   * @return values The shortened array.
+   */
+  function cutToLength(
```
jaycenhorton, line 148:
```suggestion
  function shrink(
```

jaycenhorton: also, do we still use .slice elsewhere? If not, we should delete that from ArrayLib.

jaycenhorton (changes requested):

contracts/ArrayLib.sol
amiecorso: per our discussion -- we're going to keep `slice` because it's used in some tests and it's already audited and may be useful in the future

jaycenhorton (approved):

PR #706, opened 2023-09-28
Update Fireblocks Signer

amiecorso:
NO-3258

Updates the contracts repo hardhat infrastructure away from our old custom Fireblocks signer to instead use the recommended hardhat integration (which didn't exist when we were initially building this so Scott put together something custom).  We also extend the base implementation of the fireblocks-hardhat integration to implement complete ethers Signer functionality. The custom integration was no longer working when we revisited it this month to perform some script-based listing and while it's not clear whether it's deprecated beyond salvation (for example, if the Fireblocks rest API is no longer live, then it is) we figure the easier and likely inevitable fix is to migrate to this finally mature third-party library created by Fireblocks, especially given the complexity of the custom implementation.


package.json
```diff
@@ -116,5 +116,9 @@
     "snapshot:production": "FOUNDRY_PROFILE=production forge snapshot --snap .gas-snapshot-production",
     "lint:diff": "npx eslint --cache --cache-location ./.eslintcache -c .eslintrc.js $(git diff --relative --name-only --diff-filter=d $(git merge-base HEAD origin/master) -- \"*.ts\" \"*.js\" \"*.env\" \"*.toml\")",
     "artifactregistry-login": "npx google-artifactregistry-auth"
+  },
+  "dependencies": {
```
jaycenhorton, line 120: should these be devDependencies?

plugins/index.ts
```diff
@@ -119,12 +119,9 @@ const deployOrUpgradeProxy = async <
...
   ) {
     hre.trace('Deploying proxy and instance', contractName);
-    const fireblocksSigner = signer as FireblocksSigner;
-    if (typeof fireblocksSigner.setNextTransactionMemo === 'function') {
-      fireblocksSigner.setNextTransactionMemo(
-        `Deploy proxy and instance for ${contractName}`
-      );
-    }
+    const fireblocksSigner = signer as unknown as FireblocksWeb3Provider;
+    // eslint-disable-next-line
```
jaycenhorton, line 123: can you add the rule name so it only disables the violated rule?

plugins/index.ts
```diff
@@ -119,12 +119,9 @@ const deployOrUpgradeProxy = async <
     // )
   ) {
     hre.trace('Deploying proxy and instance', contractName);
-    const fireblocksSigner = signer as FireblocksSigner;
-    if (typeof fireblocksSigner.setNextTransactionMemo === 'function') {
-      fireblocksSigner.setNextTransactionMemo(
-        `Deploy proxy and instance for ${contractName}`
-      );
-    }
+    const fireblocksSigner = signer as unknown as FireblocksWeb3Provider;
```
jaycenhorton, line 122: why is this casting needed? What is signer initially? Should we instead add a check on the type of the signer so we can use type narrowing and throw an error instead if the wrong signer is being used?

plugins/index.ts
```diff
@@ -146,14 +143,11 @@ const deployOrUpgradeProxy = async <
...
+        await hre.upgrades.erc1967.getImplementationAddress(maybeProxyAddress);
       hre.trace('Existing implementation at:', existingImplementationAddress);
-      const fireblocksSigner = signer as FireblocksSigner;
-      if (typeof fireblocksSigner.setNextTransactionMemo === 'function') {
-        fireblocksSigner.setNextTransactionMemo(
-          `Upgrade contract instance for ${contractName}`
-        );
-      }
+      const fireblocksSigner = signer as unknown as FireblocksWeb3Provider;
+      // eslint-disable-next-line
```
jaycenhorton, line 149: can you add the rule name so it only disables the violated rule?

plugins/index.ts
```diff
@@ -211,10 +205,9 @@ const deployNonUpgradeable = async <
     contractName,
     { ...options, signer }
   );
-  const fireblocksSigner = signer as FireblocksSigner;
-  if (typeof fireblocksSigner.setNextTransactionMemo === 'function') {
-    fireblocksSigner.setNextTransactionMemo(`Deploy ${contractName}`);
-  }
+  const fireblocksSigner = signer as unknown as FireblocksWeb3Provider;
+  // eslint-disable-next-line
```
jaycenhorton, line 209: can you add the rule name so it only disables the violated rule?

plugins/index.ts
jaycenhorton: same question for the several other places this happens

plugins/index.ts
```diff
@@ -233,19 +226,22 @@ extendEnvironment((hre) => {
   hre.debug = lazyFunction(() => debug);
   // All live networks will try to use fireblocks and fall back to hd wallet
   if (hre.network.config.live) {
+    if (hre.config.fireblocks === undefined) {
+      throw new Error(
+        'Fireblocks config is required for live networks. Please set FIREBLOCKS_API_KEY and FIREBLOCKS_SECRET_KEY_PATH in your environment.'
```
jaycenhorton, line 231:
```suggestion
        'Fireblocks config is required for live networks. Please set FIREBLOCKS_API_KEY, FIREBLOCKS_SECRET_KEY_PATH, and FIREBLOCKS_VAULT_ID in your environment.'
```

tasks/migrate-certificates.ts
```diff
@@ -417,7 +417,6 @@ export const GET_MIGRATE_CERTIFICATES_TASK = () =>
         `Certificate contract address: ${certificateContract.address}`
       );
       logger.info(`Signer address: ${signerAddress}`);
-      // const fireblocksSigner = removalContract.signer as FireblocksSigner;
```
jaycenhorton, line 420: curious why this was needed in the past-- any idea?

tasks/utils/contract-functions.ts
```diff
@@ -1,5 +1,5 @@
 import '@nomiclabs/hardhat-ethers';
-import '@/plugins/fireblocks';
+import '@fireblocks/hardhat-fireblocks';
```
jaycenhorton, line 2: do you know if these tasks work with the fireblocks signer plugin?

tasks/vesting.ts
```diff
@@ -845,19 +844,17 @@ const CREATE_SUBTASK = {
...
       const owner = await fireblocksSigner.getAddress();
       const name = await bpNori.name();
       const nonce = await bpNori.nonces(owner);
       const chainId = await fireblocksSigner.getChainId();
-      if (typeof fireblocksSigner.setNextTransactionMemo === 'function') {
-        fireblocksSigner.setNextTransactionMemo(
-          `Permit BridgedPolygonNORI Spend by LockedNORI: ${memo || ''}`
-        );
-      }
+      // eslint-disable-next-line
```
jaycenhorton, line 856: add rule you're intending to disable

tasks/vesting.ts
```diff
@@ -996,12 +993,11 @@ const REVOKE_SUBTASK = {
...
       } else {
-        const fireblocksSigner = lNori.signer as FireblocksSigner;
-        if (typeof fireblocksSigner.setNextTransactionMemo === 'function') {
-          fireblocksSigner.setNextTransactionMemo(
-            `Vesting Revoke: ${memo || ''}`
-          );
-        }
+        const fireblocksSigner =
+          lNori.signer as unknown as FireblocksWeb3Provider;
+        // eslint-disable-next-line
```
jaycenhorton, line 998: add rule you're intending to disable

types/global.d.ts
```diff
@@ -55,6 +55,13 @@ declare module 'hardhat/config' {
 
   function extendEnvironment(extender: EnvironmentExtender): void;
 
+  interface HardhatConfig {
+    fireblocks?: FireblocksProviderConfig;
```
jaycenhorton, line 59: I think we need to import `FireblocksProviderConfig` otherwise I believe this gets typed as any

types/global.d.ts
```diff
@@ -230,7 +235,7 @@ declare global {
     upgrades: CustomHardhatUpgrades;
     network: Omit<Network, 'name'> & { name: keyof typeof networks };
     ethers: typeof ethers;
-    getSigners: () => Promise<(Signer & TypedDataSigner)[]>;
+    getSigners: () => Promise<Signer[]>;
```
jaycenhorton, line 238: should this be `| FireBlocksSignerTypeHere` ?

package.json
amiecorso: mm yeah good catch.  I think I can remove `fireblocks-sdk` and `fireblocks-defi-sdk` while I'm at it too

tasks/migrate-certificates.ts
amiecorso: I don't think it ever was.  I'm sure there was an original intention for it but I remember that line living commented in the scripts for their lifetime

types/global.d.ts
jaycenhorton: Even still, in case the two types deviate in the future we won't know if we don't make the differentiation here.

jaycenhorton: also, `FireblocksSigner['_fireblocksWeb3Provider']` is only valid in one context afaict

tasks/utils/contract-functions.ts
amiecorso: they sure doooo

jaycenhorton (approved):

plugins/fireblocks/type-extensions.ts
```diff
@@ -1,16 +1,23 @@
...
-import type { FireblocksSignerConfig } from './types';
+import type { FireblocksProviderConfig } from '@fireblocks/fireblocks-web3-provider';
+
 import type { FireblocksSigner } from './fireblocks-signer';
 
 declare module 'hardhat/types/config' {
-  interface HardhatUserConfig {
-    fireblocks?: FireblocksSignerConfig;
+  interface HttpNetworkUserConfig {
+    fireblocks?: FireblocksProviderConfig;
```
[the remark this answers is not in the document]
amiecorso: nope -- just the config that's used to set up the signers under the hood

PR #711, opened 2023-10-03
Add audit 4 to readme

amiecorso:


docs/audits/README.md
```diff
@@ -1,5 +1,8 @@
 ## Audits
+### 2023
+- [Macro Audit #4](https://library.0xmacro.com/library/audits/nori-4) audited _ArrayLib_, _Certificate_, _Errors_, _ICertificate_, _IERC20WithPermit_, _IMarket_, _IRemoval_, _IRestrictedNORI_, _Market_, _Removal_, _RemovalIdLib_, _RemovalsByYearLib_, _RestrictedNORI_, and _RestrictedNORILib_ in September of 2023.
 
+(Audit #3 was only partially completed)
```
amiecorso, line 5: wondering if we should even make a comment about this? or if so, a better one?

jaycenhorton: If there's any public info (even a PR) that references the issue and the solution, then keep it. Otherwise I'd just drop it and decrement #4

PR #716, opened 2023-10-17
Polygon upgrades post audit 4

amiecorso:


contracts/Market.sol
```diff
@@ -684,7 +683,6 @@ contract Market is
    */
   function swap(
     address recipient,
-    address permitOwner,
```
amiecorso, line 687: the permit fix

rlfarman (approved):

PR #725, opened 2023-11-13
Add allowlist role to `swap` endpoint for sanctions compliance

amiecorso:
[NO-3363]

Put `swap` behind a role that we can use to allowlist addresses.

[NO-3363]: https://norinauts.atlassian.net/browse/NO-3363?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


contracts/Market.sol
```diff
@@ -194,6 +195,12 @@ contract Market is
    */
   bytes32 public constant ALLOWLIST_ROLE = keccak256("ALLOWLIST_ROLE");
 
+  /**
+   * @notice Role conferring the ability to purchase using the `swap` endpoint.
+   */
+  bytes32 public constant SANCTION_ALLOWLIST_ROLE =
+    keccak256("SANCTION_ALLOWLIST_ROLE");
```
rlfarman, line 202: As much as I enjoy the idea of enforcing US sanctions through restricting our endpoints with this role, the naming could be a little clearer, and could maybe allow this role to be used so that it semantically extends beyond just enforcing sanctions.

In any case, it also makes sense to rename the now too-generic `ALLOWLIST_ROLE` to be something like `PRIORITY_ALLOWLIST_ROLE` or something like that. And change this role to `ALLOWLIST_ROLE`, `SWAP_ALLOWLIST_ROLE`, etc...

Otherwise, I would at least change this to `UNSANCTIONED_ALLOWLIST_ROLE`, since the naming is a little confusing as is. As an uninformed user this might make me think we're allowing certain sanctioned actors using this role.

But these are just naming nits!

amiecorso: agreed -- I assumed we might be doing a global rename on this.  At first I wasn't sure if we'd already granted some `ALLOWLIST_ROLE`s, which would become invalidated by changing the name to `PRIORITY_ALLOWLIST_ROLE`.  But then I realized you can get the count of role members on polygonscan for a given role, and the only account with the allowlist role currently is the deployer account which is automatically assigned the role in the contract initializer.  So I agree we should make both names more descriptive and go with `PRIORITY_ALLOWLIST_ROLE` and `SWAP_ALLOWLIST_ROLE` which I think is a little more abstract and appropriate.

rlfarman (approved):
【﻿ＬＧＴＭ】

PR #730, opened 2024-01-25
Remove USDC (and all purchase-related functionality) from onchain market

amiecorso:
Contracts compile and all foundry tests are now passing -- some are obsolete and were removed, others needed some modification now that `swap` has been removed from the Market in favor of only one endpoint, which is still called `swapWithoutFeeSpecialOrder`

TODO: 

- [ ] still need to get Hardhat tests passing as well as a few commented-out forge tests, but this captures the gist of what would be necessary to remove all token flows from our onchain market.
- [ ] need to update all documentation / READMEs


contracts/Market.sol
```diff
@@ -136,35 +124,32 @@ contract Market is
   Certificate private _certificate;
 
   /**
-   * @notice The IERC20WithPermit token used to purchase from this market.
+   * @notice Deprecated.
    */
-  IERC20WithPermit private _purchasingToken;
+  address private _purchasingToken;
```
amiecorso, line 129: I figured we might leave these storage variables in place if we were to perform these changes to protect against upgrade-related incompatibilities

PR #735, opened 2024-03-06
Public transfer endpoint for `CONSIGNOR` to move tokens

amiecorso:
[NO-3614]

I chose to make this a separate endpoint from the existing `safeBatchTransferFrom` because that calls the superclass `safeBatchTransferFrom` which does enforce that the sender is the owner or approved of the tokens (a limitation we need our `CONSIGNOR` to be exempt from)

This separate endpoint 1) leaves the security of the original `safeBatchTransferFrom` intact and 2) makes it simpler to reason about permission and functionality.

[NO-3614]: https://norinauts.atlassian.net/browse/NO-3614?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


contracts/Removal.sol
```diff
@@ -648,6 +648,37 @@ contract Removal is
...
+   * @notice Allows an address with the `CONSIGNOR_ROLE` to transfer tokens.
+   * @dev Emits a `TransferBatch` event.
+   *
+   * ##### Requirements:
+   *
+   * - Can only be called by an address with the `CONSIGNOR_ROLE`.
+   * - Respects the rules of `Removal._beforeTokenTransfer`.
+   * - `ids` and `amounts` must have the same length.
+   * - If `to` refers to a smart contract, it must implement {IERC1155Receiver-onERC1155BatchReceived} and return the
+   * acceptance magic value.
```
rlfarman, line 661: 😂

rlfarman (approved):

PR #736, opened 2024-03-08
Update the `withdraw` method on Market contract to allow `CONSIGNOR` withdrawals

amiecorso:
[NO-3615]

[NO-3615]: https://norinauts.atlassian.net/browse/NO-3615?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


test/Market.t.sol
```diff
@@ -629,31 +661,50 @@ contract Market_SANCTION_ALLOWLIST_ROLE is UpgradeableMarket {
   }
 }
 
-contract Market__isAuthorizedWithdrawal_true is NonUpgradeableMarket {
+contract Market__isAuthorizedWithdrawal_true is
+  NonUpgradeableMarket,
+  UpgradeableRemoval
+{
+  Removal private _mockRemoval;
```
amiecorso, line 668: since the function changed to now make a call to `_removal.hasRole.(...)` (which itself calls `_removal.CONSIGNOR_ROLE()` we needed to make the mock for these return values more robust and explicit than the basic mock that was in place for `isApprovedForAll` which is what this is about

contracts/Market.sol
```diff
@@ -1349,16 +1352,26 @@ contract Market is
...
   /**
-   * @dev Authorizes withdrawal for the removal. Reverts if the caller is not the owner of the removal and
-   * does not have the role `MARKET_ADMIN_ROLE`.
-   * @param owner The owner of the removal.
-   * @return Returns true if the caller is the owner, an approved spender, or has the role `MARKET_ADMIN_ROLE`,
-   * false otherwise.
+   * @dev Authorizes withdrawal for the removal. Reverts if the caller is not the supplier of the removal and
+   * does not have the role `MARKET_ADMIN_ROLE`, or if the recipient of the removal is either not the supplier
+   * of the removal or does not have the `Removal.CONSIGNOR_ROLE`.
+   * @param supplier The supplier of the removal.
```
rlfarman, line 1358: Missing param definition for @param to?

contracts/Market.sol
```diff
@@ -1349,16 +1352,26 @@ contract Market is
...
+  ) internal view returns (bool) {
+    return
+      (_msgSender() == supplier ||
+        hasRole({role: MARKET_ADMIN_ROLE, account: _msgSender()}) ||
+        _removal.isApprovedForAll({
+          account: supplier,
+          operator: _msgSender()
+        })) &&
+      (_removal.hasRole({role: _removal.CONSIGNOR_ROLE(), account: to}) ||
+        supplier == to);
```
rlfarman, line 1374: Might read a little cleaner to split these up into additional internal functions IE - `_isApprovedSpender || _hasMarketAdminRole`, etc...
Nit/nice-to-have, just hard to parse all these conditional statements.

rlfarman (changes requested):
One change to the documentation otherwise LGTM!

contracts/Market.sol
amiecorso: took the middle route by defining each predicate as its own well-named variable, but leaving the helper as one function

rlfarman: That helps a lot, thanks!

rlfarman (approved):

